PR #1, opened 2018-01-03
Create core project skeleton

danielpassos:


core/build.gradle
```diff
@@ -0,0 +1,31 @@
+apply plugin: 'com.android.library'
+
+android {
+    compileSdkVersion 26
+
+    defaultConfig {
+        minSdkVersion 21
```
wtrocki: [Idea] If we want to keep this consistent across all sdk modules we can use constants and import them.
For example here:
https://github.com/facebook/facebook-android-sdk/blob/master/constants.gradle

core/src/androidTest/java/org/aerogear/android/core/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,26 @@
+package org.aerogear.android.core;
```
wtrocki: This can be removed.

wtrocki:
@danielpassos Build locally. It will be good to have some basic readme update

danielpassos:
@wtrocki Which kind of info you wanna have in the README since we have nothing yet?

wtrocki:
I was just suggesting to add something small. Its good practice for community to just have one sentence what is the project doing and that it is in development. Empty Readme gives zero info. Aerogear has now couple sdk repos so this will reduce confusion.

danielpassos:
@wtrocki I got your point here but still having no idea what to add there. Feel free to send a PR with your thoughts

danielpassos:
@secondsun Wanna double check before I merge it?

wtrocki:
@danielpassos Ignore this comment. Feel free to merge.

PR #2, opened 2018-01-11
[WIP] Mobile core PoC

secondsun:
This is the mobile Core PoC, it is not functional but can be referenced as work continues until it is ready for a full review.  Do not merge, this should track the Android SDK Proposal.


core/src/main/java/org/aerogear/mobile/core/ServiceModuleRegistry.java
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Register the looking for a service module to a class.
+     *
+     * @param type the type of module to use
+     * @param moduleClass the class that implements a service module type
+     * @param dependsOn the other serviceTypes this Service Depends on.
+     */
+    public static void registerServiceModule(String type, Class<? extends ServiceModule> moduleClass, String... dependsOn) {
+        serviceTypeMap.put(type, moduleClass);
+        for (String dependency : dependsOn) {
```
wtrocki: Services should be independent. Can you add documentation what this will mean. 
I remember that in proposal we were saying that service may depend on keycloak, but I believe that we do not need to specify any dependency - just assume that keycloak service will protect services like sync if exist.

It's cleaner to just have all services instantiated IMHO.

core/src/main/java/org/aerogear/mobile/core/http/HttpServiceModule.java
```diff
@@ -0,0 +1,15 @@
+package org.aerogear.mobile.core.http;
+
+import org.aerogear.mobile.core.MobileCore;
+import org.aerogear.mobile.core.ServiceModule;
+import org.aerogear.mobile.core.configuration.ServiceConfiguration;
+
+public class HttpServiceModule implements ServiceModule {
```
wtrocki: Core will have services or it's just temporary?

keycloak-service-module/src/androidTest/java/org/aerogear/mobile/keycloak_service_module/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,26 @@
+package org.aerogear.mobile.keycloak_service_module;
```
wtrocki: See naming suggestions email thread. Suggestion was to drop `keycloak` and use `auth`
Module name seems redundant. WDYT `auth_service`?

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakConfig.java
```diff
@@ -0,0 +1,87 @@
...
+                        config.resource = reader.nextString();
+                        break;
+                    case "public-client":
+                        config.publicClient = reader.nextBoolean();
+                        break;
+                    case "confidential-port":
+                        config.confidentialPort = reader.nextInt();
+                        break;
+                    default:
+                        reader.skipValue();
```
wtrocki: Let's add FIXME here as this may cause some bugs later.

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
```diff
@@ -0,0 +1,54 @@
...
+     * Exchanges the google id token and configures the KeyCloakService to serve requests
+     *
+     * @param token a Google ID token
+     */
+    public void login(String token) {
+
+    }
+
+    @Override
+    public void bootstrap(MobileCore core, ServiceConfiguration config, Object... args) {
```
wtrocki: MobileCore is class. It will be cool to extract that to interface.

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
```diff
@@ -0,0 +1,54 @@
...
+        this.serverUrl = config.getProperty("auth-server-url");
```
wtrocki: [Question] Do you think we can we avoid using keys for this and just somehow pass that to keycloak.
It seems that we should not care about what config is there (as it may change or have new keys).

core/src/main/java/org/aerogear/mobile/core/ServiceModuleRegistry.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Registry of service modules.  It is responsible for maintaining a mapping of service modules to
+ * their types as well as the dependencies each module expects.
+ */
+public final class ServiceModuleRegistry {
+    private static final Map<String, Class<? extends ServiceModule>> serviceTypeMap = new HashMap<>();
```
wtrocki: How this will be used?
Will this work assuming that namespace can have only one service of certain type.?
It may be worth to check if that is not limitation for some of the planned services.

wtrocki:
@secondsun Added couple comments mostly to get answers and your point of view. 
As this is WIP do not feel that you need to address any of those.

keycloak-service-module/src/androidTest/java/org/aerogear/mobile/keycloak_service_module/ExampleInstrumentedTest.java
secondsun: Keycloak is an implementation of an auth spec we have not defined.  Calling it auth seems a bit premature.

wtrocki: @secondsun - Actually there is already service merged into the core. We will need to rebase this to get that in.

secondsun:
@wtrocki See https://github.com/secondsun/WipDemo for a demo.

core/src/test/java/org/aerogear/mobile/core/HttpServiceModuleTest.java
```diff
@@ -0,0 +1,30 @@
...
+@RunWith(RobolectricTestRunner.class)
+@SmallTest
+public class HttpServiceModuleTest {
+
+    @Test
+    public void testGet() {
+        HttpServiceModule module = new HttpServiceModule();
+        String response = module.get("http://www.mocky.io/v2/5a5f74172e00006e260a8476");
+        assertNotNull(response);
+        assertEquals("{\n" +
```
josemigallas: The content of the response should be tested, instead of the format, no? Something like:
```
Whatever response = module.get("http://www.mocky.io/v2/5a5f74172e00006e260a8476").asWhatever();
assertNotNull(response);
assertEquals(response.getStory().getTitle(), "Test Title");

secondsun: Maybe.  Right now we are verifying that the response that we expect comes back.  I've been doing this as a TDD pair programming exercise so code on the bleeding most edge would be contrived.  As things get stable I rebase them out into their own PRs for a proper review.

secondsun:
I'm closing this PR.  The WiP portion has been completed and this PR has been replaced by https://github.com/aerogear/aerogear-android-sdk/pull/16 and https://github.com/aerogear/aerogear-android-sdk/pull/17

PR #3, opened 2018-01-12
Skeleton implementation for the IDM module

ziccardi:
# Motivation
We want to have a skeleton implementation for the IDM module to be filled later on with the real implementation


auth/src/main/java/org/aerogear/auth/IRole.java
```diff
@@ -0,0 +1,10 @@
+package org.aerogear.auth;
+
+/**
+/**
+ * Created by ziccardi on 12/01/2018.
+ */
+
```
secondsun: Default javadoc, replace with somethign more descriptive

auth/src/main/java/org/aerogear/auth/ICredential.java
```diff
@@ -0,0 +1,4 @@
+package org.aerogear.auth;
+
+public interface ICredential {
```
wei-lee: @danielpassos @secondsun do we have any existing Java naming conventions for the AG android projects that we should follow?

ziccardi:
@wei-lee @secondsun ping

auth/src/main/java/org/aerogear/auth/utils/TokenPersistor.java
```diff
@@ -0,0 +1,31 @@
...
+import org.aerogear.auth.credentials.TokenCredentials;
+
+/**
+ * Saves, retrieves and delete a token.
+ */
+public class TokenPersistor {
+    private TokenPersistor() {
+    }
+
+    public static TokenCredentials load() {
```
ziccardi: @wei-lee what data should we pass here? Would it be a good idea to pass here the Principal?

auth/src/main/java/org/aerogear/auth/utils/TokenPersistor.java
```diff
@@ -0,0 +1,31 @@
...
+        throw new IllegalStateException("Not yet implemented");
+    }
+
+    /**
+     * Saves a token
+     * @param token token to be saved
+     */
+    public static void save(TokenCredentials token) {
```
ziccardi: @wei-lee Would it be a good idea to pass here the Principal? (it already has a method to extract the token credential)

auth/src/main/java/org/aerogear/auth/utils/TokenPersistor.java
```diff
@@ -0,0 +1,31 @@
...
+        throw new IllegalStateException("Not yet implemented");
+    }
+
+    /**
+     * Deletes a token
+     * @param token token to be deleted
+     */
+    public static void delete(TokenCredentials token) {
```
ziccardi: @wei-lee see previous comment

auth/src/main/java/org/aerogear/auth/AuthService.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Log in the user with the given credential.
+     *
+     * @param username The username
+     * @param credentials the credential
+     * @return a user principal
+     * @throws AuthenticationException throws if the authentication fails
+     */
+    public Principal login(final String username, final ICredential credentials) throws AuthenticationException {
```
wei-lee: The way it is done with OCID, we will not have the username/credential here. The login function will invoke a browser, and user will enter their username/credential in the browser. The app itself will have no access to username/credential at all for security reasons.

auth/src/main/java/org/aerogear/auth/utils/TokenPersistor.java
wei-lee: @ziccardi I think we don't need to pass any params here. At any time, there should be only 1 user logged into the app, the we should only keep 1 token at a time. For a new user, it should always remove the existing token and create a new one. So we just need an internal name to store the token.

auth/src/main/java/org/aerogear/auth/utils/TokenPersistor.java
wei-lee: I think pass in token is ok for now. The class is called "TokenPersistor".

auth/src/main/java/org/aerogear/auth/utils/TokenPersistor.java
wei-lee: token is fine.

auth/src/main/java/org/aerogear/auth/AbstractRole.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    private final String roleName;
+
+    /**
+     * Builds a role object
+     * @param roleID Role unique id
+     * @param roleName Role friendly name
+     */
+    AbstractRole(final String roleID, final String roleName) {
+        this.roleID = roleID;
```
TommyJ1994: I don't think we will need a role ID, as Keycloak/OpenID doesn't have that concept. Role names on the IDM server will be unique.

auth/src/main/java/org/aerogear/auth/IUserPrincipal.java
```diff
@@ -0,0 +1,30 @@
...
+     * @param role role to be checked
+     * @return true or false
+     */
+    boolean hasRole(IRole role);
+
+    /**
+     * Returns the username
+     * @return the username
+     */
+    String getName();
```
TommyJ1994: `getUsername()` might be a bit more accurate

auth/src/main/java/org/aerogear/auth/IUserPrincipal.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Returns the session token.
+     * @return the session token
+     */
+    TokenCredentials getToken();
+}
```
TommyJ1994: I think it might be a good idea to expose some more OpenID specific user attributes:

`getName()` - the user's IDM given name.
`getEmail()` - the user's IDM given email

Also maybe a function per token type - 

`getAccessToken()` - Returns the users access token
`getIdentityToken()` - Returns the users identity token
`getRefreshToken()` - Returns the users refresh token

auth/src/main/java/org/aerogear/auth/credentials/PasswordCredentials.java
```diff
@@ -0,0 +1,23 @@
+package org.aerogear.auth.credentials;
+
+import java.util.Arrays;
+
+/**
+ * Credentials for password based authentication
+ */
+public final class PasswordCredentials implements ICredential {
```
TommyJ1994: I'm not sure if we need a password credentials class, as we won't be handling any passwords in the app with OpenID Connect, only some tokens.

auth/src/main/java/org/aerogear/auth/impl/AbstractAuthenticator.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * This method must be overridden with the custom authentication for the given credential.
+     *
+     * @param username username to be authenticated
+     * @param credential user credential
+     * @return the authenticated principal
+     * @throws AuthenticationException
+     */
+    public Principal authenticate(final String username, ICredential credential) throws AuthenticationException {
```
TommyJ1994: The only things we will need to pass in here is some auth config. This will tell the appauth library where the keycloak server is, what realm and client we want to communicate under, what OpenID Connect scopes we want to pass, what browsers we want the users to use to authenticate etc. We don't need to pass any credentials here.

auth/src/main/java/org/aerogear/auth/impl/AbstractAuthenticator.java
```diff
@@ -0,0 +1,32 @@
...
+        throw new IllegalStateException("Not implemented");
+    }
+
+    /**
+     * Logout the given principal
+     * @param principal principal to be log out
+     */
+    public void logout(Principal principal) {
```
TommyJ1994: We will need to take in the identityToken as a param here. The logout URL can be constructed from some keycloak config constants, so we probably fetch them from within the method.

auth/src/main/java/org/aerogear/auth/IUserPrincipal.java
ziccardi: the interface currently extends the java.security.Principal class.
Since a Principal can be different things (not only a User) it forces the implementation of the getName method.
If you believe we could add the getUsername method or remove the inheritance from the Principal interface.

TommyJ1994: Ah, I see. I am just thinking of a use case where a username could be used as sort of an identifier or primary key for some functionality, and the Full Name would be used in the UI on the user's profile.

auth/src/main/java/org/aerogear/auth/impl/AbstractAuthenticator.java
ziccardi: We will pass here the Principal object so that we can have all the three tokens at once and any other information we want to put inside the Principal object

TommyJ1994: Cool, that makes sense too.

auth/src/main/java/org/aerogear/auth/AuthService.java
```diff
@@ -0,0 +1,68 @@
...
+     * selected by analysing the received credentials. If the credentials are null,
+     * the browser will be open asking for authentication
+     *
+     * The login will be asynchronous.
+     *
+     * @param username The username
+     * @param credentials the credential
+     * @return a user principal
+     */
+    public Future<Principal> login(final String username, final ICredential credentials) {
```
wei-lee: I don't think `username` is required as a separate parameter here. It can be included in the implementation class of the `ICredential` interface if it's required. The credential type we are going to support doesn't require `username`

auth/src/main/java/org/aerogear/auth/CredentialsType.java
```diff
@@ -0,0 +1,5 @@
+package org.aerogear.auth;
+
+public enum CredentialsType {
+    TOKEN, BROWSER;
```
wei-lee: not sure those values make sense for `credentials`. 

To make it clear, I think one of the type can be "OpenConnectId". It will be the only one we are going to support for now.

auth/src/main/java/org/aerogear/auth/AuthenticationChain.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.concurrent.Future;
+
+/**
+ * Authenticate the user by executing the authentication chain.
+ * Stops at the first successful login or at the first failing mandatory authenticator.
+ * At list one authenticator must be successful.
+ */
+public class AuthenticationChain {
+
+    private static ExecutorService EXECUTOR = Executors.newFixedThreadPool(10);
```
wei-lee: make this configurable. It's likely the core module will provide executors.

auth/src/main/java/org/aerogear/auth/IKeycloakUserPrincipal.java
```diff
@@ -0,0 +1,16 @@
+package org.aerogear.auth;
+
+
+import org.aerogear.auth.credentials.OIDCCredentials;
+
+public interface IKeycloakUserPrincipal extends IUserPrincipal {
```
wei-lee: Those tokens are not specific to Keycloak, they are all part of the OIDC protocol, so I don't think we need a separate interface for it.

auth/src/main/java/org/aerogear/auth/credentials/UsernamePasswordCredential.java
```diff
@@ -0,0 +1,19 @@
+package org.aerogear.auth.credentials;
+
+public class UsernamePasswordCredential implements ICredential {
```
wei-lee: we are not going to support username/password credentials. So better delete this one.

auth/src/main/java/org/aerogear/auth/impl/BrowserAuthenticatorImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.aerogear.auth.AuthServiceConfig;
+import org.aerogear.auth.credentials.ICredential;
+
+import java.security.Principal;
+
+/**
+ * Authenticates the user by opening a browser window asking for credentials.
+ */
+public class BrowserAuthenticatorImpl extends OICDTokenAuthenticatorImpl {
```
wei-lee: Given that the implementation is mainly based on OIDC protocol, I think it's better to name it `OIDCAuthenticatorImpl`?

auth/src/main/java/org/aerogear/auth/impl/KeycloakUserPrincipalImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+
+import org.aerogear.auth.AbstractAuthenticator;
+import org.aerogear.auth.IKeycloakUserPrincipal;
+import org.aerogear.auth.IRole;
+import org.aerogear.auth.credentials.OIDCCredentials;
+
+import java.util.Map;
+
+public class KeycloakUserPrincipalImpl extends UserPrincipalImpl implements IKeycloakUserPrincipal {
```
wei-lee: all those fields are not specific to Keycloak, they are all part of the OIDC protocol. Let's rename this to `OIDCUserPrinicipalImpl`

auth/src/main/java/org/aerogear/auth/impl/SimplePasswordAuthenticatorImpl.java
```diff
@@ -0,0 +1,38 @@
...
+import org.aerogear.auth.AbstractAuthenticator;
+import org.aerogear.auth.AuthenticationException;
+import org.aerogear.auth.credentials.ICredential;
+import org.aerogear.auth.credentials.UsernamePasswordCredential;
+
+import java.security.Principal;
+import java.util.HashMap;
+import java.util.Map;
+
+public class SimplePasswordAuthenticatorImpl extends AbstractAuthenticator {
```
wei-lee: remove this one to avoid confusion

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthenticatorImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import org.aerogear.auth.AuthServiceConfig;
+import org.aerogear.auth.credentials.ICredential;
+import org.aerogear.auth.credentials.OIDCCredentials;
+
+import java.security.Principal;
+
+/**
+ * Authenticates the user by using OpenID Connect.
+ */
+public class OIDCAuthenticatorImpl extends OIDCTokenAuthenticatorImpl {
```
TommyJ1994: For someone with existing knowledge of OIDC, this might be a bit too vague as to what OAuth flow this is referring to. Might be better to have `AuthorisationCode` mentioned somewhere in the class name, as this is the name of the flow we are using here. Maybe `OIDCAuthCodeImpl`might work.

ziccardi: good point!

secondsun:
@wei-lee @ziccardi We need to revisit this PR.  It isn't an android-library which is a HUGE problem.  Sorry I didn't catch it before.

PR #4, opened 2018-01-14
Temp service module

secondsun:
In order to test and build the service modules code, I've added in a keycloak service module to stand in until we have an IDM module to replace it.

@danielpassos @wei-lee Any problems with merging this?  Also, having a second module is important for verifying the maven artifacts are created properly.


secondsun:
@wei-lee Yup.  I 100% agree this code will die at some point.

secondsun:
Landed.

PR #5, opened 2018-01-14
Maven Build Support

secondsun:
This PR adds the ability to generate maven artifacts to the core SDK.

# How to Test

from the root of the repository you should run `./gradlew install`.  The build should complete without errors and you should have in your loval maven repository (ex ~/.m2/repository) two new artifacts: "org.aerogear.core" and "org.aerogear.keycloak-module".

This PR includes bits from #4, but once that is merged this can be rebased and merged.

PR #6, opened 2018-01-14
Agdroid 693 parse mobile core

secondsun:
This add support for parsing a mobile-core.json file from the mobile cli tool.

# How to test

run `./gradlew testDebug` from the project root.  You should see in your gradle logs that the parsing tests for the example mobile-core.json file are run.

#Notes
This depends on PRs #4 and #5   Should we make changes to those we will need to rebase this PR.

### DIFF
--- a/core/src/debug/assets/mobile-core.json
+++ b/core/src/debug/assets/mobile-core.json
@@ -1,0 +1,46 @@
+{
+  "services": [
+    {
+      "config": {
+        "headers": {},
+        "uri": "https://fh-sync-server-myproject.192.168.37.1.nip.io"
+      },
+      "name": "fh-sync-server"
+    },
+    {
+      "config": {
+        "auth-server-url": "http://keycloak-myproject.192.168.37.1.nip.io/auth",
+        "clientId": "OeFYkwEuUKXvDsSudfXc",
+        "headers": {},
+        "realm": "myproject",
+        "resource": "OeFYkwEuUKXvDsSudfXc",
+        "ssl-required": "external",
+        "url": "http://keycloak-myproject.192.168.37.1.nip.io/auth"
+      },
+      "name": "keycloak"
+    },
+    {
+      "config": {
+        "headers": {},
+        "name": "prometheus",
+        "type": "prometheus",
+        "uri": "https://prometheus-myproject.192.168.37.1.nip.io"
+      },
+      "name": "prometheus"
+    },
+    {
+      "config": {
+        "headers": {},
+        "mysql_db": "unifiedpush",
+        "mysql_password": "unifiedpush",
+        "mysql_username": "unifiedpush",
+        "name": "unified-push-server",
+        "push_kc_realm": "aerogear",
+        "type": "unified-push-server",
+        "uri": "http://ups-myproject.192.168.37.1.nip.io"
+      },
+      "name": "unified-push-server"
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -0,0 +1,103 @@
...
+        return (String message, Exception e) -> {
+            Log.e("MOBILE_CORE", message, e);
+        };
+    }
+
+    @Override
+    public void bootstrap(Object... args) {
+
+
+        try (InputStream configStream = context.getAssets().open("mobile-core.json");) {
```
wei-lee: make the path to the JSON file configurable. Use `mobile-core.json` the default value.

core/src/debug/assets/mobile-core.json
```diff
@@ -0,0 +1,46 @@
+{
```
wtrocki: @secondsun Let not commit that here. 
We may just have that as part of unit tests and write unit tests?

I think that initial name suggested was  `mobile-services.json`, but I think that best will be to not hardcode it.
We can have constant or build time variable for that.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
wtrocki: mobile-services.json as default name will make more sense. It contains info about services.

core/src/debug/assets/mobile-core.json
wtrocki: Actually best to remove all assets and test it in top level example project (we will need to have it here to test if this is usable`. WDYT?

From android docs:
`Library projects cannot include assets`

wtrocki:
@secondsun To verify I will need to have really simple android project that will:
- include core library
- include sample services json (move sample from library itself)
- display services in activity for testing purposes.
- We can then extend this sample project to make network request to service (ping etc.)
in order to test functionalities on real app. 

WDYT about this idea? I'm happy to code that app to verify your changes.

wtrocki:
@secondsun Happy to contribute that on top of your PR.

core/src/debug/assets/mobile-core.json
secondsun: Testing in a top level project has the problem that that project will have to be run in an emulator which we are trying to avoid.

secondsun: Yes, this file is just for unit tests.

In Android I had problems getting the unit test to pick up the asset unless it was in the debug source set.  This seems to be a common work around and, since it didn't get bundled in the final artifact, I didn't wsee a problem.

secondsun:
@wtrocki I can bang out an example like that in about thirty minutes.

secondsun:
@wtrocki @wei-lee I've implemented your feedback.

core/src/debug/assets/mobile-core.json
wtrocki: I think this file can be removed as we have renamed it and it's now duplicated. Is that right?

wtrocki:
@wei-lee Is that ok with you as well?
@secondsun - We can add example using separate PR.

wtrocki:
Merged as this change were approved and were blocking other changes from other PR.

PR #7, opened 2018-01-16
AGDROID-698 add initial docs

pb82:
Adds a directory for the preliminary SDK docs and a topic on how to use the BOM

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,9 @@
+# AeroGear Android SDK Documentation
+
+This directory contains the preliminary documentation for the AeroGear Android SDK.
+
+## Topics
+
+1. [Using the Android SDK BOM in your projects](./using_the_android_sdk_bom.md)
+1.1 [Concept](./using_the_android_sdk_bom.md#Concept)
+1.2 [Usage](./using_the_android_sdk_bom.md#Usage)
--- a/docs/using_the_android_sdk_bom.md
+++ b/docs/using_the_android_sdk_bom.md
@@ -1,0 +1,50 @@
+# Using the Android SDK BOM in your own projects
+
+## Concept
+
+A BOM (Bill of Materials) is used to lock the dependencies (direct and transitive) down to known compatible versions. Gradle itself does not offer such a mechanism but it is a well known and used concept in the Maven world (see [Dependency Management](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management)).
+In order to bring this functionality to Gradle the Android SDK relies on the [dependency management plugin](https://github.com/spring-gradle-plugins/dependency-management-plugin) provided by Spring. This plugin allows us to use a standard Maven BOM in a Gradle project.
+
+## Usage
+
+If your project depends on the AeroGear Android SDK you can make use of the same BOM to ensure that your dependencies are compatible with the SDK. All you need to do is to apply the plugin and reference the BOM.
+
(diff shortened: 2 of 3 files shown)


docs/README.md
```diff
@@ -0,0 +1,7 @@
+# AeroGear Android SDK Documentation
+
+This directory contains the preliminary documentation for the AeroGear Android SDK and will extended during development.
```
wtrocki: `and will extended during development.` This may be not needed as we will need to remove it after.

docs/using_the_android_sdk_bom.md
```diff
@@ -0,0 +1,38 @@
+# Using the Android SDK BOM in your own projects
```
wtrocki: Content is really good but I will just reformat that to 2 headers:

1) (Concept) Using the maven BOM or something like above.
Explain here what we using etc.

2) (Procedure) Add SDK SDK
It's cleaner and faster for reader to just jump in and copy paste steps.  

This is pretty standard way how docs is formatted.

docs/using_the_android_sdk_bom.md
```diff
@@ -0,0 +1,38 @@
...
+
+A BOM (Bill of Materials) is used to lock the dependencies (direct and transitive) down to known compatible versions. Gradle itself does not offer such a mechanism but it is a well known and used concept in the Maven world (see [Dependency Management](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management)).
+In order to bring this functionality to Gradle the Android SDK relies on the [dependency management plugin](https://github.com/spring-gradle-plugins/dependency-management-plugin) provided by Spring. This plugin allows us to use a standard Maven BOM in a Gradle project.
+
+If your project depends on the AeroGear Android SDK you can make use of the same BOM to ensure that your dependencies are compatible with the SDK. All you need to do is to apply the plugin and reference the BOM.
+
+At the top of your gradle build file (but below `buildscripts`) add the plugin itself:
```
wtrocki: I think that it will be worth to mention mobile services here and finish this procedure by importing core and sync service.

docs/using_the_android_sdk_bom.md
```diff
@@ -0,0 +1,38 @@
...
+subprojects {
+    apply plugin: 'io.spring.dependency-management'
+
+    dependencyManagement {
+        imports {
+            mavenBom 'org.jboss.aerogear:aerogear-android-sdk-bom:<latest version>'
+        }
+    }
+}
+```
```
wtrocki: Missing example with dependency.

docs/using_the_android_sdk_bom.md
paolobueno: Might be nice to say it's the `build.gradle` file, just to be explicit about the name

pb82: I didn't want to give this example because they do not yet exist and artifact names could change

pb82: But you're right that it would be good to show how to import a dependency without specifying a version

docs/using_the_android_sdk_bom.md
pb82: Done 👍

docs/using_the_android_sdk_bom.md
pb82: Done, added a usage example for a package we will definitely use (okhttp)

PR #10, opened 2018-01-17
Initial (conceptual) version of simple core POC

wtrocki:
Alternative proposal for Mobile SDK architecture. 
Proposition based on interfaces and constructor based DI.

For original proposal see: https://github.com/aerogear/aerogear-android-sdk/pull/2

## Explanation

Core exposes single interface that is being used in the services. 
Interface has methods that provide core services.
Core interface is defined by individual core services (like networking) interfaces etc.

Core doesn't need know anything about other gradle modules that implement top level SDK.
This aproach is clean, simply to understand - basically how people do java for long time. 
By using interfaces we can make sure that we can provide compatibility layer between services.

End user Api will hide complexity of core - Users will need to deal only with creation of the top level SDK object. Community can create new services without much documentation and explanation and use core services by simply reviewing top level core interface - or using intelisense. 

## Why 

This aproach has numerous advantages:

- End user api is cleaner and not exposing any implementation details to end users.
```
new SyncSdk(context).performSync();
```

Both Dagger and our custom DI solutions require significant knowledge and codebase on user side:
https://github.com/secondsun/WipDemo/blob/master/app/src/main/java/org/feedhenry/mcp/prdemo/MainActivity.java#L29-L39

- It's provides a lots of flexibility in terms of how we can use our services internally.
- Thanks to interface we can handle versions properly and support different versions of core.
- It's not requiring any external dependencies
- It is relaying on language standards. 

This implementation will be similar to what dagger will offer, but it is just relaying on constructor injection or direct initialization in the service.

Happy to explain that in more details.

## Notes 

This is just a proof of concept. I will be happy to explain how we can extend that to provide keycloak security using OkHttp interceptors etc. If accepted I will be happy to provide real end to end example that will work with services:

- Sync service
- Protection by keycloak
- Core with networking, configuration etc.

**This PR is only for showing idea. If we decide to go this way I will close that and provide full end to end implementation that will work with the sync library"**


core/src/main/java/org/aerogear/mobile/core/api/SdkCore.java
```diff
@@ -0,0 +1,74 @@
...
+import org.aerogear.mobile.core.http.NetworkApi;
+import org.aerogear.mobile.core.logging.Logger;
+import org.json.JSONException;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Map;
+import java.util.concurrent.Executor;
+
+public class SdkCore implements Core {
```
wtrocki: Ideally this class should not have any code and it should just provide implementations for core services.

`getNetworkLayer` can be renamed to provideNetworkLayer etc. and return network interface. 
The same with configuration but to simplify this I have made implementations directly in this class

wei-lee:
nice work @wtrocki. If I understand correctly,  I think the biggest difference between this approach and @secondsun's is wether we need the `ServiceRegistry` and who is responsible to instantiate all the service modules. 

In this approach:
* we do not need to maintain a service registry
* developers will be responsible to instantiate the service modules they need
* the service modules do not need to implement the same interface

So I think the biggest question is do we need the `ServiceRegistry` and how are we suppose to use it.  

Based on the fact that:
1. we don't have many services to work with at the moment
2. we don't want to have hard-dependencies between service modules (no need to manage dependencies between service modules)

I can't really think of a good use case of the `ServiceRegistry` at the moment. We can add it back later if we find there is a use case for it. But for now it's better to keep it simple.

So, from my point of view, this approach is a lot simpler and cleaners, and easier for developers to understand and use. So it is a +1 from me.

sync/src/main/java/eu/metawired/wtrocki/sync/SyncSdk.java
```diff
@@ -0,0 +1,32 @@
...
+ * Need to be used in top level app:
+ *
+ * new SyncSdk(context).performSync();
+ */
+public class SyncSdk {
+
+    private Core core;
+
+    public SyncSdk(Context context) {
+        this.core = new SdkCore(context);
```
secondsun: So does this mean that each ServiceModule has its own instance of core?  If so I don't really like that because this means that for every service module a user is using they are also making new instances of the thread pools, network resources, etc.  While it probably won't be a problem in practice, I think that SDK Core should be a shared resource.

wei-lee: I am pretty sure we can make the core singleton. Another option is to instantiate Core once and pass it to each of the service module.

secondsun:
So I think that the service registry will be something we need in the future because most of these components will have non trivial configuration and resource needs.  I think in the future we will.

However, "Here's a shiny new box of code now, now configure the DI mechanism" is a pain in the ass that I don't want to inflict on our users which the registry does in its current state.

The block of core @wtrocki linked really isn't representative of the final goal of SR but its current state.  Ideally the following
```
ServiceModuleRegistry registry = new ServiceModuleRegistry();
        registry.registerServiceModule("fh-sync-server", LocalTestServiceModule.class);
        registry.registerServiceModule("keycloak", LocalTestServiceModule.class);
        registry.registerServiceModule("prometheus", LocalTestServiceModule.class);
        registry.registerServiceModule("unified-push-server", LocalTestServiceModule.class);


        MobileCore core = new MobileCore.Builder(this.getApplicationContext())
                                            .setRegistryService(registry)
                                            .setMobileServiceFileName("mobile-core.json")
                                            .build();
```java

would more closely look like 
```
@Inject
ServiceRegistry serviceRegiststry;

serviceRegistry.registerServiceModule("my-application", MyApplicationServiceModule.class, "keycloak","promethius","http");//[1]

MobileCore core = new MobileCore.Builder(this.getApplicationContext())
                                            .build();

core.getService("my-application").doStuff();
```java

as MobileCore would be able to identify ServiceModule's in the mobile-service.json file by the "type" or "name" parameter.  

@wtrocki From your code so far similar code would look like (from my I have not had coffee POV): 

```
SdkCore core = new SdkCore(this.getApplicationContext());
AuthService auth = new KeyCloakService(core);
SyncService sync = new FHSyncService(core, keycloakService);
MetricsService metrics = new PromethiusService(core);

MyApplicationPresenter presenter = new MyApplicationPresenter(core, auth,sync,metrics);
presenter.doStuff();
```java

File: sync/src/main/java/eu/metawired/wtrocki/sync/SyncSdk.java (hunk 2 above)
Comment (wtrocki) [reply sync/src/main/java/eu/metawired/wtrocki/sync/SyncSdk.java]: There are two variants of this. One is using constructor DI second one initializes core inside. I pointed out both and agree that constructor based aproach will be better way to go (config and some initialization is happening once. Problem with that is that network based layer will need to be shared with the services, so that's something I would like to talk about in terms of interceptors etc. 

TL;DR - service initialization means service specific seed data, 
Global initialization for core will mean some global handlers.

Comment (david-martin) [reply sync/src/main/java/eu/metawired/wtrocki/sync/SyncSdk.java]: Is there a way to avoid the developer having to manually inject a Core instance into each SDK they want to use?
Looking at Firebase as an example https://firebase.google.com/docs/database/android/start/ it seems you only need to:
- include the json defition file
- firebase dependencies

then you can do this:
```java
FirebaseDatabase database = FirebaseDatabase.getInstance();
DatabaseReference myRef = database.getReference("message");

myRef.setValue("Hello, World!");
```

That seems like a nice goal for our SDK. The develop only needs the mobile-config.json and the SDK dependencies, then they can get an instance of whatever SDK they want and use it.

paolobueno: I think we need some clearing up on how services depending on others that are not the Core would look like. It wouldn't be safe to assume that we won't ever have dependencies between them.

We might need to not cater to diamond-style dependencies and avoid those ourselves, though.

secondsun: @david-martin @paolobueno I added some thoughts here : https://github.com/aerogear/aerogear-android-sdk/pull/10#issuecomment-358636869

paolobueno:
@secondsun in that example would `core.getService("my-application").doStuff();` be strongly typed somehow?

sync/src/main/java/eu/metawired/wtrocki/sync/SyncSdk.java
wtrocki: The whole idea is to not initialize services in core - just to have core injected.
Having core singleton is possible and definitely way to go.

>  It wouldn't be safe to assume that we won't ever have dependencies between them.

That will be huge problem and general anti-pattern.

david-martin:
> serviceRegistry.registerServiceModule("my-application", MyApplicationServiceModule.class, "keycloak","promethius","http");//[1]

@secondsun Would it be possible for Core to automatically detect what SDK's are included in the App and register automatically?

secondsun:
@paolobueno Yes.  I'm not sure the "best" way to do it but there are several options and they are all more verbose than my stub.  

@david-martin That is the implication.  As long as the "type" parameter is set and known in advance we can.

david-martin:
> @david-martin That is the implication. As long as the "type" parameter is set and known in advance we can.

@secondsun What would/should a developer have to set the type? Don't 'we' know what SDK class will be used for each service, and handle any init or inter-dependcies for the developer?

sync/src/main/java/eu/metawired/wtrocki/sync/SyncSdk.java
paolobueno: "everything's a singleton" seems very nice too. Considering caching done for commonjs modules I think this would become very idiomatic for the cordova sdk too.

wtrocki:
TL;DR - This aproach matches better with Dagger2 running on the user app when they will want to specify providers on the top level application. 

Summers aproach has custom DI (something like Dagger2) that leaks complexity to the end user app and is less flexible. 

As android developer without any knowledge of the library I would expect any SDK to give me choice how I do dependency injection. 

When designing this aproach I focused on:
- Standards in Android - How android developers will expect this to work
- Flexibility for developers how objects are created **Really important**
If we using dagger users can just provider. Users can use static methods on Application level.
- Simplification of user based api
- Clear path and exposure what is our end user api and what is our service based api
- Hide complexity of the internal part of the SDK to SDK minimize amount of knowledge required to use it. 
- Do not expose way we bind things in end user API - as this may change and then we will need to break end user apis. 

I might be lacking some facts here - but I think that with aproach in this PR we getting best deal in terms of flexibility, future support and developer experience, while not overthinking how future API will look like. 

We can still use dagger 2.0 inside or even service registry but users will not need to deal or be aware of that in their end user API. 

If we really want to use @annotation based DI we can introduce this at any time later, by providing binding directly on service level (for the moment is just overkill for module developers to understand and maintain that - IMHO)

aliok:
#### About the creation of objects and DI:
> If we want to build SDK's as android developer without any knowledge of the library I would expect any SDK to give me choice how I do dependency injection and do not change my development workflow.

+1 on this .  Since beginning of this work, I was always thinking "can't we use something like Guice?". With keeping things super simple, we could let users pick their own DI tool.

However, there also should be dead simple way of getting everything instantiated w/o configuring the DI tool. Like a default builder/constructor/whatever.

secondsun:
> @secondsun What would/should a developer have to set the type? Don't 'we' know what SDK class will be used for each service, and handle any init or inter-dependcies for the developer?

@david-martin The type should be set by mobile-core when the service is provisioned in OpenShift.  The Core SDK should parse the mobile-services.json file and make a service for each <name, type> tuple.  

Services can register themselves to the serviceregistry in their static initialization blocks, or they can be provided to the core SDK through some other mechanism that the end user developer does not have to be aware of for most common cases.

I can go into much more details.  At 10 EST I have a meeting for Pair Programming where wei and woiciech and I have been discussing and developing these mechanisms this week.  Feel free to join on my bluejeans (ask me off channel for the url)

secondsun:
> About the creation of objects and DI:
>>If we want to build SDK's as android developer without any knowledge of the library I would expect any SDK to give me choice how I do dependency injection and do not change my development workflow.
>
>+1 on this . Since beginning of this work, I was always thinking "can't we use something like Guice?". With keeping things super simple, we could let users pick their own DI tool.
>
>However, there also should be dead simple way of getting everything instantiated w/o configuring the DI tool. Like a default builder/constructor/whatever.

@aliok That is the ultimate goal of the ServiceRegistry as well.

wtrocki:
Closing as we going to use different aproach that will be posted as separate PR.

PR #11, opened 2018-01-18
Gradle changes

ziccardi:


build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:3.0.1'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:2.0'
-
+        classpath 'net.openid:appauth:0.7.0'
```
wei-lee: I don't think this should be at the top level build.gradle file. This is only required by the auth module, so I say move it to the `auth/build.gradle` file and it should be an implementation dependency.

auth/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="org.aerogear.android.core" />
```
danielpassos: s/package="org.aerogear.android.core"/package="org.aerogear.android.auth" ?

core/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,2 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="org.aerogear.android.core" />
+    package="org.aerogear.auth" />
```
danielpassos: Why core package was changed?

build.gradle
secondsun: Just +1ing wei's comment.  Additionally this block is the dependencies for building (think devDependencies from node.js) as opposed to dependencies for the code.

secondsun:
@ziccardi bump

danielpassos:
@ziccardi Can you rebase and fix the conflict?

secondsun:
So right now there is only one line in this file, is it safe to assume that these changes are part of another PR that has been merged?

wtrocki:
Let's close this as it's surely outdated.

ziccardi:
Yes, closing it now

PR #13, opened 2018-01-19
Implement auth state manager for persisting

aidenkeating:
Renamed the TokenPersistor to AuthStateManager as this name seems to be closer to what the class will do. Added a few tests to cover some small stuff, tried not to go to far and end up testing `SharedPreferences` as this addition is really just a thin wrapper around that.

Not at all familiar with Java so there may be some weird stuff.

@wei-lee @ziccardi @TommyJ1994 Would you mind taking a look?


auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
```diff
@@ -0,0 +1,60 @@
...
+import android.content.SharedPreferences;
+import org.aerogear.auth.credentials.OIDCCredentials;
+import org.json.JSONException;
+
+/**
+ * Saves, retrieves and delete a token.
+ */
+public class AuthStateManager {
+
+    private static final String STORE_NAME = "AuthState";
```
wei-lee: we might want to namespace the store name here to avoid conflicts. One option could be the package name

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
```diff
@@ -0,0 +1,60 @@
...
+            if(!mPrefs.edit().putString(KEY_STATE, authState.serialise()).commit()) {
+                throw new IllegalStateException("Failed to update state from shared preferences");
+            }
+        }
+    }
+
+    /**
+     * Deletes a token
+     */
+    public void clear() {
```
wei-lee: also to support multiple threads to read and write the auth state, you might want to consider using a lock for write and clear operations.

TommyJ1994:
Looks good.

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -1,40 +1,35 @@
...
 /**
  * Credentials for OIDC based authentication
  */
-public final class OIDCCredentials implements ICredential {
+public class OIDCCredentials implements ICredential {
 
-    private OIDCToken identityToken;
-    private OIDCToken accessToken;
-    private OIDCToken refreshToken;
+    private AuthState authState;
```
ziccardi: This field is initialized only in constructors. Should be final.

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Saves, retrieves and delete a token.
+ */
+public class AuthStateManager {
+
+    private static final String STORE_NAME = "org.aerogear.android.auth.AuthState";
+    private static final String KEY_STATE = "state";
+
+    private SharedPreferences mPrefs;
```
ziccardi: This is initialised only in constructors. Should be final.
Do we use the 'm' prefix for member variables?

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public AuthStateManager(Context context) {
```
ziccardi: It is a good practice to make the method parameter 'final'

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
```diff
@@ -0,0 +1,60 @@
...
+        } catch (JSONException ex) {
+            return new OIDCCredentials();
+        }
+    }
+
+    /**
+     * Saves a token
+     * @param authState token to be saved
+     */
+    public void write(OIDCCredentials authState) {
```
ziccardi: See previous comment about method parameters

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+    }
+
+    /**
+     * Deletes a token
+     */
+    public void clear() {
+        if (!mPrefs.edit().remove(KEY_STATE).commit()) {
+            throw new IllegalStateException("Failed to clear state from shared preferences");
```
ziccardi: Is it correct to throw an unchecked exception here?

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
aidenkeating: @wei-lee I've declared the methods synchronized in this commit https://github.com/aerogear/aerogear-android-sdk/pull/13/commits/736d254e4834dbd93a02054d95e04971174c9465

aidenkeating:
@ziccardi Would you mind taking another look? I _think_ I covered everything

PR #14, opened 2018-01-23
AGDROID-682 Include checkAuthStatus

aidenkeating:
## Motivation

Allow checking for current auth status. Both is authorized and token isn't expired.

JIRA: https://issues.jboss.org/browse/AGDROID-682

## Description

Wraps AuthState some more with OIDCCredentials, mainly wrapping `isAuthorized()` and `getNeedsTokenRefresh()`.

Check these to determine the current auth state using `checkAuthState()`.

## Progress

- [x] Check if authorized
- [x] Check if needs refresh (is expired)
- [x] Unit tests (Everything is a thin wrapper around AppAuth Android)

### DIFF
--- a/auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
+++ b/auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
@@ -44,7 +44,7 @@ public class UserPrincipalImpl extends AbstractPrincipal {
      * @param roles roles assigned to the user
      * @param authenticator the authenticator that authenticated this user
      */
-    private UserPrincipalImpl(final String username,
+    protected UserPrincipalImpl(final String username,
                               final ICredential credentials,
                               final String email,
                               final Map<String, IRole> roles,
@@ -60,13 +60,13 @@ public class UserPrincipalImpl extends AbstractPrincipal {
      * Builds and return a UserPrincipalImpl object
      */
     static class Builder {
-        private String username;
-        private String email;
-        private Map<String, IRole> roles = new HashMap<>();
-        private AbstractAuthenticator authenticator;
+        protected String username;
+        protected String email;
+        protected Map<String, IRole> roles = new HashMap<>();
+        protected AbstractAuthenticator authenticator;
         protected ICredential credentials;
 
-        private Builder() {
+        protected Builder() {
         }
 
         Builder withUsername(final String username) {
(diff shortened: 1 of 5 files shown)


aidenkeating:
@TommyJ1994 Would you mind taking a look? Still a bit more work to be done, just sanity checking the current work

TommyJ1994:
Looks good so far :+1:

aidenkeating:
@TommyJ1994 Only coming back to this now. I'm not sure the value testing would add here as we're wrapping AppAuth very closely. Is there anything else this check needs to do?

@ziccardi Would you mind taking a look also?

TommyJ1994:
@aidenkeating Agreed, its a light wrapper so I think we can do without tests.

ziccardi:
The only comment I have about this, is about adding the get/set methods and the isAuthorized method.
The ICredential interface is a generic interface that could be used for other type of authentication as well. I think it would be better to move those methods the OIDCCredential class

aidenkeating:
@ziccardi Just to confirm, you mean changing UserPrincipalImpl to use OIDCCredential instead of the generic ICredential? https://github.com/aidenkeating/aerogear-android-sdk/blob/a63f7db58971b61d6fd55d8a9ad1ae482db9a9da/auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java#L37 If so, that sounds good to me, will make that change

ziccardi:
@aidenkeating No, I just mean moving those methods to the OIDCCretential object.
`UserPrincipalImpl` should always store an `ICredential` object so that different kind of credentials can be stored.
Since OIDCCredentials implements the ICredential object, `UserPrincipalImpl` is able to store an `OIDCCredential` object without any change

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -137,14 +137,22 @@ public String getName() {
         return username;
     }
 
+    /**
+     * Check whether the user is authorized and not expired.
+     * @return <code>true</code> if user is authorized and token is not expired.
+     */
+    @Override
+    public boolean checkValidAuth() {
```
ziccardi: I think it would be better to add this method to the `credentials` object so that the `UserPrincipalImpl` remains general purpose. WDYT?

aidenkeating: Ah I understand what you mean now, yep that makes sense to me

aidenkeating:
@ziccardi Would you mind taking another look? Moved `checkAuthState` into the credentials and also generalised the credential method names a bit too.

aidenkeating:
@ziccardi Mind taking another look?

PR #15, opened 2018-01-23
[WIP] - implement user get roles and has roles functionality

rachael-oregan:
## Motivation

JIRA: https://issues.jboss.org/browse/AGDROID-684

## Description

When a user is created an authenticator is passed to the User Object.  Inside this authenticator Object (`OIDCAuthCodeImpl`) is where the users data/identity is decoded from their credentials.  This decoded data is then used to `build` the user Object.

The `getRoles` function for the User should return a collection of Roles.

Note: `hasRole` function implementation has been included in this PR too.  See this [ticket](https://issues.jboss.org/browse/AGDROID-685) for more info.

## Progress

- [x] Implemented get user roles functionality
- [x] Unit Testing
- [x] JavaDocs


rachael-oregan:
@TommyJ1994 @aidenkeating @ciaranRoche @wei-lee @ziccardi @matzew mind having a look please

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -20,6 +43,111 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
+                String clientRolesString = tokenClientRolesJSON.substring(1, tokenClientRolesJSON.length() - 1);
+                String roles[] = clientRolesString.split(COMMA);
+                clientRoles = Arrays.asList(roles);
+            }
+        }
+        return clientRoles;
+    }
+
+    private JSONObject getIdentityInformation(final ICredential credential) throws JSONException {
+        String accessToken = credential.getAccessToken();
```
aidenkeating: This won't compile will it? ICredential doesn't declare a `getAccessToken` function

rachael-oregan: Ya you're right.  There' another compile error too. Nice catch, thanks

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -85,19 +84,11 @@ Builder withEmail(final String email) {
...
-        Builder withRoles(final IRole[] roles) {
-            if (roles != null) {
-                return withRoles(Arrays.asList(roles));
-            } else {
-                return this;
-            }
-        }
-
-        Builder withRoles(final Collection<IRole> roles) {
+        Builder withRoles(final List<IRole> roles) {
```
aidenkeating: This looks to be causing a compile time error.

```
/Users/aidenkeating/work/aerogear-android-sdk/auth/src/main/java/org/aerogear/auth/impl/GuestAuthenticatorImpl.java:54: error: incompatible types: IRole[] cannot be converted to List<IRole>
                .withRoles(roles)
```

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -20,6 +43,111 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
+            String tokenRealmRolesJSON = userIdentity.getJSONObject(REALM).getString(ROLES);
+
+            String realmRolesString = tokenRealmRolesJSON.substring(1, tokenRealmRolesJSON.length() - 1); //remove the array brackets
+            String roles[] = realmRolesString.split(COMMA);
+            realmRoles = Arrays.asList(roles);
+        }
+        return realmRoles;
+    }
+
+    private List<ClientRole> getClientRoles() throws JSONException {
```
aidenkeating: Should we be doing all this parsing at "build time"?

To me, this should be done on build and the `clientRoles` and `realmRoles` variables set, and `getClientRoles` and `getRealmRoles` should just return those values instead of performing the parsing here? 

It's a bit confusing when a getter throws a JSONException, to me. wdyt?

rachael-oregan:
@ziccardi am I correct in presuming that the credentials passed into this `authenticate` function [here](https://github.com/aerogear/aerogear-android-sdk/pull/15/files#diff-55c39efb3165e4a3b37dbe595456384cR45) are the credentials that will contain the users access token and identity token etc?

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
aidenkeating: Edit: I was misreading this block of code

auth/src/main/java/org/aerogear/auth/AuthServiceConfig.java
```diff
@@ -1,8 +1,15 @@
...
+
 /**
  * Authentication service configuration
  */
 public class AuthServiceConfig {
     // TODO: to be implemented.
+
+    public JSONObject toJSON() {
+        // TODO: to be implemented.
+        return null;
```
ziccardi: Please, throw an IllegalStateException("Not implemented") here instead of returning null.

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -20,6 +43,112 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
+        if (userIdentity != null) {
+            // get the users email
+            if (userIdentity.has(EMAIL) && userIdentity.getString(EMAIL).length() > 0) {
+                emailAddress = userIdentity.getString(EMAIL);
+            }
+        }
+        return emailAddress;
+    }
+
+    private List<IRole> parseRoles() throws JSONException {
```
ziccardi: I would suggest returning a collection instead of a List, so that we can freely change the implementation without affecting the caller

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -20,6 +43,112 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
+            }
+            List<ClientRole> clientRoles = parseClientRoles();
+            if (clientRoles != null) {
+                roles.addAll(clientRoles);
+            }
+        }
+        return roles;
+    }
+
+    private List<RealmRole> parseRealmRoles() throws JSONException {
```
ziccardi: I would suggest returning a collection instead of a List, so that we can freely change the implementation without affecting the caller

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -20,6 +43,112 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
+            String tokenRealmRolesJSON = userIdentity.getJSONObject(REALM).getString(ROLES);
+
+            String realmRolesString = tokenRealmRolesJSON.substring(1, tokenRealmRolesJSON.length() - 1); //remove the array brackets
+            String roles[] = realmRolesString.split(COMMA);
+            realmRoles = Arrays.asList(roles);
+        }
+        return realmRoles;
+    }
+
+    private List<ClientRole> parseClientRoles() throws JSONException {
```
ziccardi: I would suggest returning a collection instead of a List, so that we can freely change the implementation without affecting the caller

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -29,7 +27,7 @@
     /**
      * Roles associated with this principal.
      */
-    private final Map<String, IRole> roles;
+    private final List<IRole> roles;
```
ziccardi: I think saving the roles inside a map would be more efficient for the 'hasRole' methods.

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -47,12 +45,12 @@
...
                               final ICredential credentials,
                               final String email,
-                              final Map<String, IRole> roles,
+                              final List<IRole> roles,
                               final AbstractAuthenticator authenticator) {
         super(authenticator);
         this.username = username;
         this.email = email;
-        this.roles = Collections.unmodifiableMap(new HashMap<String, IRole>(roles));
+        this.roles = roles;
```
ziccardi: Even if we want to use a List instead of a Map, we should copy the List instead of simply assigning it.
If we simply assign, the caller still has a reference to the List and could change the List content corrupting the `UserPrincipalImpl` content.
The copy should be `unmodifiable`.

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -62,7 +60,7 @@ private UserPrincipalImpl(final String username,
     static class Builder {
         private String username;
         private String email;
-        private Map<String, IRole> roles = new HashMap<>();
+        private List<IRole> roles = new ArrayList<>();
```
ziccardi: See previous comment

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -129,7 +119,9 @@ UserPrincipalImpl build() {
      */
     @Override
     public boolean hasRole(final IRole role) {
-        return roles.containsKey(role.getRoleName());
+//        TODO: to be implemented
+//        return roles.containsKey(role.getRoleName());
+        return false;
```
ziccardi: If a method is not implemented, it is better to throw an exception than returning a wrong value.

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
rachael-oregan: The issue with this is that rolenames are not unique.  Role's can have the same name at the realm and client level so I'm not sure this would work?

ziccardi: A Key object with `role name` and `role type` should be implemented.
Then the map should become `private final Map <RoleKey, IRole> roles`.

In the `RoleKey` object you should override the `equals` method so that both role name and role type gets compared. Then it should work.

ziccardi:
> am I correct in presuming that the credentials passed into this authenticate function here are the credentials that will contain the users access token and identity token etc?

Yes

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,24 @@
...
+    public RoleKey(IRole role) {
+        roleName = role.getRoleName();
+        roleType = role instanceof RealmRole ? RoleType.REALM : RoleType.CLIENT;
+    }
+
+    @Override
+    public boolean equals(Object roleKey) {
+        if (this == roleKey) return true;
+        if (roleKey == null || roleKey.getClass() != getClass()) return false;
+        return ((RoleKey) roleKey).roleName == roleName && ((RoleKey) roleKey).roleType == roleType;
```
ziccardi: Is the role name case sensitive?

ziccardi: Strings must be compared using the `equals` method

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Created by rachael on 25/01/2018.
+ */
+
+public class RoleKey {
+
+    private String roleName;
+    private RoleType roleType;
+
+    public RoleKey(IRole role) {
```
ziccardi: Parameter should be final

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,24 @@
...
+        roleName = role.getRoleName();
+        roleType = role instanceof RealmRole ? RoleType.REALM : RoleType.CLIENT;
+    }
+
+    @Override
+    public boolean equals(Object roleKey) {
```
ziccardi: Parameter should be final

auth/src/main/java/org/aerogear/auth/RoleType.java
```diff
@@ -0,0 +1,9 @@
+package org.aerogear.auth;
+
+/**
+ * Created by rachael on 25/01/2018.
```
ziccardi: Replace default comment

auth/src/main/java/org/aerogear/auth/impl/GuestAuthenticatorImpl.java
```diff
@@ -22,21 +25,20 @@
...
-    private final IRole[] roles;
+    private final Map<RoleKey, IRole> roles;
 
     /**
      * Builds a new guest authenticator object
      * @param guestUser the user to be returned after the 'authentication'
      * @param roles the roles to be assigned to the user
      */
-    public GuestAuthenticatorImpl(final String guestUser, final IRole[] roles) {
+    public GuestAuthenticatorImpl(final String guestUser, final Map<RoleKey, IRole> roles) {
```
ziccardi: I think an array or collection of roles as parameter would be better. The map can be constructed inside the class.

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -20,6 +43,120 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
+        if (userIdentity != null) {
+            // get the users email
+            if (userIdentity.has(EMAIL) && userIdentity.getString(EMAIL).length() > 0) {
+                emailAddress = userIdentity.getString(EMAIL);
+            }
+        }
+        return emailAddress;
+    }
+
+    private Map<RoleKey, IRole> parseRoles() throws JSONException {
```
pb82: Is `IRole` an interface? I could be wrong but i thought the `I` prefix for interfaces is more of a .Net thing and not common in Java. @ziccardi @matzew what do you think?

auth/src/main/java/org/aerogear/auth/credentials/ICredential.java
```diff
@@ -4,4 +4,5 @@
  * Base interface for credential objects.
  */
 public interface ICredential {
+    public String getAccessToken();
```
ziccardi: This is only about OIDC authentication, so it should be only inside the `OIDCCredential` object

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -20,6 +43,120 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
+                    ClientRole clientRole = new ClientRole(rolename);
+                    clientRoles.put(new RoleKey(clientRole), clientRole);
+                }
+            }
+        }
+        return clientRoles;
+    }
+
+    private JSONObject getIdentityInformation(final ICredential credential) throws JSONException {
+        String accessToken = credential.getAccessToken();
```
ziccardi: A cast to `OIDCCredential` is needed here.
A check that `ICredential` is an instance of `OIDCCredential` would be a good idea. If it is not, an IllegalArgumentException should be raised

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -20,6 +43,120 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
+
+        try {
+            // Decode the Access Token to Extract the Identity Information
+            String[] splitToken = accessToken.split("\\.");
+            byte[] decodedBytes = Base64.decode(splitToken[1], Base64.URL_SAFE);
+            String decoded = new String(decodedBytes, "UTF-8");
+            try {
+                decodedIdentityData = new JSONObject(decoded);
+            } catch (JSONException e) {
+                e.printStackTrace();
```
ziccardi: Throw an exception here, passing the JSONException as root exception.
There is an AuthenticationException that could be used here.

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -85,13 +85,14 @@ Builder withEmail(final String email) {
...
 
         Builder withRole(final IRole role) {
-            this.roles.put(role.getRoleName(), role);
+            RoleKey roleKey = new RoleKey(role);
+            this.roles.put(roleKey, role);
             return this;
         }
 
-        Builder withRoles(final IRole[] roles) {
+        Builder withRoles(final Map<RoleKey, IRole> roles) {
```
ziccardi: This should take an array of roles. The `Map` should be populated internally. The `RoleKey` object is not to be seen by the caller.

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -129,7 +130,8 @@ UserPrincipalImpl build() {
      */
     @Override
     public boolean hasRole(final IRole role) {
-        return roles.containsKey(role.getRoleName());
+        // TODO: to be implemented
+        throw new IllegalStateException("Not implemented");
```
ziccardi: The containsKey can be used here

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+public class RoleKey {
+
+    private String roleName;
+    private RoleType roleType;
+
+    public RoleKey(IRole role) {
+        roleName = role.getRoleName();
+        roleType = role instanceof RealmRole ? RoleType.REALM : RoleType.CLIENT;
```
ziccardi: Instead of checking object class, wouldn't be better to add a 'RoleType getType()' method in the `IRole` interface and implement it in each role class, returning the appropriate enum?

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
ziccardi: Lately, the idea is that the prefix should not be used, you are right. It used to be very common in the past, even in the Java World. 
However, it really depends on whether we have or not a standard naming convention in aerogear.
IMHO I think that the 'I' prefix for Interfaces is very useful.

Other opensource software like eclipse uses it extensively. An example: http://git.eclipse.org/c/actf/org.eclipse.actf.common.git/tree/plugins/org.eclipse.actf.core/src/org/eclipse/actf/core/adapt

@matzew WDYT?

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
ciaranRoche: I am working on that ticket, I pushed it to the PR on your fork @rachael-oregan , just testing it now.

TommyJ1994:
I think at the moment, there will still be role naming collisions with using a Map. For example, roles across different clients won't be unique, yet here all roles will be attached to a User Principal by classing them under a realm or clients? I think they might need to be classed under realm or client and clientID as a hasClientRole('web-role') looks like it would return true if I am just checking the roles for mobile client, since all client roles will be mapped to a user? 

TL;DR I think we will need to keep track of what clientID a client role is for, otherwise the access control might be easy to bypass.

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -19,7 +44,122 @@ public OIDCAuthCodeImpl(final AuthServiceConfig config) {
...
      */
     @Override
-    public Principal authenticate(final ICredential credential) {
-        throw new IllegalStateException("Not implemented");
+    public Principal authenticate(final ICredential credential) throws AuthenticationException {
+        if (credential instanceof OIDCCredentials == false) throw new IllegalArgumentException("Invalid Credentials");
+        UserPrincipalImpl user;
+        try {
+            userIdentity = getIdentityInformation(credential);
+           user = UserPrincipalImpl
```
aidenkeating: I think this will now need to be a `OIDCUserPrincipalImpl`

rachael-oregan:
@ziccardi @TommyJ1994 @ciaranRoche @aidenkeating mind having another look please?

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param role
+     * @param clientID
+     */
+    public RoleKey(final IRole role, final String clientID) {
+        roleName = role.getRoleName();
+        roleType = role.getRoleType();
+        this.clientID = clientID;
+    }
+
+    public RoleKey(final String roleName, final String clientId, final RoleType roleType){
```
ciaranRoche: Missing a comment from one of my changes

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,55 @@
...
+        if (((RoleKey) roleKey).roleType == RoleType.CLIENT) { //do a check on clientID
+            return ((RoleKey) roleKey).roleName.equals(roleName) && ((RoleKey) roleKey).roleType == roleType &&
+                ((RoleKey) roleKey).clientID == clientID;
+        } else {
+            return ((RoleKey) roleKey).roleName.equals(roleName) && ((RoleKey) roleKey).roleType == roleType;
+        }
+    }
+
+    @Override
+    public int hashCode(){
```
ciaranRoche: Just another comment missing from my changes.

Other then that LGTM

auth/src/main/java/org/aerogear/auth/RoleKey.java
rachael-oregan: Nice catch, thanks.

auth/src/main/java/org/aerogear/auth/RoleKey.java
rachael-oregan: I'll add a java doc here too, cheers

TommyJ1994:
Sorry @rachael-oregan, I'll try get eyes on this as soon as i can.

auth/src/main/java/org/aerogear/auth/IRole.java
```diff
@@ -2,4 +2,6 @@
 
 public interface IRole {
     String getRoleName();
+    RoleType getRoleType();
+    String getClientID();
```
ziccardi: This is just for ClientRole, so should not be in the general interface

auth/src/main/java/org/aerogear/auth/RealmRole.java
```diff
@@ -1,7 +1,31 @@
...
+        return RoleType.REALM;
+    }
+
+    /**
+     * Get's the client id for the realm role
+     *
+     * @return null (realms don't have client ids)
+     */
+    @Override
+    public String getClientID() {
```
ziccardi: If the RealmRole do not have a client ID than this method should not be here. Moreover, default implementations should be put in the AbstractRole class

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,66 @@
+package org.aerogear.auth;
+import java.util.Objects;
+
+/**
+ * Stores a user's keycloak roles information
+ */
+
+public class RoleKey {
+
+    private String roleName;
```
ziccardi: Make these final

auth/src/main/java/org/aerogear/auth/IRole.java
rachael-oregan: Can this function be kept here if your comment below is addressed?

auth/src/main/java/org/aerogear/auth/RoleKey.java
rachael-oregan: Nice catch thanks

auth/src/main/java/org/aerogear/auth/RealmRole.java
rachael-oregan: See comment above

auth/src/main/java/org/aerogear/auth/IRole.java
ziccardi: I think it would be better to remove it, since it is not a property of all the roles

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,66 @@
...
+     * Compares RoleKey objects for equality
+     *
+     * @param roleKey
+     * @return <code>true</code> or <code>false</code>
+     */
+    @Override
+    public boolean equals(final Object roleKey) {
+        if (this == roleKey) return true;
+        if (roleKey == null || roleKey.getClass() != getClass()) return false;
+        if (((RoleKey) roleKey).roleType == RoleType.CLIENT) { //do a check on clientID
```
ziccardi: Too many casts. I would suggest to change this to:
```java
public boolean equals(final Object o) {
    if (this == roleKey) return true; 
    if (roleKey == null || roleKey.getClass() != getClass()) return false;
    RoleKey roleKey = (RoleKey) o;
```

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,66 @@
...
+            return ((RoleKey) roleKey).roleName.equals(roleName) && ((RoleKey) roleKey).roleType == roleType &&
```
ziccardi: If passed it RoleKey has a null `roleName` a `NullPointerException` will be thrown here.

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Creates a new RoleKey object
+     *
+     * @param roleName
+     * @param roleType
+     * @param clientId
+     */
+    public RoleKey(final String roleName, final String clientId, final RoleType roleType){
+        this.roleName = roleName;
```
ziccardi: Add a check that roleName and roleType are not null, so that no other checks are needed later

auth/src/main/java/org/aerogear/auth/AbstractRole.java
```diff
@@ -22,4 +22,8 @@
     public String getRoleName() {
         return this.roleName;
     }
+
+    public String getClientID() {
```
wei-lee: not sure if we need this on the base abstract class. I thought this is a specific field for client roles?

auth/src/main/java/org/aerogear/auth/AuthServiceConfig.java
```diff
@@ -1,8 +1,15 @@
...
 
+import org.json.JSONObject;
+
 /**
  * Authentication service configuration
  */
 public class AuthServiceConfig {
     // TODO: to be implemented.
+
+    public JSONObject toJSON() {
```
wei-lee: what is this method for?

auth/src/main/java/org/aerogear/auth/IUserPrincipal.java
```diff
@@ -11,11 +11,19 @@
...
 
     /**
-     * Checks if the user has the specified role.
+     * Checks if the user has the specified Client role.
      * @param role role to be checked
+     * @param clientId clientID related to role
      * @return true or false
      */
-    boolean hasRole(IRole role);
+    boolean hasClientRole(String role, String clientId);
```
wei-lee: I prefer to put the `clientId` in front as it is the name space for the role.

wei-lee: actually, I am question if we need the `clientId` here at all as the value of `clientId` is already provided in the configuration file for the authentication service?

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -1,25 +1,217 @@
...
+
+    /**
+     * Parses the user's realm roles from the user identity {@link #userIdentity}
+     *
+     * @return user's realm roles
+     * @throws JSONException
+     * @see RoleKey
+     * @see RealmRole
+     */
+    private Map<RoleKey, IRole> parseRealmRoles() throws JSONException {
```
wei-lee: not sure why we return a Map here. Why not just return an array or list of RealmRole?

auth/src/main/java/org/aerogear/auth/impl/OIDCAuthCodeImpl.java
```diff
@@ -1,25 +1,217 @@
...
+                String clientRolesString = tokenClientRolesJSON.substring(1, tokenClientRolesJSON.length() - 1).replace("\"", "");
+                String roles[] = clientRolesString.split(COMMA);
+
+                for (String rolename : roles) {
+                    ClientRole clientRole = new ClientRole(rolename, initialClientID);
+                    clientRoles.put(new RoleKey(clientRole, initialClientID), clientRole);
+                }
+            }
+        }
+        return clientRoles;
```
wei-lee: similarly, not sure why we return a map here. We can just return an array or list of ClientRole?

auth/src/main/java/org/aerogear/auth/RoleKey.java
```diff
@@ -0,0 +1,66 @@
+package org.aerogear.auth;
+import java.util.Objects;
+
+/**
+ * Stores a user's keycloak roles information
+ */
+
+public class RoleKey {
```
wei-lee: not sure why we need this class?

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
```diff
@@ -29,7 +33,7 @@
     /**
      * Roles associated with this principal.
      */
-    private final Map<String, IRole> roles;
+    private final Map<RoleKey, IRole> roles;
```
wei-lee: I believe we can change this to an array or a list of IRole.

ciaranRoche: I think we were using an array list, but as all roles where mapped to User Principal, there could be a clash with the names of roles between realm and client. This in turn could make the access control easy to by pass.
So it was decided to create `RoleKey`,`RoleType` and track `ClientId` too and store them in a Map, so we could easily save and search by the above types and params.
Also just to clarify, this went through multi changes over the last two weeks, so it could be worth syncing up @rachael-oregan to ensure best path forward to close out this PR.

wei-lee: to avoid the clash, you can define the `equals` method of a role class to compare roleType, roleName and clientId. Two roles will only be the same if all three fields are the same. 

Once you have that, you can put them inside a list or array without worrying about clashing.

auth/src/main/java/org/aerogear/auth/ClientRole.java
```diff
@@ -1,7 +1,41 @@
 package org.aerogear.auth;
 
+/**
+ * Creates a new keycloak client role for a user
+ */
 public final class ClientRole extends AbstractRole {
```
matzew: this is framework code, right? be careful w/ making things final... If they not really really really need to be final

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
wei-lee: If we do as I described above, I also think we don't need this many interfaces & classes to represent the roles. We could just have 1 class - UserRole. It will have 3 fields:

* type
* name
* namespace (clientId)

A realm role is just a role with a null name space. 

I think this will allow us to achieve the same object, but with far less code.

auth/src/main/java/org/aerogear/auth/RoleType.java
```diff
@@ -0,0 +1,8 @@
+package org.aerogear.auth;
+
+/**
+ * Keycloak role types
+ */
+public enum RoleType {
+    REALM, CLIENT
```
matzew: perhaps do a ctor ? 

https://github.com/aerogear/aerogear-unifiedpush-server/blob/f6fe1622c595a40f54b6eb1c4f5a643e2e133f9a/service/src/main/java/org/jboss/aerogear/unifiedpush/service/impl/health/Status.java#L24-L27

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
ciaranRoche: Ah good call :+1:

auth/src/main/java/org/aerogear/auth/AbstractRole.java
ziccardi: +1

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
ziccardi: @wei-lee The main reason why a map has been used is to be able to get a role by name without the need to search through the whole list.
Since both realm and client roles can have the same name, the RoleKey class became useful to be used as key inside the map.

auth/src/main/java/org/aerogear/auth/ClientRole.java
ziccardi: It is like that by purpose. The idea is that a ClientRole is a ClientRole: we don't want it to be customized. If another role should be created, it should extend AbstractRole, so that the framework know that a ClientRole behave in a certain way and can't be changed.

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
wei-lee: @ziccardi once we have the name of the role, we can do this:

```
IRole roleToFind = new UserRole(rolename, clientId, RoleType.Client);
return roles.contains(roleToFind)
```

This way we don't need the extra classes & interfaces. I don't really see the value in them. Also I don't think it will be a problem performance wise as there won't be a lot of roles.

auth/src/main/java/org/aerogear/auth/ClientRole.java
matzew: than, let's state that, why it's locked down.

and why we don't think that a developer should not extend `ClientRole` for his `SuperDopeClientRole`. 

I've seen lot's of _traffic_ on mailing lists before that certains things, in OSS frameworks, were locked/final.  

Sure, delegation is possible - but stinks :)

auth/src/main/java/org/aerogear/auth/impl/UserPrincipalImpl.java
ziccardi: @wei-lee Yes, we can, but that will loop through the whole list (in the worst case). With this implementation, we have the overhead of looping through the whole list and comparing 2 strings for equality across all the compared objects.

wei-lee: Yes, I agree. But as I said, in reality, we are only looking at probably less than 10 items in the list. In this case, I don't see it become a problem. But it dramatically simplifies the implementation.

rachael-oregan:
There's been some back and forth on whether an arraylist or map should be used for storing a user's roles.  

Reason's for map:
- quickly get a user's roles - `roles.values()`
- quickly check does a user have a role by passing the role name and/or client id - `roles.containsKey()`
- a RoleKey is used in the map to avoid naming conflicts between realm and client role names
@ziccardi might have more reasons here

Reasons for arraylist
- simplifies the code - one class for roles that will have the attributes `type`, `name` `namespace` (clienId).  To avoid naming conflicts between client and realm roles we can override the `equals` method to compare the attributes
- there should be no performance issues as there shouldn't be a lot of roles
@wei-lee might have more reasons here.

I'd be in favour for using an arraylist as it's a lot more readable and maintainable to me.  And from the discussions above there shouldn't be any performance issues.  A lot of the code seems like overkill to me for something small like a user's roles.  I'm not a Java expert though so guys @ziccardi @wei-lee @TommyJ1994 @secondsun @aidenkeating @ciaranRoche @matzew  what do ye think?

pb82:
@rachael-oregan I don't want to further complicate things but instead of an ArrayList i would suggest a HashSet: it works like the ArrayList but guarantees that you can't have duplicates (you also need to  override `equals` and `hashCode`). Lookups (contains) are not iterative but use the hash value.

ciaranRoche:
+1 for the arraylist.

ziccardi:
We don't want duplicates, so if we want to remove the `Map`, I would go for a `HashSet` as suggested by @pb82.

auth/src/main/java/org/aerogear/auth/RoleType.java
rachael-oregan: I'm not really seeing the benefit of this?

rachael-oregan:
@wtrocki mind having a look please?

wtrocki:
@rachael-oregan I'm not sure if you want another person to take look on this, considering number of reviews that happened :)  I took a look on the code and I do not have anything against that. Personally not like that we use enum for types. You could avoid a lots of different if's etc. if we will have just two separate classes, but that's personal point of view.

I would say at this point it's best to get agreement with people who rejected PR and merge it. 
We could always improve it/refactor it when having bigger picture. From all the comments I think that @TommyJ1994 collisions is the most important one as it will be hard to track possible bugs once we merge that to master.

rachael-oregan:
@wtrocki thanks for having a look.  Just to give you some context, it was decided yesterday on a meeting that we use HashSet as the structure for a user's roles. This addresses both concerns of 1) its currently being over-engineered with extra interfaces/classes and 2) we want to ensure no duplicate roles.  

In relation to your comment on the enum types, do you think it would be better to use two different classes for the Role's? I'm afraid that might lead us down the over-engineered road again with extra classes/interfaces that will probably need to be introduced.  It is something to consider though like you said when we have the bigger picture.

The issue with collisions has been addressed by using HashSet's for a user roles.  If you think its best to get approval from other members then that's OK I'll wait until they are free.  Thanks :)

PR #16, opened 2018-01-25
Service registry

secondsun:
## Motivation

This implements the ServiceRegistry as described in the aerogear-android-sdk proposal.

## Description

This implements the service registry system.  The Unit tests remonstrate how to use it, and additional sample code can be found here : https://github.com/secondsun/WipDemo

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -18,7 +18,7 @@ import java.util.Map;
 /**
  * MobileCore is the entry point into AeroGear mobile services that are managed by the mobile-core
  * ¿feature( TODO: Get correct noun )? in OpenShift.
- * <p>
+ *
  * Usage.java
  * ```
  * MobileCore core = new MobileCore.Builder(context, R.raw.mobile_core).build();
@@ -31,7 +31,6 @@ public final class MobileCore implements ServiceModule {
 
     private final Context context;
     private final String mobileServiceFileName;
-
     private Map<String, ServiceConfiguration> configurationMap;
 
     private MobileCore(@NonNull Context context, String mobileServiceFileName) {
@@ -83,7 +82,6 @@ public final class MobileCore implements ServiceModule {
             this.context = context;
         }
 
-
         /**
          * The filename of the mobile service configuration file in the assets directory.
          *
@@ -125,6 +123,7 @@ public final class MobileCore implements ServiceModule {
             if (!built) {
                 built = true;
                 MobileCore core = new MobileCore(context, mobileServiceFileName);
+
                 core.bootstrap();
                 return core;
             } else {
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -2,7 +2,6 @@ package org.aerogear.mobile.core;
 
 import android.content.Context;
 import android.support.annotation.NonNull;
-import android.util.JsonReader;
 import android.util.Log;
 
 import org.aerogear.mobile.core.configuration.MobileCoreJsonParser;
@@ -12,12 +11,18 @@ import org.json.JSONException;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
+import java.util.AbstractSequentialList;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 
 /**
  * MobileCore is the entry point into AeroGear mobile services that are managed by the mobile-core
- * ¿feature( TODO: Get correct noun )? in OpenShift.
+ * feature( TODO: Get correct noun )? in OpenShift.
  *
  * Usage.java
  * ```
@@ -26,16 +31,20 @@ import java.util.Map;
  * core.feature(MyRestService.class).upload(myData);// Would upload myData and be secured by KeyCloak
  * ```
  */
-public final class MobileCore implements ServiceModule {
+public final class MobileCore {
 
 
     private final Context context;
     private final String mobileServiceFileName;
-    private Map<String, ServiceConfiguration> configurationMap;
+    private final ServiceModuleRegistry serviceRegistry;
 
-    private MobileCore(@NonNull Context context, String mobileServiceFileName) {
+    private Map<String, ServiceConfiguration> configurationMap;
+    private Map<String, ServiceModule> services = new HashMap<>();
+
+    private MobileCore(@NonNull Context context, @NonNull String mobileServiceFileName, @NonNull ServiceModuleRegistry registryService) {
         this.context = context.getApplicationContext();
         this.mobileServiceFileName = mobileServiceFileName;
+        this.serviceRegistry = registryService;
     }
 
     public Logger defaultLog() {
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -45,23 +53,87 @@ public Logger defaultLog() {
...
+                                                        //throw an exception.
+            while (!servicesPending.isEmpty()) {
+                unresolvableServiceDetected = true;
+                String serviceUnderInstanciation = "";
+                for (String serviceName : servicesPending) {
+                    List<String> dependencies = serviceRegistry.getDependenciesFor(serviceName);
+                    serviceUnderInstanciation = serviceName;
+                    if (dependencies.isEmpty() || servicesInitted.containsAll(dependencies)) {
+                        unresolvableServiceDetected = false;
+                        Class<? extends ServiceModule> serviceClass = serviceRegistry.getServiceClass(serviceName);
```
josemigallas: We could extract a method here called `initService` or `startService` for the sake of clarity. This piece is a bit convoluted.
```java
private void startService(String serviceName) {
    Class<? extends ServiceModule> serviceClass = serviceRegistry.getServiceClass(serviceName);
    ServiceModule serviceInstance = serviceClass.newInstance();
    serviceInstance.bootstrap(this, getConfig(serviceName));
    this.services.put(serviceName, serviceInstance);
    servicesInitted.add(serviceName);
}
```
I also would move sentence in 92 to right after service is started and before it breaks from the loop. The final condition would be a bit clearer:
```java
if (unresolvableServiceDetected) {
    throw new ...
}
```

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -45,23 +53,87 @@ public Logger defaultLog() {
...
+                } else {
+                    throw new BootstrapException(String.format("Unresolvable service detected %s", serviceUnderInstanciation));
+                }
+            }
+
         } catch (JSONException | IOException e) {
             defaultLog().error(e.getMessage(), e);
             throw new BootstrapException(String.format("%s could not be loaded", mobileServiceFileName), e);
+        } catch (IllegalAccessException | InstantiationException e) {
+            e.printStackTrace();
```
josemigallas: Question: if the Exception is thrown upwards as a BootstrapException, it will probably be printed eventually, either in other catch block or when the app breaks. We would have in that case the same stacktrace printed twice, wouldn't we?

core/src/main/java/org/aerogear/mobile/core/configuration/ServiceConfiguration.java
```diff
@@ -1,5 +1,7 @@
 package org.aerogear.mobile.core.configuration;
 
+import org.json.JSONObject;
```
josemigallas: Unused dependency?

core/src/test/java/org/aerogear/mobile/core/ServiceRegistryTest.java
```diff
@@ -0,0 +1,112 @@
...
+    private Application application;
+
+    @Before
+    public void parseMobileCore() {
+        this.application = RuntimeEnvironment.application;
+
+    }
+
+    @Test()
+    public void testSimpleServiceInit() {
```
josemigallas: It doesn't make it less clear but IMHO test cases' names would be cleaner w/o the "test" prefix.

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
```diff
@@ -17,34 +19,37 @@
...
+    private MobileCore core;
 
     public KeyCloakService(@NonNull Context appContext) {
         this.appContext = appContext.getApplicationContext();
     }
 
-
     /**
-     * Loads keycloak.json from R.raw.keycloak
+     * Exchanges the google id token and configures the KeyCloakService to serve requests
```
josemigallas: Google? Or Keycloak? This method is empty, without a TODO, should we include it in the commit anyway?

secondsun: This is a test class that is going away in favor of IDM soon.  We shouldn't waste time on it.

core/src/test/java/org/aerogear/mobile/core/ServiceRegistryTest.java
secondsun: No they aren't.

"Test" is a verb that makes the method name read like a sentence.

core/src/main/java/org/aerogear/mobile/core/ServiceModuleRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+    public ServiceModuleRegistry() {}
+
+    /**
+     * Register the looking for a service module to a class.
+     *
+     * @param type the type of module to use
+     * @param moduleClass the class that implements a service module type
+     * @param dependsOn the other serviceTypes this Service Depends on.
+     */
+    public void registerServiceModule(String type, Class<? extends ServiceModule> moduleClass, String... dependsOn) {
```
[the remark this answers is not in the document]
secondsun: it out load

"register service module keycloak module class KeyCloakServiceModule dependsOn http"

vs

"register service module keycloak module class KeyCloakServiceModule dependencies http"

which sounds better?

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -12,31 +11,40 @@
...
 
     private final Context context;
     private final String mobileServiceFileName;
+    private final ServiceModuleRegistry serviceRegistry;
 
     private Map<String, ServiceConfiguration> configurationMap;
+    private Map<String, ServiceModule> services = new HashMap<>();
 
-    private MobileCore(@NonNull Context context, String mobileServiceFileName) {
+    private MobileCore(@NonNull Context context, @NonNull String mobileServiceFileName, @NonNull ServiceModuleRegistry registryService) {
```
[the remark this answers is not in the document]
secondsun: Good catch

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -45,23 +53,87 @@ public Logger defaultLog() {
         };
     }
 
-    @Override
     public void bootstrap(Object... args) {
 
-
         try (InputStream configStream = context.getAssets().open(this.mobileServiceFileName);) {
+
+        Set<String> servicesInitted = new HashSet<>();
```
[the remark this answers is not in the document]
secondsun: since the method that is called is bootstrap I will rename it to servicesBootstrapped.  It is more directly tied to the code (otherwise people may look for spurious init or start methods.)(I'm avoiding start in Android because onStart has a lot of meanings in the platform)

core/src/main/java/org/aerogear/mobile/core/ServiceModuleRegistry.java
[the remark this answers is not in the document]
secondsun: hrm  actually on second thought dependencies is better because I get to say dependencies everywhere.

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
secondsun: @josemigallas this method is removed in PR 17 https://github.com/secondsun/aerogear-android-sdk/blob/2c526fda60670ea761f27201b42058b2802a1730/keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
secondsun: +1

secondsun:
@josemigallas  Updated

core/src/test/java/org/aerogear/mobile/core/ServiceRegistryTest.java
josemigallas: You're absolutely right

PR #19, opened 2018-01-26
Agdroid 711 config params

secondsun:
## Motivation

JIRA: https://issues.jboss.org/browse/AGDROID-771

## Description

See the description for the JIRA.

## Progress

- [x] Adding instances to the service registry
- [x] Updating mobile-core to start services because they are defined in the mobile-services.json instead of because they are defined in the service registry
- [x] Minor cleanups
- [x] Test updates

## Additional Notes

@aidenkeating Does this help?

To test this you can run the unit tests `gradlew testDebug` and create a sample application that connects to mobile services.  You may use https://github.com/secondsun/WipDemo as a reference

### DIFF
--- a/core/src/debug/assets/undefined-mobile-services.json
+++ b/core/src/debug/assets/undefined-mobile-services.json
@@ -6,40 +6,6 @@
         "uri": "https://fh-sync-server-myproject.192.168.37.1.nip.io"
       },
       "name": "fh-sync-server"
-    },
-    {
-      "config": {
-        "auth-server-url": "http://keycloak-myproject.192.168.37.1.nip.io/auth",
-        "clientId": "OeFYkwEuUKXvDsSudfXc",
-        "headers": {},
-        "realm": "myproject",
-        "resource": "OeFYkwEuUKXvDsSudfXc",
-        "ssl-required": "external",
-        "url": "http://keycloak-myproject.192.168.37.1.nip.io/auth"
-      },
-      "name": "keycloak"
-    },
-    {
-      "config": {
-        "headers": {},
-        "name": "prometheus",
-        "type": "prometheus",
-        "uri": "https://prometheus-myproject.192.168.37.1.nip.io"
-      },
-      "name": "prometheus"
-    },
-    {
-      "config": {
-        "headers": {},
-        "mysql_db": "unifiedpush",
-        "mysql_password": "unifiedpush",
-        "mysql_username": "unifiedpush",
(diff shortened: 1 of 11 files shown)


core/src/debug/assets/undefined-mobile-services.json
```diff
@@ -0,0 +1,12 @@
+{
```
wtrocki: Probably pushed by mistake. File name is wrong etc.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -88,15 +96,23 @@ public void bootstrap() {
     }
 
     /**
-     *
+     * There are some services that are "core" and usually won't appear in mobile-services.json
+     */
+    private void addCoreServices() {
```
wtrocki: I think that core services may be well known at compilation time and will not change. 
Having runtime, string based helpers (I do not want to call them services to not make confusion with top level SDK) seems like we trying to invent something that is already supported by Java language. I might be missing actual reason, but we could just make core helpers available as part of the API. WDYT?  Service configuration is not needed here and it may be misleading. 

I do not have strong opinion about it, but this code definitely sticks out of the rest of the changes and just wanted to point that out.

wtrocki:
@secondsun It will be cool to get documentation explaining usage of core and end user application usage. 
Some top level document that will help people to get understanding how this may work.

core/src/debug/assets/undefined-mobile-services.json
secondsun: Oh no this is right.  It is for some of the new tests

secondsun:
@wtrocki Yeah more docs are on my short term todo list.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
secondsun: @wtrocki You are probably right.  So what I was trying to do was 1) test the servicemodule concept with as much usage as I could and 2) provide a mechanism for hiding concrete classes.  The SMR just really isn't the right way is the feeling I am getting.

I'll take it to the mailing soon, now that we've gotten more ideas about other parts of the system it is a perfect time to revisit.

secondsun:
@wtrocki or @danielpassos can I get a merge merge?

wtrocki:
Leaving that for @danielpassos

PR #21, opened 2018-01-31
AGDROID-686 Token integrity check

aidenkeating:
## Motivation

JIRA: https://issues.jboss.org/browse/AGDROID-686

## Description

Function to Verify JWT Tokens (Access, Identity, Refresh) using a Keycloak Realm Public Key before they are used as part of client side access control decisions.

Provide the configuration from the JSON mobile configuration to the OIDCCredentials.

Tested successfully with both valid & tampered Access/Identity tokens. Identity Brokering originated tokens can also be validated.

## Progress

- [x] Get the Public Key from the Mobile Core Config (when available)
- [x] Get the Audience from the Mobile Core Config (when available)
- [x] Construct the integrity parameters (build issuer etc.) from the core config.
- [x] Add Unit tests - Test Wrong issues, audiences, tampered signature, payload, algorithm info sections etc


aidenkeating:
@ziccardi @secondsun Would you mind taking a look?

This adds `IntegrityCheckParameters` that are part of the `OIDCCredentials` containing the `issuer`, `audience` and `publicKey` from the mobile JSON config.

At the moment it looks like the SDK user would provide the credential at login. In this case I'm kind of confused as at the moment Credential seems to be used as a container for the tokens when a user has already authenticated. I don't understand why/how a user would provide that information at login.

Something tells me `OIDCCredential` has become more like AuthState and `OIDCCredentials` should be refactored to `OIDCAuthState`. Something that is created once a user is logged in instead of something the SDK user provides.

A call may be better to explain this.

ziccardi:
@aidenkeating The authentication is performed through the usage of an authentication chain that can be configured at runtime.
Some of the rings of the authentication chain require a credential to be supplied (for example, to check that the token is still valid), some of them do not require the credentials (for example the browser authentication).
In both cases, after the user has been authenticated, an UserPrincipal object is returned containing the UserCredentials (in case of OIDC, it will contain the auth tokens).

auth/src/main/java/org/aerogear/auth/AbstractAuthenticator.java
```diff
@@ -12,10 +13,11 @@
     /**
      * Authentication service configuration.
      */
-    private final AuthServiceConfig config;
+    private final ServiceConfiguration serviceConfig;
 
-    public AbstractAuthenticator(final AuthServiceConfig config) {
-        this.config = config;
+
+    public AbstractAuthenticator(ServiceConfiguration serviceConfig) {
```
ziccardi: Make the parameter final

auth/src/main/java/org/aerogear/auth/AuthService.java
```diff
@@ -21,14 +24,15 @@
 
     /**
      * Instantiates a new AuthService object
-     * @param config Authentication Service configuration
      */
-    private AuthService(final AuthServiceConfig config) {
+    public AuthService() {}
+
+    public void bootstrap(MobileCore core, ServiceConfiguration serviceConfig) {
```
ziccardi: Make the params final

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public String serialize() throws JSONException {
+        return new JSONObject()
+            .put("audience", this.audience)
+            .put("issuer", this.issuer)
+            .put("publicKey", this.publicKey)
+            .toString();
+    }
+
+    public static IntegrityCheckParameters deserialize(String serializedParams) throws JSONException {
```
ziccardi: Make the param final

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class IntegrityCheckParameters implements IIntegrityCheckParameters {
+
+    private String issuer;
+    private String audience;
+    private String publicKey;
+
+    public IntegrityCheckParameters(String audience, String issuer, String publicKey) {
```
ziccardi: Make the params final

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,77 @@ public String getRefreshToken() {
...
     }
 
+    public IIntegrityCheckParameters getIntegrityCheckParameters() { return this.integrityCheckParameters; }
+
+    /**
+     * Verify the authenticity of a JWT token against integrity parameters (provided config).
+     * @param jwtToken The JWT token to verify.
+     * @return <code>true</code> if the token integrity is good.
+     */
+    public boolean verifyToken(String jwtToken) {
```
ziccardi: Make the param final

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,77 @@ public String getRefreshToken() {
...
+        String issuer = integrityCheckParameters.getIssuer();
```
ziccardi: Always instantiate final variables if they are not meant to be changed

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,77 @@ public String getRefreshToken() {
...
+
+    /**
+     * A function to verify the authenticity of a JWT token against a public key, expected issuer and audience.
+     * @param jwtToken - The JWT Token to Verify.
+     * @param publicKey - The Public Key from Keycloak, without the Begin/End tags.
+     * @param issuer - The expected Issuer of the JWT
+     * @param audience - The expected Audience of the JWT
+     * @return <code>true</code> if the token integrity is good.
+     */
+    public boolean verifyToken(String jwtToken, String publicKey, String issuer, String audience) {
```
ziccardi: See previous comments

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,77 @@ public String getRefreshToken() {
...
+                        new AlgorithmConstraints(AlgorithmConstraints.ConstraintType.WHITELIST, // which is only RS256 here
+                                AlgorithmIdentifiers.RSA_USING_SHA256))
+                .build(); // create the JwtConsumer instance
+
+        try
+        {
+            //  Validate the JWT and process it to the Claims
+            JwtClaims jwtClaims = jwtConsumer.processToClaims(jwtToken);
+            verified = true;
+            System.out.println("JWT Verified Successfully.");
```
ziccardi: Don't use System.out

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,77 @@ public String getRefreshToken() {
...
+        }
+        catch (InvalidJwtException e)
+        {
+            // InvalidJwtException will be thrown, if the JWT failed processing or validation in anyway.
+            System.out.println("JWT Validation Failed. " + e.getLocalizedMessage());
```
ziccardi: See previous comment

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -67,8 +153,19 @@ public boolean isAuthorized() {
...
-    public String serialise() {
-        return this.authState.jsonSerializeString();
+    public String serialize() throws JSONException {
+        return new JSONObject()
+            .put("authState", this.authState.jsonSerializeString())
+            .put("integrityCheck", this.integrityCheckParameters.serialize())
+            .toString();
+    }
+
+    public static OIDCCredentials deserialize(String serializedCredential) throws JSONException {
```
ziccardi: Make parameter final

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -67,8 +153,19 @@ public boolean isAuthorized() {
      * Returns stringified JSON for the OIDCCredential.
      * @return Stringified JSON OIDCCredential
      */
-    public String serialise() {
-        return this.authState.jsonSerializeString();
+    public String serialize() throws JSONException {
+        return new JSONObject()
+            .put("authState", this.authState.jsonSerializeString())
+            .put("integrityCheck", this.integrityCheckParameters.serialize())
```
ziccardi: integrityCheckParameters could be null

secondsun:
@aidenkeating Do you have a demo project that shows the code being used somewhere?

auth/build.gradle
```diff
@@ -26,14 +26,23 @@ android {
...
     testImplementation 'org.mockito:mockito-core:2.10.0'
     testImplementation 'junit:junit:4.12'
+    testImplementation "org.robolectric:robolectric:3.6.1"
+    compile 'org.bitbucket.b_c:jose4j:0.6.3'
 }
 
 sourceCompatibility = "1.7"
-targetCompatibility = "1.7"
\ No newline at end of file
+targetCompatibility = "1.7"
```
secondsun: Remove these two lines, they are already set above to 1.8

aidenkeating:
@secondsun Not yet. I can implement it in https://github.com/feedhenry/mobile-security-android-template

auth/src/main/java/org/aerogear/auth/AbstractAuthenticator.java
```diff
@@ -1,6 +1,7 @@
 package org.aerogear.auth;
```
secondsun: packages need to be renamed to conform to https://github.com/aerogear/proposals/blob/master/sdks/SDK-naming.md

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,69 @@ public String getRefreshToken() {
...
+     * @param jwtToken - The JWT Token to Verify.
+     * @param publicKey - The Public Key from Keycloak, without the Begin/End tags.
+     * @param issuer - The expected Issuer of the JWT
+     * @param audience - The expected Audience of the JWT
+     * @return <code>true</code> if the token integrity is good.
+     */
+    public boolean verifyToken(final String jwtToken, final String publicKey, final String issuer, final String audience) {
+        // add the Begin/End tags to the public key generated from Keycloak
+        String beginPublicKey = "-----BEGIN PUBLIC KEY-----";
+        String endPublicKey = "-----END PUBLIC KEY-----";
```
secondsun: these should be private static final constant fields instead of local variables.

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,69 @@ public String getRefreshToken() {
...
+                        new AlgorithmConstraints(AlgorithmConstraints.ConstraintType.WHITELIST, // which is only RS256 here
+                                AlgorithmIdentifiers.RSA_USING_SHA256))
+                .build(); // create the JwtConsumer instance
+
+        try {
+            //  Validate the JWT and process it to the Claims
+            JwtClaims jwtClaims = jwtConsumer.processToClaims(jwtToken);
+            return true;
+        } catch (InvalidJwtException e) {
+            e.printStackTrace();
```
secondsun: For now use Android's Log object instead of printing the stack trace.

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -67,8 +145,21 @@ public boolean isAuthorized() {
      * Returns stringified JSON for the OIDCCredential.
      * @return Stringified JSON OIDCCredential
      */
-    public String serialise() {
-        return this.authState.jsonSerializeString();
+    public String serialize() throws JSONException {
```
secondsun: How is serialize used?

secondsun: Personally I would remove the JSONException from the method declaration and rethrow it as a runtime exception and add that clue to the documentation.

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -67,8 +145,21 @@ public boolean isAuthorized() {
...
+        JSONObject jsonCredential = new JSONObject()
+            .put("authState", this.authState.jsonSerializeString());
+        if (this.integrityCheckParameters != null) {
+            jsonCredential.put("integrityCheck", this.integrityCheckParameters.serialize());
+        }
+        return jsonCredential.toString();
+    }
+
+    public static OIDCCredentials deserialize(final String serializedCredential) throws JSONException {
```
secondsun: See above comments about JSONException.

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
```diff
@@ -39,11 +39,11 @@ public OIDCCredentials read() {
      * Saves a token
      * @param authState token to be saved
      */
-    public synchronized void write(final OIDCCredentials authState) {
+    public synchronized void write(final OIDCCredentials authState) throws JSONException {
```
secondsun: I don't like the name "write".  I think save is more descriptive.  Also if we wrap the JSONExceptions above we aren't "leaking" it here.

auth/src/test/java/org/aerogear/auth/credentials/OIDCCredentialsTest.java
```diff
@@ -0,0 +1,125 @@
...
+        OIDCCredentials oidcCredentials = new OIDCCredentials();
+        assertTrue("Expect Token Validation with Valid Refresh Token to Succeed", oidcCredentials.verifyToken(VALID_REFRESH_TOKEN, VALID_PUBLIC_KEY, VALID_ISSUER, VALID_AUDIENCE));
+    }
+
+    @Test
+    public void testTamperedRefreshToken() {
+        OIDCCredentials oidcCredentials = new OIDCCredentials();
+        assertFalse("Expect Token Validation with Tampered Refresh Token to Fail", oidcCredentials.verifyToken(TAMPERED_REFRESH_TOKEN, VALID_PUBLIC_KEY, VALID_ISSUER, VALID_AUDIENCE));
+    }
+
```
secondsun: Could the repeated "OIDCCredentials oidcCredentials = new OIDCCredentials();" be in the setup block?

auth/src/main/java/org/aerogear/auth/AuthService.java
```diff
@@ -21,14 +24,15 @@
 
     /**
      * Instantiates a new AuthService object
-     * @param config Authentication Service configuration
      */
-    private AuthService(final AuthServiceConfig config) {
+    public AuthService() {}
+
+    public void bootstrap(final MobileCore core, final ServiceConfiguration serviceConfig) {
```
ziccardi: When is the bootstrap called?
If we don't initialise the object inside the constructor, we should than throw an IllegalStateException if other methods are called before the bootstrap one.

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public IntegrityCheckParameters(final String audience, final String issuer, final String publicKey) {
+        this.issuer = issuer;
+        this.audience = audience;
+        this.publicKey = publicKey;
+    }
+
+    public IntegrityCheckParameters() {}
+
+    public String getIssuer() {
```
ziccardi: Add some documentation here. What is the issuer?

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+        return this.issuer;
+    }
+
+    public String getAudience() {
+        return this.audience;
+    }
+
+    public String getPublicKey() { return this.publicKey; }
```
ziccardi: What is this public key? What is the format it is returned? It is a DSA? RSA? or what?

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public String serialize() throws JSONException {
+        return new JSONObject()
+            .put("audience", this.audience)
+            .put("issuer", this.issuer)
+            .put("publicKey", this.publicKey)
+            .toString();
+    }
+
+    public static IntegrityCheckParameters deserialize(final String serializedParams) throws JSONException {
```
ziccardi: Is this object supposed to be constructed only through this method? If yes, the constructor should be private

ziccardi: A null check should be added here

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,69 @@ public String getRefreshToken() {
...
+        String endPublicKey = "-----END PUBLIC KEY-----";
+        String constructedPublicKey = beginPublicKey + publicKey + endPublicKey;
+
+        // Convert the public key from a string to a Java security key
+        RsaKeyUtil utils = new RsaKeyUtil();
+        PublicKey jwtPublicKey = null;
+        try {
+            jwtPublicKey = utils.fromPemEncoded(constructedPublicKey);
+        } catch (JoseException e) {
+            e.printStackTrace();
```
ziccardi: ignoring exceptions is never a good idea.

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
aidenkeating: @secondsun serialize is used when persisting the auth state in the AuthStateManager.

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -32,6 +47,69 @@ public String getRefreshToken() {
...
+
+
+        // Validate and process the JWT.
+        JwtConsumer jwtConsumer = new JwtConsumerBuilder()
+                .setRequireExpirationTime() // require the JWT to have an expiration time
+                .setAllowedClockSkewInSeconds(30) // allow some leeway in validating time based claims to account for clock skew
+                .setRequireSubject() // require the subject claim
+                .setExpectedIssuer(issuer) // whom the JWT needs to have been issued by
+                .setExpectedAudience(audience) // to whom the JWT is intended for
+                .setVerificationKey(jwtPublicKey) // verify the signature with the public key
```
ziccardi: is it ok to have a null `jwtPublicKey` here? If it is not, make it `final` and do not assign `null` to it. 
```java
final PublicKey jwtPublicKey;
try {
    jwtPublicKey = utils.fromPemEncoded(constructedPublicKey);
} catch (JoseException e) {
```

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
ziccardi: If we can safely return `false` here, it would be better to use logger to log exceptions.

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -67,8 +145,21 @@ public boolean isAuthorized() {
      * Returns stringified JSON for the OIDCCredential.
      * @return Stringified JSON OIDCCredential
      */
-    public String serialise() {
-        return this.authState.jsonSerializeString();
+    public String serialize() throws JSONException {
+        JSONObject jsonCredential = new JSONObject()
```
ziccardi: Every time a variable is not supposed to be assigned again, it should be declared as final.

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
ziccardi: If this object is supposed to be created only through this method, the constructor should be private.

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
```diff
@@ -39,11 +39,11 @@ public OIDCCredentials read() {
...
         if (authState == null) {
             clear();
         } else {
-            if(!prefs.edit().putString(KEY_STATE, authState.serialise()).commit()) {
+            if(!prefs.edit().putString(KEY_STATE, authState.serialize()).commit()) {
                 throw new IllegalStateException("Failed to update state from shared preferences");
```
ziccardi: Is it correct to throw an unchecked exception here?

auth/src/main/java/org/aerogear/auth/AbstractAuthenticator.java
ziccardi: @secondsun : just to be sure, should we rename it to `org.aerogear.android.ags.auth`?

auth/src/main/java/org/aerogear/auth/AuthService.java
aidenkeating: @secondsun Am I correct in saying the core will initialize and then call bootstrap on a ServiceModule?

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
aidenkeating: No this will need to be constructed through the constructor when initially created at authentication time.

auth/src/main/java/org/aerogear/auth/AuthService.java
secondsun: @aidenkeating Yes.  https://github.com/aerogear/proposals/blob/master/android-sdk/README.md#starting-up-mobile-core

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
secondsun: @ziccardi Yes.

auth/build.gradle
```diff
@@ -26,14 +26,20 @@ android {
...
 }
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'net.openid:appauth:0.7.0'
+    implementation project(path: ':core')
     testImplementation 'org.mockito:mockito-core:2.10.0'
     testImplementation 'junit:junit:4.12'
+    testImplementation "org.robolectric:robolectric:3.6.1"
+    compile 'org.bitbucket.b_c:jose4j:0.6.3'
```
secondsun: compile is deprecated.  replace with implementation and move above the test dependencies.  Also can you include the dependency in the android bom https://github.com/aerogear/aerogear-parent/blob/master/aerogear-android-sdk-bom/pom.xml, passos is working on moving this to use a more managed system.

auth/src/main/java/org/aerogear/auth/AbstractAuthenticator.java
secondsun: @ziccardi Yes

auth/src/main/java/org/aerogear/auth/credentials/IIntegrityCheckParameters.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.json.JSONException;
+
+public interface IIntegrityCheckParameters {
+
+    String getAudience();
+    String getIssuer();
+    String getPublicKey();
+    boolean isValid();
+    String serialize() throws JSONException;
```
secondsun: nix the exception

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,69 @@
...
+        return this.audience;
+    }
+
+    /**
+     * Get the PEM encoded public key (RSA).
+     * @return public key parameter.
+     */
+    public String getPublicKey() { return this.publicKey; }
+
+    public String serialize() throws JSONException {
```
secondsun: This code will never throw a JSON exception, best not to declare it.

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,69 @@
...
+            .put("issuer", this.issuer)
+            .put("publicKey", this.publicKey)
+            .toString();
+    }
+
+    /**
+     * Return json representation of the parameters
+     * @return json string representation of parameters
+     */
+    public static IntegrityCheckParameters deserialize(final String serializedParams) throws JSONException {
```
secondsun: Kill the exception.

If you have to throw a checked exception (pro tip, you don't), make it more specific to the task at hand.  For instance if there is no "issuer" property on the object then throwing as JSON exception is less useful than throwing a new IllegalArgumentException with a useful message.

Also, add that the object needs to be a json object with those parameters in the header.

Also also, if all we are doing is storing this in shared preferences can we make it a parcelable instead?

auth/src/main/java/org/aerogear/auth/credentials/OIDCCredentials.java
```diff
@@ -1,23 +1,42 @@
...
 public class OIDCCredentials implements ICredential {
 
+    private static final String TAG = "OIDCCredentials";
+
     private final AuthState authState;
+    private final IIntegrityCheckParameters integrityCheckParameters;
 
-    public OIDCCredentials() {
-        this.authState = new AuthState();
+    public OIDCCredentials(final String serialisedCredential, final IIntegrityCheckParameters integrityCheckParameters) throws JSONException {
```
secondsun: Kill the JSONException.  Add javadocs.

auth/src/main/java/org/aerogear/auth/utils/AuthStateManager.java
secondsun: However, we should add `@throws`  to the javadocs

auth/src/main/java/org/aerogear/auth/credentials/IntegrityCheckParameters.java
aidenkeating: @secondsun I'm not sure I understand how parcelable would help here?

aidenkeating:
@ziccardi Mind taking another look?

auth/src/main/java/org/aerogear/android/ags/auth/AbstractAuthenticator.java
```diff
@@ -40,7 +42,5 @@ public void logout(final Principal principal) {
      * Returns the authentication service configuration
      * @return the authentication service configuration
      */
-    protected AuthServiceConfig getConfig() {
-        return config;
-    }
+    public ServiceConfiguration getServiceConfig() { return this.serviceConfig; }
```
ziccardi: Is ServiceConfiguration an immutable object?
If it is not, different threads could change the content of the `ServiceConfiguration` object, thus changing the internal state of the Authenticator without any control by the latter, making it crash prone.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,77 @@
+package org.aerogear.android.ags.auth.credentials;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class IntegrityCheckParameters implements IIntegrityCheckParameters {
+
+    private String issuer;
```
ziccardi: Make these final

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,206 @@
...
+
+    public IIntegrityCheckParameters getIntegrityCheckParameters() { return this.integrityCheckParameters; }
+
+    /**
+     * Verify the authenticity of a JWT token against integrity parameters (provided config).
+     * @param jwtToken The JWT token to verify.
+     * @return <code>true</code> if the token integrity is good.
+     */
+    public boolean verifyToken(final String jwtToken) {
+        final String issuer = integrityCheckParameters.getIssuer();
```
ziccardi: There is no added value in using variables to store the values here.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,206 @@
...
+        final String constructedPublicKey = beginPublicKey + publicKey + endPublicKey;
+
+        // Convert the public key from a string to a Java security key
+        final RsaKeyUtil utils = new RsaKeyUtil();
+        final PublicKey jwtPublicKey;
+        try {
+            jwtPublicKey = utils.fromPemEncoded(constructedPublicKey);
+        } catch (JoseException e) {
+            Log.e(TAG, e.getMessage(), e);
+            throw new RuntimeException(e);
```
ziccardi: Please use a more specific exception. In this case, if the parameters are not correct, an IllegalArgumentException could be the right one.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,206 @@
...
+    public String serialize() {
+        try {
+            final JSONObject jsonCredential = new JSONObject()
+                .put("authState", this.authState.jsonSerializeString());
+            if (this.integrityCheckParameters != null) {
+                jsonCredential.put("integrityCheck", this.integrityCheckParameters.serialize());
+            }
+            return jsonCredential.toString();
+        } catch(JSONException e) {
+            throw new RuntimeException(e);
```
ziccardi: See previous comment

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,206 @@
...
+
+    public static OIDCCredentials deserialize(final String serializedCredential) {
+        try {
+            final JSONObject jsonCredential = new JSONObject(serializedCredential);
+            final String serializedAuthState = jsonCredential.getString("authState");
+            final String serializedIntegrityChecks = jsonCredential.getString("integrityCheck");
+            final IntegrityCheckParameters icParams = IntegrityCheckParameters.deserialize(serializedIntegrityChecks);
+            return new OIDCCredentials(serializedAuthState, icParams);
+        } catch(JSONException e) {
+            throw new RuntimeException(e);
```
ziccardi: See previous comment

auth/src/main/java/org/aerogear/android/ags/auth/AbstractAuthenticator.java
aidenkeating: Not too familiar with any core stuff. @secondsun would you have an answer to this?

auth/build.gradle
```diff
@@ -26,14 +26,20 @@ android {
...
 }
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'net.openid:appauth:0.7.0'
+    implementation 'org.bitbucket.b_c:jose4j:0.6.3'
+    implementation project(path: ':core')
     testImplementation 'org.mockito:mockito-core:2.10.0'
     testImplementation 'junit:junit:4.12'
+    testImplementation "org.robolectric:robolectric:3.6.1"
```
secondsun: You'll need to rebase on master and add these dependencies to the aerogear-android-sdk-bom.  @danielpassos Can show you how.

auth/src/main/java/org/aerogear/android/ags/auth/AbstractAuthenticator.java
secondsun: ServiceConfiguration is meant to be immutable.  If it isn't then feel free to make it so.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/IIntegrityCheckParameters.java
```diff
@@ -0,0 +1,10 @@
+package org.aerogear.android.ags.auth.credentials;
+
+public interface IIntegrityCheckParameters {
+
+    String getAudience();
+    String getIssuer();
+    String getPublicKey();
+    boolean isValid();
+    String serialize();
```
secondsun: Do you think JavaDocs are needed?  The properties and things look self explanatory to me, but if there are any weirdnesses or edge cases to be aware of (like what does isValid check to validate) then I would want them added.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,81 @@
+package org.aerogear.android.ags.auth.credentials;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class IntegrityCheckParameters implements IIntegrityCheckParameters {
```
secondsun: For the methods you implement you need to add the `@Override` annotation.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/IntegrityCheckParameters.java
```diff
@@ -0,0 +1,81 @@
...
+                .put("publicKey", this.publicKey)
+                .toString();
+        } catch(JSONException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Return json representation of the parameters
+     * @return json string representation of parameters
```
secondsun: Add @throws to describe why the runtime exception may be thrown.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,206 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.security.PublicKey;
+import java.security.spec.InvalidKeySpecException;
+
+/**
+ * Credentials for OIDC based authentication
+ */
+public class OIDCCredentials implements ICredential {
```
secondsun: Add `@Override` where appropriate.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,206 @@
...
+    private final AuthState authState;
+    private final IIntegrityCheckParameters integrityCheckParameters;
+
+    /**
+     * OpenID Connect credentials containing the identity, refresh and access tokens provided on a
+     * successful authentication with OpenID Connect.
+     * @param serialisedCredential JSON string representation of the authState field produced by
+     *                             {@link #deserialize(String)}.
+     * @param integrityCheckParameters Integrity check parameters for the token.
+     */
```
secondsun: Add @throws to describe the runtime exception to the javadoc

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,206 @@
...
+            if (this.integrityCheckParameters != null) {
+                jsonCredential.put("integrityCheck", this.integrityCheckParameters.serialize());
+            }
+            return jsonCredential.toString();
+        } catch(JSONException e) {
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    public static OIDCCredentials deserialize(final String serializedCredential) {
```
secondsun: javadocs.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,206 @@
...
+        return isAuthorized() && getNeedsRenewal();
+    }
+
+    /**
+     * Renew the token
+     * @return
+     * @throws AuthenticationException
+     */
+    public boolean renew() throws AuthenticationException {
+        throw new IllegalStateException("Not yet implemented");
```
secondsun: Need implementation :)

auth/src/main/java/org/aerogear/android/ags/auth/utils/AuthStateManager.java
```diff
@@ -23,34 +23,32 @@ public AuthStateManager(final Context context) {
...
-        } catch (JSONException ex) {
-            return new OIDCCredentials();
-        }
+        return OIDCCredentials.deserialize(currentState);
     }
 
     /**
      * Saves a token
      * @param authState token to be saved
+     * @throws IllegalArgumentException
```
secondsun: Shouldn't this be illegal state exception?

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
aidenkeating: Not sure why this is showing up as changed, but the renew was there before this PR and I think it's outside the scope of this PR. I think its implementation is kind of dependent on knowing how login will work.

auth/src/main/java/org/aerogear/android/ags/auth/credentials/IIntegrityCheckParameters.java
aidenkeating: I think you're right about isValid requiring some docs here. For the others I don't think so, they're self explanatory I think.

auth/src/main/java/org/aerogear/android/ags/auth/AbstractAuthenticator.java
aidenkeating: Have made this change in a separate PR. Don't want this one to bloat much more. https://github.com/aerogear/aerogear-android-sdk/pull/24/files

aidenkeating:
@secondsun Dependencies were added in [this PR](https://github.com/aerogear/aerogear-parent/pull/78), waiting on a new release now. Have made a temporary commit that will fail so I don't forget

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
aidenkeating: `ICredential` is just empty at the moment.

auth/src/main/java/org/aerogear/android/ags/auth/utils/AuthStateManager.java
```diff
@@ -1,60 +1,68 @@
...
 public class AuthStateManager {
 
+    private static AuthStateManager instance = null;
     private static final String STORE_NAME = "org.aerogear.android.auth.AuthState";
     private static final String KEY_STATE = "state";
 
     private final SharedPreferences prefs;
 
-    public AuthStateManager(final Context context) {
+    private AuthStateManager(final Context context) {
```
rachael-oregan: I'm not sure if Singleton's are suppose to have parameters? This way couldn't the singleton be instantiated multiple times which defeats the purpose?

auth/src/main/java/org/aerogear/android/ags/auth/utils/AuthStateManager.java
```diff
@@ -1,60 +1,75 @@
...
      * Deletes a token
+     * @throws IllegalArgumentException
      */
     public synchronized void clear() {
         if (!prefs.edit().remove(KEY_STATE).commit()) {
             throw new IllegalStateException("Failed to clear state from shared preferences");
         }
     }
+
+    public static AuthStateManager getInstance(final Context context) {
```
aidenkeating: @rachael-oregan To respond to this comment https://github.com/aerogear/aerogear-android-sdk/pull/21/commits/260a5b358f4678d653b940a4d4fab4738b4d6bcc#r166298337 yeah I'm not sure the best way to approach this. I don't think providing a 'setter' of any kind is a good idea as then context could change if the setter is called multiple times. This can't be initialized multiple times but I can see the awkwardness of it.

This way context is bound to whatever is provided first, instead of being able to be set multiple times.

There's also the option of providing context on each function call I guess.

Perhaps @ziccardi or @secondsun have a more informed view on how to approach this.

auth/src/main/java/org/aerogear/android/ags/auth/utils/AuthStateManager.java
secondsun: Private constructor and prefs is a final variable.  IN this case it isn't bad because the references set up aren't mutable.  However I am generally -1 to static singletons because they have annoying behaviors when you are testing that you have to code around.

auth/src/main/java/org/aerogear/android/ags/auth/AuthService.java
```diff
@@ -75,7 +79,7 @@ public void setAuthenticatorChain(AuthenticationChain newChain) {
     public static synchronized AuthService getInstance() {
         if (INSTANCE == null) {
             // FIXME: load the configurations from core and pass it here
-            INSTANCE = new AuthService(null);
+            INSTANCE = new AuthService();
```
secondsun: We should NOT be using singletons.  References to services should be fetch through the mobile core and not plucked from the ether.

See https://github.com/aerogear/proposals/blob/master/android-sdk/README.md#referencing-a-service

aidenkeating: Yeah I don't think this is used anyway, doesn't seem to be referenced anywhere. Must have been left over. Will remove

aidenkeating:
@secondsun Removed unused singleton stuff and updated the android sdk bom stuff. Tests have started to pass again.

aidenkeating:
@secondsun Would you mind taking a look? One question, we currently have the auth module typed as `keycloak` because it matches up with the actual keycloak service. We also have a keycloak module in this SDK. I'm guessing them both being typed as `keycloak` is a bad thing?

danielpassos:
@aidenkeating Keycloak Service is a temporary one, It was to help us in the core PoC and going to be removed

aidenkeating:
@danielpassos @ziccardi @secondsun Would you mind taking a look? I think this is ready to merge

auth/src/main/java/org/aerogear/android/ags/auth/credentials/OIDCCredentials.java
```diff
@@ -0,0 +1,215 @@
...
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    /**
+     * Check whether the user is authorized and not expired.
+     * @return <code>true</code> if user is authorized and token is not expired.
+     */
+    public boolean checkValidAuth() {
+        return isAuthorized() && getNeedsRenewal();
```
ziccardi: Shouldn't be `return isAuthorized() && !getNeedsRenewal();` ?

auth/src/main/java/org/aerogear/android/ags/auth/utils/AuthStateManager.java
ziccardi: @aidenkeating Just a proposal: we could add a `bootstrap` method with package level visibility. It could then be called from the `bootstrap` method of the AuthService. This way other classes will see only a `getInstance` with no parameters. WDYT?

aidenkeating:
@ziccardi Mind taking a look? Have move the AuthStateManager out to the `auth` package. Have also included an `init(Context)` function in the `AuthService` to allow the user to provide the context. This is following the example used in this PR https://github.com/aerogear/aerogear-android-sdk/pull/27/files#diff-7037942ee1f06907ac7d09f6d65ccd6dR88

auth/src/main/java/org/aerogear/android/ags/auth/AuthService.java
```diff
@@ -1,35 +1,30 @@
...
 /**
  * Entry point for authenticating users.
  */
-public class AuthService {
-
-    /**
-     * Authentication service singleton.
-     */
-    private static AuthService INSTANCE;
+public class AuthService implements ServiceModule {
```
rachael-oregan: I presume this won't compile because its not implementing the methods from `ServiceModule` right?

Also how is the AuthService object got now?

aidenkeating: It is. Well CircleCI tests are passing. Jenkins is failing because the gradle daemon decided to go ahead and stop.

The ServiceModule interface has changed in this PR https://github.com/aerogear/aerogear-android-sdk/pull/22

rachael-oregan: Ah right cool, thanks for explaining

aidenkeating:
@wtrocki Would you mind taking a look? I *think* this is ready to merge. Don't want this PR to bloat too much more. Jenkins appears to be failing because of an internal issue

wtrocki:
I have spent some time checking changes on local machine. 
PR recieved pretty good feedback. Considering that this changes cannot be properly tested until we get login implementation I think it's best to get that into master and follow up with other PR's. 

Really good job @aidenkeating

danielpassos:
I didn't have time to review it, I was about to start now but it's already merged. I didn't test it but I have 2 comments about this PR:

1. Correct or not all other modules are using `org.aerogear.mobile.[module-name]` as the package
1. I think all method who going to throw the exception needs to have that explicit in the method not only in the Javadoc.

aidenkeating:
@danielpassos The package name was to address this comment https://github.com/aerogear/aerogear-android-sdk/pull/21#discussion_r165096520 and this proposal https://github.com/aerogear/proposals/blob/master/sdks/SDK-naming.md#proposed-solution

PR #22, opened 2018-02-02
Kill ServiceModuleRegistry with fire

danielpassos:
Relates issues: 

* [AGDROID-712](https://issues.jboss.org/browse/AGDROID-712)

This is a proposal to kill the requirement of register all ServiceModule `ServiceModuleRegistry.getInstance().registerServiceModule` and also the responsibility of `MobileCore` to instantiate all `ServiceModule`

Instead of the developer need to do that:

```java
static {
        ServiceModuleRegistry.getInstance().registerServiceModule("keycloak", KeyCloakService.class);
        ServiceModuleRegistry.getInstance().registerServiceModule("echo-service", EchoServiceModule.class, "keycloak");
}
```

They will do:

```java
MobileCore.init();

KeyCloakService kc = KeyCloakService.getInstance();
```

Services don't need to be registered anymore and it will be self managed


keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
```diff
@@ -28,30 +23,22 @@
...
-        this.audience = config.getProperty("audience");
-        this.grantType = config.getProperty("grant_type");
-        this.subjectTokenType = config.getProperty("subject_token_type");
-        this.requestedTokenType = config.getProperty("requested_token_type");
-        this.resource = config.getProperty("resource");
-        this.realm = config.getProperty("realm");
-        this.core = core;
-        this.httpModule = (HttpServiceModule) core.getService("http");
+    public static KeyCloakService create() {
+        ServiceConfiguration config = MobileCore.getServiceConfiguration("keycloak");
```
wtrocki: Move keycloak to constant

wtrocki:
Verifying that locally and building WIP

core/src/main/java/org/aerogear/mobile/core/ServiceModule.java
```diff
@@ -1,15 +1,4 @@
 package org.aerogear.mobile.core;
 
-import org.aerogear.mobile.core.configuration.ServiceConfiguration;
-
-
 public interface ServiceModule {
```
wtrocki: [Optional] This interface may no longer be needed.

danielpassos: Yeap, maybe in the future, let's leave it for now only to _mark_ the class is a _Service_

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
danielpassos: Done!

wtrocki:
Sample app available here: https://github.com/secondsun/WipDemo/compare/master...wtrocki:singleton-wtr

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -2,273 +2,153 @@
...
-        for (String serviceName : workingList) {
-            List<String> dependencies = serviceRegistry.getDependenciesFor(serviceName);
-            serviceUnderInstanciation = serviceName;
-            if (dependencies.isEmpty() || sortedServices.containsAll(dependencies)) {
-                unresolvableServiceDetected = false;
-                break;
+        // -- Setting default http layer
+
+        if (options.getHttpServiceModule() == null) {
+            ServiceConfiguration configuration = this.servicesConfig.get("http");
```
wtrocki: Configuration will not have http entry. It's there only because it was added manually. 
Suggestion is to remove this as it may be misleading.

wtrocki:
@danielpassos This changes are great. 
Once we merge that we could finally document Android SDK for developers the same way as we have done for IOS and provide sample applications.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
danielpassos: I left that because @secondsun are using it on his demo app. @secondsun any opinion about this?

wtrocki: @danielpassos Good point. We may need to write mini demo and documentation for this again.

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
```diff
@@ -28,30 +25,34 @@
...
-        this.realm = config.getProperty("realm");
-        this.core = core;
-        this.httpModule = (HttpServiceModule) core.getService("http");
+    public static KeyCloakService getInstance() {
+        ServiceConfiguration config = MobileCore.getServiceConfiguration("keycloak");
+        KeyCloakServiceConfiguration keyCloakServiceConfiguration = new KeyCloakServiceConfiguration(config);
+        return getInstance(keyCloakServiceConfiguration);
     }
 
+    public static KeyCloakService getInstance(KeyCloakServiceConfiguration keyCloakServiceConfiguration) {
```
pb82: Do we need to make this public? Why would the user want to override the `keyCloakServiceConfiguration` here? This would decouple it from mobile-config.json.

wtrocki: @pb82 that is requirement derived from current work. Developers may provide path to certificate files etc and extend configuration.

danielpassos: Like @wtrocki said, you should need to add more config not coming from the JSON file. But we can create some sets for it, if people don't agree to have it. @aidenkeating wdyt?

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
secondsun: So the theory was we need an http provider, we need to be able to configure it, hey let's use configuration from mobile-services.  It is probably not a great idea, but I can see configuration being passed in from the build plugins based on various configuration options.

wtrocki:
@pb82 I'm working now on documenting and testing this aproach as part of https://issues.jboss.org/browse/AGDROID-721

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
wtrocki: @secondsun - That is still valid use case. Going to document that in separate PR.

pb82:
@wtrocki sounds good 👍

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
```diff
@@ -28,30 +25,34 @@
...
-        this.audience = config.getProperty("audience");
-        this.grantType = config.getProperty("grant_type");
-        this.subjectTokenType = config.getProperty("subject_token_type");
-        this.requestedTokenType = config.getProperty("requested_token_type");
-        this.resource = config.getProperty("resource");
-        this.realm = config.getProperty("realm");
-        this.core = core;
-        this.httpModule = (HttpServiceModule) core.getService("http");
+    public static KeyCloakService getInstance() {
+        ServiceConfiguration config = MobileCore.getServiceConfiguration("keycloak");
```
secondsun: So one of the things we have been talking about is having multiple instance of the same service type, so I'm a bit iffy on this section right here.  If we have multiple "keycloak typed" services this won't resolve and there isn't a fallback way to instanciate them by name.

keycloak-service-module/src/main/java/org/aerogear/mobile/keycloak_service_module/KeyCloakService.java
secondsun: @pb82 If we have multiple "keycloak typed" services this won't resolve and there isn't a fallback way to instantiate them by name.  This really doesn't solve the problem, but we will have to have something like the method here in order to do it in this paradigm

secondsun:
ETOOMANYSINGLETONS
So I don't think that managing the lifecycle of service modules is something the service modules themselves should have to be aware of.  By removing the service registry we lose a single point of reference for starting them up and shutting them down.  This makes testing a lot harder as well as having more interesting integrations with the android lifecycle system.

danielpassos:
@secondsun @wtrocki I have made it a little less singleton and make `MobileCore` have more power of the services especially with destroy it and make the developer have an entry point to do some stuff when it occurred

_MyAppApplication_
```java
public class MyAppApplication extends Application {

    private MobileCore core;

    public MobileCore getCore() {
        return core;
    }

    @Override
    public void onCreate() {
        super.onCreate();

        core = MobileCore.init(this);
    }

    @Override
    public void onTerminate() {
        super.onTerminate();

        core.destroy();
    }

}
```

_MyActivity_
```java
public class MyActivity extends AppCompatActivity {

    private ServiceModule serviceModule;

    @Override
    public void onCreate(@Nullable Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        MyAppApplication application = (MyAppApplication) getApplication();
        MobileCore core = application.getCore();
        KeyCloakService service = (KeyCloakService) core.getInstance(KeyCloakService.class);
    }
}
```

wtrocki:
Last commit reverts back to service registry - instances managed by core, which is something that is on the master already. The main purpose of this PR was to revert that, which was posted on the mailing list.
We can remove singletons without moving object creation to core. I already started documenting current approach, providing examples and fitting into SDK use cases - that will make SDK more solid and friendly for community.

I'm not sure why we reverting changes/entire api when having PR already approved by other developers etc. 
This is completely different idea (not simple PR fix or rewrite etc.) Every opinion matters, but what I believe is happening here is that we trying to influence entire SDK by single opinion every time we want to innovate, without listening to other arguments and requirements. New usage is more or less in the master and was proven to be controversial and hard to document (API has limitations etc.) Why we cannot try different aproach and let SDK naturally evolve, with meritocracy and community focus. With this PR we followed patterns used in the major realtime database provider so we in fact rely on something solid and standard for community - which are our developers.

IMHO We should iterate with freedom of ideas. Many different use cases may shape our core SDK to different directions (see Cordova and ReactNative plugins) and we cannot have dictatorship model at the moment as no one will be able to get full overview of the solution we do considering multiple streams.

wtrocki:
Last commit reverts back to service registry - instances managed by core, which is something that is on the master already. The main purpose of this PR was to revert that, which was posted on the mailing list.
We can remove singletons without moving object creation to core. I already started documenting current approach, providing examples and fitting into SDK use cases - that will make SDK more solid and friendly for community.

I'm not sure why we reverting changes/entire api when having PR already approved by other developers etc.
This is completely different idea (not simple PR fix or rewrite etc.) Every opinion matters, but what I believe is happening here is that we trying to influence entire SDK by single opinion every time we want to innovate, without listening to other arguments and requirements. New usage is more or less in the master and was proven to be insufficient. We just trying to try different aproach and let SDK naturally evolve, with meritocracy and community focus. With this PR we followed patterns used in the major realtime database provider so we in fact rely on something solid and standard for community - which are our developers.

IMHO We should iterate with freedom of ideas. Many different use cases may shape our core SDK to different directions (see Cordova and ReactNative plugins) and we cannot have dictatorship model at the moment as no one will be able to get full overview of the solution we do considering multiple streams.

secondsun:
@wtrocki Show me where ServiceModuleRegistry.java comes back.  As far as I can tell that is not what this PR does.

core/src/main/java/org/aerogear/mobile/core/ServiceModule.java
```diff
@@ -2,14 +2,25 @@
...
+     */
+    String type();
+
+    /**
+     * A method how create/configure the service
+     *
+     * @param serviceConfiguration
      */
-    void bootstrap(MobileCore core, ServiceConfiguration configuration);
+    void configure(ServiceConfiguration serviceConfiguration);
```
secondsun: Let's pass core here.  This way modules can reference modules they depend on by using MobileCore#getInstance($Module.class).  In getInstance modules are already lazy loaded.  The only catch is that circular dependencies will trigger a stack overflow exception.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpServiceModule.java
```diff
@@ -21,25 +16,39 @@
...
-        this.httpServiceConfiguration = new HttpServiceConfiguration(configuration);
-        OkHttpClient.Builder clientBuilder = new OkHttpClient.Builder();
+    public String type() {
+        return "http";
+    }
 
-        this.client = clientBuilder.build();
+    @Override
+    public void configure(ServiceConfiguration serviceConfiguration) {
+        this.httpServiceConfiguration = new HttpServiceConfiguration(serviceConfiguration);
```
secondsun: We should instanciate a client here as well.  There may be information on the ServiceConfiguration which is relevant to OKHTTP

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -2,273 +2,174 @@
...
-        String serviceUnderInstanciation = "";
-
-        for (String serviceName : workingList) {
-            List<String> dependencies = serviceRegistry.getDependenciesFor(serviceName);
-            serviceUnderInstanciation = serviceName;
-            if (dependencies.isEmpty() || sortedServices.containsAll(dependencies)) {
-                unresolvableServiceDetected = false;
-                break;
-            }
+    public ServiceModule getInstance(Class<? extends ServiceModule> serviceClass,
```
secondsun: How do we resolve concrete vs abstract super types?

For instance if we have a OkHTTPModule in services, but someone calls getInstance(HttpModule.class) this will fail because HttpModule is a interface.

danielpassos:
@secondsun Update with the last suggestions and rebased

PR #23, opened 2018-02-03
Move library dependencies to the BOM

danielpassos:
Related issues:

* [AGDROID-714](https://issues.jboss.org/browse/AGDROID-714)

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -29,11 +29,9 @@ android {
 }
 
 dependencies {
-    implementation fileTree(dir: 'libs', include: ['*.jar'])
-    implementation 'net.openid:appauth:0.7.0'
-    testImplementation 'org.mockito:mockito-core:2.10.0'
-    testImplementation 'junit:junit:4.12'
-}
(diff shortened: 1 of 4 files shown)


wtrocki:
Verified by local build. Tried in sample application with duplicated dependencies.
----
[Offtopic] I noticed that there are two names for bom
https://mvnrepository.com/artifact/org.jboss.aerogear/aerogear-android-bom
https://jcenter.bintray.com/org/jboss/aerogear/aerogear-android-sdk-bom

It will be cool if we align that with names from proposal, but it may be a lots of hassle at this point.
https://github.com/aerogear/proposals/blob/master/sdks/SDK-naming.md

danielpassos:
@wtrocki `aerogear-android-bom` was used in the old AeroGear maven library (some years ago), `aerogear-android-sdk-bom` was re introduced to the aerogear-android-sdk

secondsun:
LGTM, I see a green bar and a happy @wtrocki .

PR #24, opened 2018-02-06
Update service configuration to be immutable

aidenkeating:
## Motivation

Currently the service configuration that is parsed from the JSON
configuration file for each service can be modified after its
creation. This means that after the configuration has been provided
to a module the configuration can still change, even if the module
itself makes no modifications directly. This can result in unexpected
behaviours.

## Description

This change makes the service configuration immutable. There is also
a smaller change to remove the HTTP module service configuration which
extends but makes no changes to the normal service configuration.

## Progress

- [x] Make ServiceConfiguration immutable


aidenkeating:
@ziccardi Would you mind taking a look?

core/src/main/java/org/aerogear/mobile/core/configuration/ServiceConfiguration.java
```diff
@@ -35,28 +76,19 @@ public String getProperty(String name) {
         return properties.get(name);
```
ziccardi: Does this compile?
`properties` is a `Map<String,String>`, so `properties.get(name)` returns a string, however the method returns a `Map<String,String>`,

aidenkeating: Perhaps that was the way GitHub was excluding some sections. If I expand this block this is the code

```java
    public Map<String, String> getProperties() {
        return Collections.unmodifiableMap(properties);
    }

    public String getProperty(String name) {
        return properties.get(name);
    }
```

ziccardi: My bad ;)

aidenkeating:
@grdryn Seeing the Jenkins build failing with

```
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```

any way to find out a bit more detail? Not getting the same thing locally or on CircleCI.

PR #27, opened 2018-02-07
Add Metrics

pb82:
## Motivation

Initial version of the android metrics sdk.

JIRA: [AGDROID-719](https://issues.jboss.org/browse/AGDROID-719)

## Description

Example Usage:

```
    @Override
    protected void onCreate(Bundle savedInstanceState) {
        ...
        MetricsService metrics = this.core.getInstance(MetricsService.class);
        metrics.init(getApplicationContext());
        ...
    }
```

## Progress

- [x] Initial POC
- [x] Reponse handling and logging
- [x] Tests


metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @param context Android application context
+     */
+    public void init(Context context) {
+        try {
+            // Send request to backend
+            HttpRequest request = httpService.newRequest();
+            request.post(config.getUri(), getMetricsData(context));
+            request.execute();
+            // TODO: response callback and response status logging
```
pb82: @secondsun I tried
```
            final HttpResponse response = request.execute();
            response.onComplete(new Runnable() {
                @Override
                public void run() {
                    Log.d(LOG_TAG, "status: " + response.getStatus());
                }
            });
```
but that results in a null pointer exception when it's trying to get the status code from the response. What's the recommended way to use the http module in such a case? It looks like it's already executing the http requests on the network thread, so no need to use AsyncTask?

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,112 @@
...
+import org.aerogear.mobile.core.http.HttpRequest;
+import org.aerogear.mobile.core.http.HttpServiceModule;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.UUID;
+
+public class MetricsService implements ServiceModule {
+    public final static String LOG_TAG = "AEROGEAR/METRICS";
+    private final static String STORAGE_NAME = "aerogear-metrics";
```
wei-lee: I would normally suggest using the full package name to avoid potential naming conflicts.

build.gradle
```diff
@@ -33,6 +33,14 @@ subprojects {
     }
 }
 
+// We want to expose the SDK version and name to the metrics subproject
+project(':metrics') {
```
danielpassos: I think it should be the core responsibility

metrics/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+android {
+    compileSdkVersion project.ext.targetSdkVersion
+
+    defaultConfig {
+        minSdkVersion project.ext.minSdkVersion
+        targetSdkVersion project.ext.targetSdkVersion
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+
+        // Required for metrics, exposed by the parent project
+        versionName project.ext.versionName
```
danielpassos: I think it should be the core responsibility

metrics/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+}
+
+dependencies {
+    implementation project(path: ':core')
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
```
danielpassos: It should be removed

metrics/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+
+    implementation 'com.android.support:appcompat-v7:26.1.0'
+    testImplementation 'junit:junit:4.12'
+    androidTestImplementation 'com.android.support.test:runner:1.0.1'
+    androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.1'
```
danielpassos: Version is coming from the BOM and should be removed from here

metrics/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="mobile.aerogear.org.metrics" />
```
danielpassos: We are using a different kind of package (correctly or not) in the other modules `org.aerogear.mobile.[module-name]`

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsConfig.java
```diff
@@ -0,0 +1,19 @@
+package mobile.aerogear.org.metrics;
+
+import org.aerogear.mobile.core.configuration.ServiceConfiguration;
+
+public class MetricsConfig {
```
danielpassos: Why we need this if ServiceConfiguration already has the URI?

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,122 @@
...
+        return clientId;
+    }
+
+    /**
+     * Get the version of the app itself
+     *
+     * @param context Android application context
+     * @return String version name
+     */
+    private String getAppVersion(final Context context) {
```
danielpassos: I think it should be the core responsibility

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,122 @@
...
+        return result.toString().getBytes();
+    }
+
+    /**
+     * Send the metrics data to the server. The data is contained in a JSON object with the
+     * following properties: clientId, appId, sdkVersion and appVersion
+     *
+     * @param context Android application context
+     */
+    public void init(final Context context) {
```
danielpassos: We should raise a RuntimeException here if we got some problem. I'd like to suggest to use `InitializationException`

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,122 @@
...
+                }
+            });
+        } catch (JSONException e) {
+            logger.error(LOG_TAG, e);
+        }
+    }
+
+    @Override
+    public void configure(MobileCore core, ServiceConfiguration serviceConfiguration) {
+        config = new MetricsConfig(serviceConfiguration);
```
danielpassos: I think we can use ServiceConfiguration instead of creating a new one

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    public String type() {
+        return MODULE_NAME;
+    }
+
+    @Override
+    public void destroy() {
+        // Not used
```
danielpassos: I think the comment is not necessarily, and it can be a empty method

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,122 @@
...
+            return context
+                .getPackageManager()
+                .getPackageInfo(context.getPackageName(), 0).versionName;
+        } catch (PackageManager.NameNotFoundException e) {
+            logger.error(LOG_TAG, e);
+            return null;
+        }
+    }
+
+    private byte[] getMetricsData(final Context context) throws JSONException {
```
danielpassos: Only a personal opinion it can be `createMetricsData` or only `metricsData`, but again it's only a personal thought

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,122 @@
...
+     * @param context Android application context
+     * @return String version name
+     */
+    private String getAppVersion(final Context context) {
+        try {
+            return context
+                .getPackageManager()
+                .getPackageInfo(context.getPackageName(), 0).versionName;
+        } catch (PackageManager.NameNotFoundException e) {
+            logger.error(LOG_TAG, e);
```
danielpassos: Not sure if's going to happen but I think we should raise an Exception instead of return null

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,122 @@
...
+import org.aerogear.mobile.core.http.HttpServiceModule;
+import org.aerogear.mobile.core.logging.Logger;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.UUID;
+
+public class MetricsService implements ServiceModule {
+    private final static String MODULE_NAME = "metrics";
+    private final static String LOG_TAG = "AEROGEAR/METRICS";
```
danielpassos: s/LOG_TAG/TAG

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
pb82: Hmm, not sure. If the metrics request fails for some reason, do we really want to crash the whole app? Maybe the the backend is fully functional and only the metrics server has been stopped. In that case the apps should still work.

danielpassos: Makes sense

aliok: IMO this depends on how we call this `init` method. Perhaps we can still throw the exception and let the caller decide what to do with it.
Looked for callers, couldn't find any at the moment. How is this supposed to be called?

metrics/build.gradle
wtrocki: Using Core for versioning on IOS

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsConfig.java
wtrocki: +1 for having that as it may include security tokens and other stuff in near future. 
It's simple at the moment. I have the same on IOS

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
pb82: @aliok At the moment the user has to call this explicitely in their App. We discussed the option of making this implicit, but decided against it for now.

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsConfig.java
danielpassos: So, do it in the near future, not now :)

We should not include stuff we don't need now IMO

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
danielpassos: I really don't have any strong opinion on this. I'd like to hear @wei-lee @wtrocki and @secondsun

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Get or create the client ID that identifies a device as long as the user doesn't
+     * reinstall the app or delete the app storage. A random UUID is created and stored in the
+     * application shared preferences.
+     *
+     * @param context Android app context
+     * @return String Client ID
+     */
+    private String getOrCreateClientId(final Context context) {
```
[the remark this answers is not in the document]
danielpassos: Very nice catch!!!

metrics/src/main/java/mobile/aerogear/org/metrics/MetricsService.java
wtrocki: +1 for collecting metrics manually now. I will be more happy to change method name to something similar to IOS implementation. For example `collectStartupMetrics` etc.

wtrocki:
See also IOS Sdk implementation: https://github.com/aerogear/aerogear-ios-sdk/pull/13

gradle-mvn-push.gradle
```diff
@@ -103,7 +103,7 @@ afterEvaluate { project ->
             source = variant.javaCompile.source
             ext.androidJar = project.files(android.getBootClasspath().join(File.pathSeparator))
             classpath = files(variant.javaCompile.classpath.files) + files(ext.androidJar)
-            exclude '**/BuildConfig.java'
+            exclude '**/BuildConfig'
```
pb82: For some reason those exclusions don't work with `.java` and gradle gives out warnings.

metrics/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
pb82: This allows Mockito to mock final classes and methods

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
```diff
@@ -0,0 +1,120 @@
...
+import org.aerogear.mobile.core.http.HttpResponse;
+import org.aerogear.mobile.core.http.HttpServiceModule;
+import org.aerogear.mobile.core.logging.Logger;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.UUID;
+
+public class MetricsService implements ServiceModule {
+    public final static String STORAGE_NAME = "mobile.aerogear.org.metrics";
```
danielpassos: s/mobile.aerogear.org.metrics/org.aerogear.mobile.metrics

danielpassos:
@pb82 Just need a very small fix and we are good to go. Awesome work here man.

danielpassos:
@pb82 Just need a very small fix and we are good to go. Awesome work here man.

aliok:
I haven't actually tried running the code, but looks good to me 👍 
I still have the feeling that we should use `protected` for almost every `private` thing in order to allow extension of things, but this is something that could be done anytime.

pb82:
@danielpassos Changed the storage name to the actual package name
@aliok Yes that can be done at any time. I also feel that we need to think about how we want people to extend the SDK. Right now they could override those methods in their own implementation but then, what do they do with it? There must be a documented way to register your own implementation in the SDK (core is responsible for creating the instances).

PR #28, opened 2018-02-07
Add logger in the MobileCore

danielpassos:
Related issues:

* [AGDROID-727](https://issues.jboss.org/browse/AGDROID-727)


danielpassos:
@wtrocki I only added a test to check if the default implementation and whatever new implementation are going to be correctly returned by MobileCore. I have no idea **how** to test the real implementation :confused:

PR #29, opened 2018-02-08
Add example app

danielpassos:
Relates issue: 

* [AGDROID-729](https://issues.jboss.org/browse/AGDROID-729)


danielpassos:
@secondsun @wtrocki @pb82 @aidenkeating I think it can help us to test what we are doing and a good way to show how to use those stuff.

settings.gradle
```diff
@@ -1 +1 @@
-include ':core', ':keycloak-service-module', ':auth'
```
wtrocki: I'm not sure if we want to have example as top level gradle project or just be completely standalone - which will more or less simulate how end users will use this library

Current aproach looks ok for people to try this out and it's ok but raising to get other opinions.

danielpassos: I can't see any problem to have the sample here. Most of the libraries do the same

See: https://github.com/facebook/facebook-android-sdk/blob/master/settings.gradle

Anyway, I have no problem to remove it, if you want!

danielpassos: I can't see any problem to have the sample here. Most of the libraries do the same

See: https://github.com/facebook/facebook-android-sdk/blob/master/settings.gradle

Anyway, I have no problem to remove it, if you want!

danielpassos:
@AdamSaleh can you take a look at the Jenkins build?

settings.gradle
aliok: @danielpassos @wtrocki I think having it separate would make users' lives easier. They would just clone the example app and modify it.

wtrocki: I do not have strong opinion about that. Just wanted to flag that and see other opinions. IMHO we should merge that PR as it will block Ali documentation work

pb82: @aliok @wtrocki @danielpassos hmm, on the other hand i always like it when libraries have an `example` directory. Isn't @secondsun also working on a separate kitchensink app?

wtrocki: @pb82 This was just initial app for demo purposes. This one may be be more or less documentation how to use libraries and it should be very simple

danielpassos: @aliok This is what we call cookbook or template, we are also going to create a separated one for that, this is more for us playing with features.

aliok: sounds good :)

PR #32, opened 2018-02-13
Add auth UI boilerplate

aidenkeating:
## Motivation

Have some initial UI to start with when integrating the auth module into the example app. This is taken from https://github.com/feedhenry/mobile-security-android-template

## Description

Moves views from the security template to this example app.

## Progress

- [x] Move views


aidenkeating:
@wei-lee @rachael-oregan @ziccardi @danielpassos Would you mind taking a look?

PR #33, opened 2018-02-13
Refactor auth package

aidenkeating:
## Motivation

Refactor auth packages to be inline with the rest of the SDK, not the proposal https://github.com/aerogear/proposals/pull/17/files#diff-18f11f90545248c2674f9c32745d3d6cR19

## Description

Refactor package names

## Progress

- [x] Refactor


wei-lee:
so should we update the proposal then?

aidenkeating:
@wei-lee Yes, I'm doing that now.

Jenkins seems to be an internal issue

PR #34, opened 2018-02-14
Auth module documentation

aidenkeating:
## Motivation

Documentation is good

JIRA: https://issues.jboss.org/browse/AGDROID-669

## Description

Added documentation for auth module.

## Progress

- [x] Documentation


aidenkeating:
@wei-lee @TommyJ1994 @rachael-oregan Mind taking a look? Most places like `AuthService#currentUser` are intended to be linked to the JavaDoc once it's ready.

docs/auth/README.adoc
```diff
@@ -13,3 +13,118 @@ dependencies {
 
 == Usage
 
+To use the Auth SDK you'll first need to import the Core module. See link:../core/README.adoc[the Core documentation].
```
wei-lee: I think it should link to this doc: https://github.com/aerogear/aerogear-android-sdk/blob/master/docs/getting-started.adoc

docs/auth/README.adoc
```diff
@@ -13,3 +13,118 @@ dependencies {
...
+
+=== Authenticating
+
+To start the authentication invoke the `AuthService#performAuthRequest` method.
```
wei-lee: I think this needs to be updated to reflect the right name of the method. It should be `login`?

docs/auth/README.adoc
```diff
@@ -13,3 +13,118 @@ dependencies {
...
+
+[source,java]
+----
+static int LOGIN_RESULT_CODE = 1;
+Activity myActivity = this.getActivity();
+
+// Initialise the auth service.
+AuthService authService = mobileCore.getInstance(AuthService.class);
+authService.init(context);
```
wei-lee: see https://github.com/wei-lee/aerogear-android-sdk/blob/77281e718bc4876a96eed93205e3d57aacbb1b31/example/src/main/java/org/aerogear/mobile/example/ui/MainActivity.java#L52-L59

docs/auth/README.adoc
wei-lee: ah, never mind. It is for the core module.

wei-lee: I think we need at least 2 prerequisites steps:

1. import the core module.
2. specify the redirect uri. It should be specified in `android.defaultConfig.manifestPlaceholders` of the build.gradle file (see https://github.com/aerogear/aerogear-android-sdk/blob/master/example/build.gradle#L17). This value should also be the one of the "valid redirect URIs" configuration in Keycloak

aidenkeating:
@wei-lee I think I addressed all those comments

docs/auth/README.adoc
```diff
@@ -13,3 +13,128 @@ dependencies {
...
+AuthService authService = (AuthService) mobileCore.getInstance(AuthService.class);
+
+AuthServiceConfiguration authServiceConfig = new AuthServiceConfiguration.AuthConfigurationBuilder()
+    .withRedirectUri("org.aerogear.mobile.example:/callback")
+    .allowSelfSignedCertificate(true)
+    .build();
+
+authService.init(context, authServiceConfig);
+
+// Build the credentials object and start the authentication flow.
```
wei-lee: replace `credentials` with `options`

docs/auth/README.adoc
```diff
@@ -13,3 +13,128 @@ dependencies {
...
+OIDCAuthenticateOptions credentials = new OIDCAuthenticateOptions(myActivity, LOGIN_RESULT_CODE);
```
wei-lee: better not call it `credentials`

docs/auth/README.adoc
```diff
@@ -13,3 +13,128 @@ dependencies {
...
+
+Callback authCallback = new Callback<Principal>() {
```
wei-lee: I believe the type is changed. It is now `UserPrincipal`

docs/auth/README.adoc
```diff
@@ -13,3 +13,128 @@ dependencies {
...
+        // The core will return the same instance of the auth service as before
+        AuthService authService = mobileCore.getInstance(AuthService.class);
+        authService.handleAuthResponse(data);
+    }
+}
+----
+
+The callback provided in `AuthService#login` will be invoked.
+
+=== Retrieving the current user
```
wei-lee: I say move this to the first and have the AuthService initialised here.  This is how most of the apps should use the module I think.

aidenkeating: Yeah that's a good point

docs/auth/README.adoc
```diff
@@ -13,3 +13,128 @@ dependencies {
...
+* Specify the authentication redirect URI in `android.defaultConfig.manifestPlaceholders` of the apps `build.gradle` file. See link:https://github.com/aerogear/aerogear-android-sdk/blob/master/example/build.gradle[example app build.gradle].
+* Specify the authentication redirect URI in the "valid redirect URIs" configuration of the Keycloak Client the app is authenticating with.
+
+=== Retrieving the current user
+
+To retrieve the current authenticated user the `AuthService#currentUser` method can be invoked.
+
+[source,java]
+----
+AuthService authService = mobileCore.getInstance(AuthService.class);
```
wei-lee: missing the init part here.

docs/auth/README.adoc
```diff
@@ -13,3 +13,128 @@ dependencies {
 
 == Usage
 
+To use the Auth SDK you'll first need to:
+
+* Import the Core module. See link:../core/README.adoc[the Core documentation].
+* Specify the authentication redirect URI in `android.defaultConfig.manifestPlaceholders` of the apps `build.gradle` file. See link:https://github.com/aerogear/aerogear-android-sdk/blob/master/example/build.gradle[example app build.gradle].
```
TommyJ1994: "authentication redirect URI" might not mean much to someone not familiar with OAuth based protocols. Maybe we can add a sentence to say that in Android, this should be the package name of your App, and let them know its required because the browser used to perform the login needs to know where to return to after authentication has taken place. WDYT?

TommyJ1994:
The existing doc states - "Mobile authentication SDK based on Keycloak". Can we update this to mention that its using OpenID Connect perhaps too, as Keycloak also supports SAML etc. Maybe also mention some of the features of the IDM SDK briefly here too - access control, 2FA via Keycloak etc WDYT?

docs/auth/README.adoc
aidenkeating: Addressed

docs/auth/README.adoc
```diff
@@ -13,3 +15,136 @@ dependencies {
 
 == Usage
 
+To use the Auth SDK you'll first need to:
+
+* Import the Core module. See link:../core/README.adoc[the Core documentation].
+* Specify the package name of your Android project in `android.defaultConfig.manifestPlaceholders` of the apps `build.gradle` file.
```
wei-lee: actually, the value there isn't necessary to be the `package name` of the android project. It is an option, but it really should be the schema of redirect url (and it is recommended to use the package name). See https://github.com/openid/AppAuth-Android#capturing-the-authorization-redirect.

docs/auth/README.adoc
```diff
@@ -13,3 +15,136 @@ dependencies {
...
+
+To retrieve the current authenticated user the `AuthService#currentUser` method can be invoked.
+
+[source,java]
+----
+AuthService authService = mobileCore.getInstance(AuthService.class);
+
+AuthServiceConfiguration authServiceConfig = new AuthServiceConfiguration.AuthConfigurationBuilder()
+    .withRedirectUri("org.aerogear.mobile.example:/callback")
+    .allowSelfSignedCertificate(true)
```
wei-lee: let's remove it this config for now as it promotes bad practice. We may not keep this API in the future when we have certificate pinning.

docs/auth/README.adoc
```diff
@@ -13,3 +15,136 @@ dependencies {
...
+----
+static int LOGIN_RESULT_CODE = 1;
+Activity myActivity = this.getActivity();
+
+// Initialise the auth service.
+AuthService authService = (AuthService) mobileCore.getInstance(AuthService.class);
+
+AuthServiceConfiguration authServiceConfig = new AuthServiceConfiguration.AuthConfigurationBuilder()
+    .withRedirectUri("org.aerogear.mobile.example:/callback")
+    .allowSelfSignedCertificate(true)
```
wei-lee: same here.

wei-lee: actually, let's not init the auth service here again. Let's add a comment here to say at this point, it should be already inited.

docs/auth/README.adoc
```diff
@@ -13,3 +15,136 @@ dependencies {
...
+[source,java]
+----
+AuthService authService = mobileCore.getInstance(AuthService.class);
+
+AuthServiceConfiguration authServiceConfig = new AuthServiceConfiguration.AuthConfigurationBuilder()
+    .withRedirectUri("org.aerogear.mobile.example:/callback")
+    .allowSelfSignedCertificate(true)
+    .build();
+
+authService.init(context, authServiceConfig);
```
wei-lee: add a comment here to say this only needs to be called once within an app.

docs/auth/README.adoc
```diff
@@ -13,3 +15,136 @@ dependencies {
...
+
+In order to check if a user has a specific role you can invoke the `UserPrincipal#hasClientRole`
+and `UserPrincipal#hasRealmRole` methods and provide the role name to check for.
+
+[source,java]
+----
+AuthService authService = mobileCore.getInstance(AuthService.class);
+UserPrincipal currentUser = authService.currentUser();
+
+bool hasAdminPermissions = currentUser.hasRealmRole("user_admin");
```
TommyJ1994: `boolean`?

docs/auth/README.adoc
```diff
@@ -13,3 +15,136 @@ dependencies {
...
+if (hasAdminPermissions) {
+    // Show some component.
+}
+
+// Check if a user has a role from a specific client named my_client.
+bool isModerator = currentUser.hasClientRole("my_client", "user_moderator");
```
TommyJ1994: `boolean`?

docs/auth/README.adoc
```diff
@@ -13,3 +15,136 @@ dependencies {
...
+    }
+}
+----
+
+The callback provided in `AuthService#login` will be invoked.
+
+=== Retrieving a users roles
+
+Once a `UserPrincipal` has been retrieved the roles of the user can be listed and checked. This can
+be used to perform client side validation, such as hiding UI components related to actions the user
```
TommyJ1994: `client side access control` might be a better term to use.

docs/auth/README.adoc
aidenkeating: Woops, wrong language

docs/auth/README.adoc
aidenkeating: Finding this hard to describe properly in a short way, any suggestions on the wording? Like should we go as in depth as https://github.com/openid/AppAuth-Android#capturing-the-authorization-redirect

docs/auth/README.adoc
aidenkeating: Moved the initialization out to a separate section. Seems to fit better on its own.

docs/auth/README.adoc
TommyJ1994: If you want to redirect back to your app, then i think its the only option to use to package name - as this is used in the auth request so Keycloak knows what to redirect back to. If you specify something else that isn't your package name, you will get an exception when trying to build too:

```
Error:Execution failed for task ':auth:processDebugAndroidTestManifest'.
> Manifest merger failed : Attribute data@scheme at manifestMerger8608265398446142713.xml requires a placeholder substitution but no value for <appAuthRedirectScheme> is provided.
```

I think the AppAuth docs are just pointing out that the redirects for mobile are different for the web and you need to use a `com.myapp.example` instead of a regular `https://` url.

wei-lee: How about something like this:

Specify the schema of the redirect url for your Android project in `android.defaultConfig.manifestPlaceholders` of the apps `build.gradle` file. 
It should match the schema specified in the "Valid Redirect URL" section for the client in Keycloak. 
It is recommended to use the package name of the Android app as the schema of the redirect  url to avoid conflicts.

aidenkeating: Thanks @wei-lee @TommyJ1994 !

aidenkeating:
@ciaranRoche Would you mind taking a look also?

docs/auth/README.adoc
```diff
@@ -13,3 +15,149 @@ dependencies {
...
+// Check if a user has a role from a specific client named my_client.
+boolean isModerator = currentUser.hasClientRole("my_client", "user_moderator");
+if (isModerator) {
+    // Enable some button.
+}
+----
+
+=== Logging out
+
+To logout invoke the `AuthService#logout` method. This accepts the `UserPrincipal` that was
```
TommyJ1994: `To logout, invoke`

docs/auth/README.adoc
```diff
@@ -13,3 +15,149 @@ dependencies {
...
+
+To retrieve the current authenticated user the `AuthService#currentUser` method can be invoked.
+
+[source,java]
+----
+// authService already initialized.
+UserPrincipal currentUser = authService.currentUser();
+
+if (currentUser) {
+    // User is already authenticated, perform an action
```
TommyJ1994: Are there are actions/properties on the `currentUser` that we could show here so the end user developer knows what the currentUser contains?

wei-lee: They should be able to find out from the API or by using auto completion. But perhaps no harm to add something here.

wei-lee:
@TommyJ1994 

> Are docs for getAccessToken(), getIdentityToken(), isAuthorized() etc planned to be included in this PR, or a later PR?

I think there will be a later PR to show that.

docs/auth/README.adoc
```diff
@@ -13,3 +15,149 @@ dependencies {
...
+----
+
+=== Authenticating
+
+To start the authentication invoke the `AuthService#login` method.
+
+[source,java]
+----
+// authService already initialized.
+AuthService authService = mobileCore.getInstance(AuthService.class);
```
wtrocki: Hint. You probably want to avoid such extensive code snippets as this can be maintenance hell. Asciidocs allow you to reference and build snippets from actual code - so we can reference code in example application and it will work (sadly it's not working on github) 
Using links etc. will work as well. This way we will keep things updated all the time (as example app code is actually compiled). This was proven and tested in RainCatcher where we had rule to remove snippets longer than 10 lines

docs/auth/README.adoc
```diff
@@ -1,6 +1,8 @@
 = AeroGear Services Auth SDK
 
-Mobile authentication SDK based on Keycloak
+Mobile authentication SDK based on link:http://www.keycloak.org/[Keycloak] using link:http://openid.net/connect/[OpenID Connect].
+
+Provides authentication features like access control and two factor authentication through Keycloak.
 
 == Adding dependency
```
wtrocki: [Optional] I have added this section without checking that works. We will need probably bom update to enable it. Worth check

docs/auth/README.adoc
```diff
@@ -13,3 +15,149 @@ dependencies {
 
 == Usage
 
+To use the Auth SDK you'll first need to:
+
+* Import the Core module. See link:../core/README.adoc[the Core documentation].
```
wtrocki: I think initial idea for core was to be independent, but now core is part of auth:
https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/build.gradle#L30

Import is not needed - in fact it may cause issues.

docs/auth/README.adoc
```diff
@@ -13,3 +15,149 @@ dependencies {
...
+* Specify the schema of the redirect url for your Android project in `android.defaultConfig.manifestPlaceholders` of the apps `build.gradle` file.
+It should match the schema specified in the `Valid Redirect URL` section for the client in Keycloak.
+It is recommended to use the package name of the Android app as the schema of the redirect url to avoid conflicts.
+
+=== Initializing the module
```
wtrocki: People will see it as SDK (internally we see it as module)

wtrocki:
@aidenkeating Really clear and good documentation. I have tried that with knowledge of keycloak and it covers most major stuff. One comment from myself is that with such heavy code loaded documentation you would probably want to reference api documentation and example app snippets to avoid maintenance hell. 

Outside this PR - I also noticed that there is no interface returned for auth service - we iterate on actual service object. It may result in breaking user applications on updates that could be hard to track if developers will start changing things - Something to keep in mind, but doesn't require immediate action.

docs/auth/README.adoc
aidenkeating: Good point. Will add these

docs/auth/README.adoc
aidenkeating: I'm not sure I understand, according to https://github.com/aerogear/aerogear-android-sdk/pull/52/files#diff-27473e51ddc3d43d92d09cbd9da9e521R58 you'll need to import core and auth in order to use auth in your app? @secondsun Is that correct?

docs/auth/README.adoc
aidenkeating: I don't think these are intended to be snippets of the example app (though they currently are). So I don't think they're meant to keep in sync. The only times these should need to be updated is when the SDK API changes I would think. Though I agree a more automated approach would be good, I couldn't get the code snippet feature in AsciiDoc to function properly on GitHub.

docs/auth/README.adoc
wtrocki: @aidenkeating Yeah - I was expecting SDK to require core,
 but it seems that core will be available in the mobile app anyway.

@secondsun we could correct some cases in documentation.

docs/auth/README.adoc
wei-lee: good catch. @aidenkeating  see https://github.com/aerogear/aerogear-android-sdk/pull/52

docs/auth/README.adoc
wei-lee: For now, I think embed those code inside this document is fine. Those code are meant to demo the API and they should relatively stable. If in the future we find these code snippets need updates a lot, we should then consider using referenced code instead.

TommyJ1994:
@aidenkeating Do you think that the `Retrieving the current user` section should come after the `Authenticating` section, rather than before?

aidenkeating:
My thoughts on this came from https://github.com/aerogear/aerogear-android-sdk/pull/34#discussion_r168229222. In our example app we check the current user before attempting to authenticate. So I'd say it's ok to stay the same. Happy to swap if there are strong opinions on it though.

TommyJ1994:
@aidenkeating Maybe we can point this out that the currentUser can be used to check if authenticated?

I just found it a bit non-conventional that when reading down through the docs, it shows what you can do with an authenticated user first, but only shows you hot to authenticate after. But maybe its just me. Feel free to ignore.

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+
+Provides authentication features like access control and two factor authentication through Keycloak.
 
 == Adding dependency
 
 Add dependency to your application module
 ----
 dependencies {
-    implementation group: 'org.aerogear.android', name: 'auth'
+    implementation 'org.aerogear:auth:5.0.0-SNAPSHOT'
```
aidenkeating: @secondsun Is this the correct syntax?

wtrocki: @aidenkeating That format without bom file. With bom you ignoring versions part. 
Rest is ok.

aidenkeating:
@wei-lee Mind taking another look?

david-martin:
@aidenkeating I'd like to try out these docs & leave a review. I don't mind if it's merged (as there are others reviewign), but I'll leave comments if I have any feedback.

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
 = AeroGear Services Auth SDK
 
-Mobile authentication SDK based on Keycloak
+Mobile authentication SDK based on link:http://www.keycloak.org/[Keycloak] using link:http://openid.net/connect/[OpenID Connect].
+
+Provides authentication features like access control and two factor authentication through Keycloak.
 
 == Adding dependency
```
david-martin: @aidenkeating I'd suggest swapping this section with the Usage section i.e. show the Usage first for including the Core SDK. Maybe even have this section as part of Usage?

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+    implementation 'org.aerogear:auth:5.0.0-SNAPSHOT'
 }
 ----
 
 == Usage
 
+To use the Auth SDK you'll first need to:
+
+* Import the Core module. See link:./core.adoc[the Core documentation].
+* Specify the schema of the redirect url for your Android project in `android.defaultConfig.manifestPlaceholders` of the apps `build.gradle` file.
```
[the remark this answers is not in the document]
david-martin: Ah, I think I get it now.
It should be in here?
```groovy
android {
    defaultConfig {
        manifestPlaceholders ""
    }
```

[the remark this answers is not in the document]
aidenkeating: Good point. I'll link out to the example app. I had it linked before but must have removed it by mistake

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+If `AuthService#init` is not invoked then an `IllegalStateException` will be thrown when using the
+service.
+
+[source,java]
+----
+AuthServiceConfiguration authServiceConfig = new AuthServiceConfiguration.AuthConfigurationBuilder()
+    .withRedirectUri("org.aerogear.mobile.example:/callback")
+    .build();
+
+// Only invoke this once, every subsequent retrieval of the AuthService will retrieve the same, already initialized, instance.
```
david-martin: Maybe rephrase as 'You only need to invoke this once'.
'Only invoke this once' makes it sound dangerous, even though the end of the sentence seems to imply its fine to invoke it many times.

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+[source,java]
+----
+AuthService authService = mobileCore.getInstance(AuthService.class);
+----
+
+Any subsequent `MobileCore#getInstance` method will return the same instance of `AuthService`.
+
+Before the `AuthService` can be used used `AuthService#init` must be invoked once in an app.
+
+If `AuthService#init` is not invoked then an `IllegalStateException` will be thrown when using the
```
david-martin: When should i call init()?
Is there a particular thing I should be doing first?
If not, why even have an init? Could it not be automated?

david-martin: Nevermind. Its called out below

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+service.
+
+[source,java]
+----
+AuthServiceConfiguration authServiceConfig = new AuthServiceConfiguration.AuthConfigurationBuilder()
+    .withRedirectUri("org.aerogear.mobile.example:/callback")
```
david-martin: Why do I have to specify the redirect Uri here if I already added it to my build.gradle?

david-martin: Also, that uri in the code example looks weird `org.aerogear.mobile.example:/callback` is there a port missing? What about the protocol? Also, the host looks backwards

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+
+To start the authentication invoke the `AuthService#login` method.
+
+[source,java]
+----
+// authService already initialized.
+AuthService authService = mobileCore.getInstance(AuthService.class);
+
+static int LOGIN_RESULT_CODE = 1;
+Activity myActivity = this.getActivity();
```
david-martin: This line doesn't make sense to me. I'm adding this code to my activity, so I use `this`.
Does that mean I'm adding it in the wrong class?

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+    @Override
+    public void onError(Throwable error) {
+        // An error occurred during login.
+    }
+}
+
+authService.login(options, authCallback);
+----
+
+Once the browser returns to the app the result can be handled. In the activity provided to the
```
david-martin: Can we add a code example for what's meant to be done here?

I'm guessing its something like this in my Activity
```java
    @Override
    public void onActivityResult() {
        Intent intent = new Intent();
        authService.handleAuthResult();
    }
```

but that doesn't seem to be an overriden method. Was I suppose to 'implement' some extra interface?
Also, what exactly am I suppose to do in this method? What if I don't override it?

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+`AuthService#handleAuthResponse`, providing the `Intent`. This will exchange the temporary tokens
+returned from `AuthService#login` for long-life tokens and will provide a
+`UserPrincipal` which can be used to access a users details.
+
+More information about the user returned is available in link:../core/README.adoc[the auth module JavaDocs].
+
+[source,java]
+----
+@Override
+public void onActivityResult(int requestCode, int resultCode, Intent data) {
```
david-martin: When & why would I ever need a different implementation of this method?
What happens if I don't implement it?

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+        // The core will return the same instance of the auth service as before
+        AuthService authService = mobileCore.getInstance(AuthService.class);
+        authService.handleAuthResponse(data);
+    }
+}
+----
+
+The callback provided in `AuthService#login` will be invoked.
+
+=== Retrieving a users roles
```
david-martin: Could probably add some links to relevant keycloak docs here?

docs/getting-started/auth.adoc
```diff
@@ -1,15 +1,169 @@
...
+provided by `AuthService#handleAuthResponse`.
+
+[source,java]
+----
+// authService already initialized.
+AuthService authService = mobileCore.getInstance(AuthService.class);
+UserPrincipal currentUser = authService.currentUser();
+
+authService.logout(currentUser);
+----
```
david-martin: @aidenkeating I think 1 thing we're missing in the guide is an example mobile-services.json file, or something that explains how that file relates to the auth sdk.
It may not always be the case that the user of the auth sdk is using the mobile cli to generate that file, so showing an example file might be good.

docs/getting-started/auth.adoc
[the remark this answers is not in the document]
david-martin: yeah, the example app has it like this

```groovy
        manifestPlaceholders = ['appAuthRedirectScheme': 'org.aerogear.mobile.example']
```

No harm in linking & maybe even having an example here

docs/getting-started/auth.adoc
aidenkeating: onActivityResult would handle more than just our single result in a real app. As for why our specific `if` block would be different, apart from logging I doubt it would.

There is an explanation above about what this does. However, I can add more information about what happens if you don't do it.

rachael-oregan:
@aidenkeating does it make sense to have a section in here explaining Keycloak setup, specific setup in the Keycloak admin console and how that can work with the Auth SDK?  See some of the setup I've documented [here](https://github.com/aerogear/aerogear-android-sdk/pull/36/files#diff-0fa64fa2a56ea5ca870374b54c69ca49R12) and [here](https://github.com/aerogear/aerogear-android-sdk/pull/36/files#diff-0fa64fa2a56ea5ca870374b54c69ca49R29).  That way the SSO docs can just refer to these docs as a prerequisite.  What do you think?

aidenkeating:
@rachael-oregan I have added the section for setting up on OpenShift and linking to standalone. I think the manual creation of clients should be left in the SSO documentation though. The Keycloak APB creates a client so the user shouldn't have to in 'normal circumstances'.

PR #35, opened 2018-02-14
AGDROID-671 - Create API documentation for the Android IDM SDK

ciaranRoche:
## Motivation

https://issues.jboss.org/browse/AGDROID-671

## Description

A gradle task was added to the auth module, so we could easily generate API documentation for the Android IDM SDK for developer reference.

## Progress

- [x] Add Gradle Task
- [ ] Optional Todo - Extend task to parent build.gradle to be used by all modules 

## Additional Notes

To run task 
```
$ ./gradlew -q generateAuthJavadoc
```
The docs will be generated and stored in `docs/auth/javadoc`

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -30,14 +30,6 @@ android {
     testOptions {
         unitTests.includeAndroidResources = true
     }
-
-    task generateAuthJavadoc(type: Javadoc, group: "Documentation"){
-        description = 'Generates Auth Javadoc from auth/src/main dir'
-        source = android.sourceSets.main.java.srcDirs
-        classpath += project.files(android.getBootClasspath().join(File.pathSeparator))
-        destinationDir = file("../docs/auth/javadoc/")
-        failOnError false
-    }
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


ciaranRoche:
@wtrocki @aidenkeating @wei-lee @secondsun mind having a look

auth/build.gradle
```diff
@@ -30,6 +30,13 @@ android {
     testOptions {
         unitTests.includeAndroidResources = true
     }
+
+    task generateAuthJavadoc(type: Javadoc){
```
wei-lee: is there a similar task for other modules like core and metrics? I think this should be a top level task that generate javadocs for ALL the modules.

ciaranRoche: I was running into problems getting it to work for all other modules, so I decided to put in the PR as this covers the scope of the ticket, and we could look at making it a top level task later on.

secondsun:
-1.

All you need to do is add "apply from: '../gradle-mvn-push.gradle'" and create a settings.gradle file in the auth directory.  `gradle-mvn-push.gradle` adds the docs and gradlew install will create them.

Sorry I didn't catch it before.

ciaranRoche:
@secondsun good catch, I made those changes but just noticed your comment that these are made in [#39](https://github.com/aerogear/aerogear-android-sdk/pull/39/files).

PR #36, opened 2018-02-14
[WIP] AEROGEAR-1967 - docs for SSO via device browser

rachael-oregan:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-1967

## Description

High level documentation describing SSO via the device browser.  Will need to be updated with links back to the SDK that will give more fine grained detail about the functions mentioned in this PR.


docs/auth/Single Sign-On via Device Browser.md
```diff
@@ -0,0 +1,27 @@
+## Single Sign-On (SSO) via the Device Browser
+The Auth SDK uses OpenID [AppAuth library](https://github.com/openid/AppAuth-Android) and implements OpenID's [auhtorization code flow](http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) to achieve SSO via the device browser.
+
+### Prerequisite's
+1. Ensure your app is using the Android SDK - [Getting Started with AeroGear Services Android SDK](https://github.com/aerogear/aerogear-android-sdk/blob/master/docs/getting-started.adoc) and add in the Auth SDK dependency - [AeroGear Services Auth SDD](https://github.com/aerogear/aerogear-android-sdk/blob/master/docs/auth/README.adoc)
+2. Ensure that keycloak is preconfigured to manage authentication and access control for your service - [Keycloak Client Registration](http://www.keycloak.org/docs/3.2/securing_apps/topics/client-registration.html)
+
+### Steps to achieve SSO via device browser
+Let's assume that you have two apps, **Email App** and **Messaging App**.  Both apps have a service, Service A, that can only be accessed by authenticated users.  The authentication and access control to Service A should already be preconfigured in Keycloak (see prerequisite's).
```
TommyJ1994: I think "Service" is a bit misleading. Realm might be a better choice of word.

TommyJ1994:
@rachael-oregan Yes, we can probably write our own summarized approach for setting the redirect URI (for example) that is specific to how it works for our case/SDK and then we can link out to Keycloak for further information that is more generic. eg [Keycloak Client Configuration](http://www.keycloak.org/docs/latest/server_admin/index.html#oidc-clients)

docs/auth/Single Sign-On via Device Browser.md
```diff
@@ -0,0 +1,27 @@
+## Single Sign-On (SSO) via the Device Browser
+The Auth SDK uses OpenID [AppAuth library](https://github.com/openid/AppAuth-Android) and implements OpenID's [auhtorization code flow](http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) to achieve SSO via the device browser.
+
+### Prerequisite's
+1. Ensure your app is using the Android SDK - [Getting Started with AeroGear Services Android SDK](https://github.com/aerogear/aerogear-android-sdk/blob/master/docs/getting-started.adoc) and add in the Auth SDK dependency - [AeroGear Services Auth SDD](https://github.com/aerogear/aerogear-android-sdk/blob/master/docs/auth/README.adoc)
```
secondsun: It may be easier to just add gradle snippets here instead of referencing a two external documents.  WDYT?

docs/auth/Single Sign-On via Device Browser.md
secondsun: +1 to using Realm.  The apps can talk to as many services as they want as long as all of ther services are authorized using the same KC Realm.

docs/auth/Single Sign-On via Device Browser.md
```diff
@@ -0,0 +1,27 @@
...
+
+3. Bob can now enter his username and password into the Keycloak login page.
+
+4. On clicking login in the Keycloak login page, the Auth SDK is now handling the authentication of Bob for Service A.
+
+5. On successful login, Bob is authenticated on **Email App**  to access Service A.
+
+Let's assume that **Messaging App** also uses the Auth SDK in the same way as the **Email App** described in step 1.
+
+Now let's say Bob tries to access Service A from **Messaging App**.  As Bob has already been authenticated via his device browser for Service A before, he is automatically granted access to Service A within the **Messaging App**.
```
secondsun: Service/Realm here.

Maybe screen shots or some sample code will be helpful too.

docs/auth/Single Sign-On via Device Browser.md
rachael-oregan: I thought of this too but would we end up just copying the same docs everywhere? I'll have 4 prereq's now - import the Android SDK, import the Core SDK, import the Auth SDK and have a keycloak instance.  I can write out the code snippets for them but I think it'll just bloat the prereq section which has already been documented somewhere else

TommyJ1994:
@ciaranRoche Good point, for the mobile security android template, the implementation was based on [RFC 8252](https://tools.ietf.org/html/rfc8252), which has lots of information about this we can reference.

 @rachael-oregan Maybe we could update the intro section to mention this? eg:

The Auth SDK uses OpenID [AppAuth library](https://github.com/openid/AppAuth-Android) and implements OpenID's [auhtorization code flow](http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) to achieve SSO via the device browser. The SDK follows the best practices set out in [RFC 8252 - OAuth 2.0 for Native Apps](https://tools.ietf.org/html/rfc8252).

docs/auth/Single Sign-On via Device Browser.adoc
```diff
@@ -0,0 +1,105 @@
+== Single Sign-On (SSO) via the Device Browser
+
+The Auth SDK uses OpenID https://github.com/openid/AppAuth-Android[AppAuth library] and implements OpenID's http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth[auhtorization code flow] to achieve SSO via the device browser.
```
TommyJ1994: Maybe this can be re-phrased to the following. I think we can leave out the reference to AppAuth,

The Aerogear Auth SDK uses the OpenID Connect http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth[Authorization Code Flow] to achieve SSO via the device browser.

rachael-oregan: Ya sure sounds good to me.  Oh and sorry forgot to respond to @ciaranRoche and your suggestion above too.  I'm thinking maybe it would be better to include the best practices stuff in the main auth doc? What are your thoughts on this?

TommyJ1994: Yeah main docs is best for this since it's not specific to SSO, so +1 from me to have it there instead.

rachael-oregan:
@TommyJ1994 @secondsun @ciaranRoche mind having a look please

TommyJ1994:
@rachael-oregan Really nicely written and thorough. I am wondering if a lot of the prerequisite docs might belong in another doc, instead of the SSO docs? I think for someone wanting to setup SSO, they will likely have a realm & client already setup and would just need to enter the 2 redirect URI's of their app's in keycloak to get it working. Wondering what others think?

docs/auth/Single Sign-On via Device Browser.adoc
```diff
@@ -0,0 +1,112 @@
+== Single Sign-On (SSO) via the Device Browser
```
wei-lee: Not sure if we need to mention `via the Device Browser` in the header. I think this doc is about how to perform SSO using our IDM SDK. Let's make it obvious.

docs/auth/Single Sign-On via Device Browser.adoc
```diff
@@ -0,0 +1,112 @@
...
+                  "ssl-required": "external",
+                  "resource": "emailapp",
+                  "public-client": true,
+                  "use-resource-role-mappings": true,
+                  "confidential-port": 0
+            }
+        }
+    ]
+}
+```
```
wei-lee: Given the context this doc is for, I think we should mention that the file can be generated by the mobile-cli tool. In this case, if developer already has this file available locally, they need to check the details of the keycloak settings to make sure they are matching the configuration in KC.

docs/auth/Single Sign-On via Device Browser.adoc
rachael-oregan: That makes sense, I'll update the heading

docs/auth/Single Sign-On via Device Browser.adoc
rachael-oregan: Right ok, good catch.  I'll include this information

rachael-oregan:
@TommyJ1994 thanks for the feedback.  Very valid point.  In most cases ya I would agree that some of the prerequisites probably aren't needed.

I was looking a this from a point of a view that the goal is to achieve SSO with our SDK and the developer has no idea how to go about it.  It could be an edge case but IMO it's better to have more information than less information. If they already have some of the pre requisites done then they can skip over that info.  So I'd go for keeping the prerequisite's in here.  But let's see what others think

TommyJ1994:
@rachael-oregan I think all the info in the pre-req's is very relevant and wouldn't say that aren't needed, but maybe this should be in the main auth doc, as unless someone wants to use sign in's across multiple apps, they won't see this information. A lot of the information here is relevent to getting Auth working in general, not just SSO.

I think the SSO configuration for keycloak could be a really short 5 sentences or so. At first look at this doc, to an outsider, it looks like SSO is not trivial to implement given the volume of docs. In reality though its very very simple. That's my only concern, but feel free to ignore.

rachael-oregan:
@TommyJ1994 ok ya I see where you're coming from.  Including these in the main auth docs will make this doc a lot shorter you're right.  I'll sync up with  the main auth docs work and try address this. Thanks

rachael-oregan:
@TommyJ1994 @wei-lee I've refined this documentation now.  When you have some time mind having a look please

rachael-oregan:
@TommyJ1994 I'll include that option too, thanks

docs/getting-started/auth-single-sign-on.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+==== Setup Keycloak Messaging Client
+. Create a new client in Keycloak by following this Keycloak guide link:http://www.keycloak.org/docs/latest/server_admin/index.html#oidc-clients[Keycloak OIDC Clients] and ensure the *Keycloak Messaging Client* configuration has the settings specified in step 1 i-vii of <<Setup Keycloak Email Client>>.
+
+'''
+Now that your two Android apps and Keycloak clients are setup, sign into the *Email App* with the user you have previously created. You should be redirected to your device browser which should have loaded the Keyclock login page.  Next, enter in your credentials and login.  You'll be asked to enter a new password for the user. You have now been authenticated on the *Email App* via your device browser.  Finally, try sign into the *Messaging App* with the same user credentials, you should be automatically logged in as that user has been previously authenticated in the *Email App* via the *_same_* device browser.
```
wei-lee: where is the section about setting up the `mobile-service.json` file? I think that is still required though.

rachael-oregan: The setting up of the `mobile-service.json` file is covered in the main auth docs now here -> https://github.com/aerogear/aerogear-android-sdk/pull/34/files#diff-ee51fcbec409a24b3518f32dce8c7097R48.  Do you think these docs should have a note on ensuring the config file has the same config as the keycloak client?

wei-lee: Yeah, I think it's best to mention that the user may need to manually update the `mobile-service.json` file to ensure the Keycloak configuration is updated to match the changes that are made in KC.

rachael-oregan:
@wei-lee @TommyJ1994 I think this is ready to be merged now would you mind having another look.  Note- the relative links should work once this is merged to master

PR #37, opened 2018-02-14
Agdroid 666 3

ziccardi:
## Motivation
Implementing OIDC login 

JIRA: [AGDROID-666](https://issues.jboss.org/browse/AGDROID-666)

**WORK IN PROGRESS**

## Description

Implements the OIDC login

## Progress

- [x] Refactored to use Callbacks
- [x] Make the example app work with login
- [x] Fix tests
- [ ] Add login tests


auth/src/main/java/org/aerogear/mobile/auth/authenticator/AuthenticateOptions.java
```diff
@@ -0,0 +1,7 @@
+package org.aerogear.mobile.auth.authenticator;
+
+/**
+ * Represents options for an authentication action.
+ */
+public interface AuthenticateOptions {
```
TommyJ1994: If this is only for Auth Response handling, maybe we could mention that in the class name? AuthenticateOptions sounds like something used to configure the auth request to login, rather than handle the auth response. Might just cause some confusion.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/OIDCAuthenticatorImpl.java
```diff
@@ -0,0 +1,208 @@
...
+    // Authentication code
+    private void performAuthRequest(final Activity fromActivity, final int resultCode) {
+        AuthorizationServiceConfiguration authServiceConfig = new AuthorizationServiceConfiguration(
+            this.keycloakConfiguration.getAuthenticationEndpoint(),
+            this.keycloakConfiguration.getTokenEndpoint()
+        );
+        this.authState = new AuthState(authServiceConfig);
+
+        AppAuthConfiguration.Builder appAuthConfigurationBuilder = new AppAuthConfiguration.Builder()
+            .setBrowserMatcher(new BrowserBlacklist(
```
TommyJ1994: I think this should be up to the end user to specify.

example/src/main/assets/mobile-services.json
```diff
@@ -0,0 +1,23 @@
...
+    "clientId": "app-cordova",
+    "services": [
+        {
+            "id": "keycloak",
+            "name": "keycloak",
+            "type": "keycloak",
+            "url": "https://keycloak.security.feedhenry.org",
+            "config": {
+                "auth-server-url": "https://keycloak.security.feedhenry.org/auth",
+                "clientId": "client-app",
```
TommyJ1994: This is duplicated with the resource field below. I think we can remove this one as Keycloak uses `resource` in its Keycloak JSON installation format. We will probably need to double check if other code is referring to this and make it refer to `resource` instead of `clientId`.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/OIDCAuthenticatorImpl.java
```diff
@@ -0,0 +1,208 @@
...
+                VersionedBrowserMatcher.CHROME_CUSTOM_TAB));
+
+        if (this.authServiceConfiguration.isAllowSelfSignedCertificate()) {
```
TommyJ1994: Is this check aimed at the tests only? If so, maybe make a note to avoid confusion?

auth/src/main/java/org/aerogear/mobile/auth/configuration/KeycloakConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+import org.aerogear.mobile.core.configuration.ServiceConfiguration;
+
+/**
+ * A class to represent the configuration options of the Keycloak service
+ */
+public class KeycloakConfiguration {
+
+    private static final String SERVER_URL_NAME = "auth-server-url";
+    private static final String REALM_ID_NAME = "realm";
+    private static final String CLIENT_ID_NAME = "clientId";
```
TommyJ1994: I think this should be `CLIENT_ID_NAME = "resource"`.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/AuthenticateOptions.java
wei-lee: It is not just for handing the auth response though. This is basically a wrapper to allow developer to provider their own configuration for certain things. For example, we could extend the implementation class to allow override some AppAuth configurations.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/OIDCAuthenticatorImpl.java
wei-lee: Agreed. But I think we can address it in a different PR.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/OIDCAuthenticatorImpl.java
wei-lee: I will address this in a follow up PR

auth/src/main/java/org/aerogear/mobile/auth/ConnectionBuilderForTesting.java
```diff
@@ -0,0 +1,112 @@
...
+import javax.net.ssl.SSLSession;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+/**
+ * An example implementation of {@link ConnectionBuilder} that permits connecting to http
+ * links, and ignores certificates for https connections. *THIS SHOULD NOT BE USED IN PRODUCTION
+ * CODE*. It is intended to facilitate easier testing of AppAuth against development servers
+ * only.
+ */
+public class ConnectionBuilderForTesting implements ConnectionBuilder {
```
TommyJ1994: Is there anything significantly different that we are doing that is different to the [AppAuth test class](https://github.com/openid/AppAuth-Android/blob/master/app/java/net/openid/appauthdemo/ConnectionBuilderForTesting.java). Do we still need this or is of benefit? It looks like this a test for self-signed certs or HTTP connections only, which AppAuth has features for, but we likely won't be using this as it will be part of our own cert pinning implementation instead and might get some more relevant tests there? WDYT?

TL;DR I think this is a test for a feature we aren't currently or might not use IMO.

auth/src/main/java/org/aerogear/mobile/auth/configuration/KeycloakConfiguration.java
wei-lee: this will be fixed in another PR

auth/src/main/java/org/aerogear/mobile/auth/authenticator/AuthenticateOptions.java
TommyJ1994: Cool, that makes better sense.

example/src/main/assets/mobile-services.json
wei-lee: will remove it in a separate PR.

auth/src/main/java/org/aerogear/mobile/auth/ConnectionBuilderForTesting.java
wei-lee: I agree. We can remove this once we have the certificate pinning implementation in place. For now we do have to support self-signed certs as most developers will run the services with self-signed cert.

wei-lee:
I am going to merge this PR. Going to create another one that will address some of the comments, also get logout work

PR #38, opened 2018-02-14
 ✨ fix for logout and some small improvement

wei-lee:
## Motivation

Implement the logout function in the example app. 

## Description

* Address some comments from PR #37 
* Update UserPrinciple interface to extends the Serializable interface to allow instances to be easily pass around between activities.
* Implement the logout function in the example app
* Update the user details view to show user roles

## Progress

- [x] interface change
- [x] logout function
- [x] user details view update

## Additional Notes

To test, run the example app. You should be able to login/logout and check current logged user details.

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
@@ -1,8 +1,5 @@
 package org.aerogear.mobile.auth.user;
 
-import org.aerogear.mobile.auth.authenticator.AbstractAuthenticator;
-
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
@@ -52,7 +49,7 @@ public class UserPrincipalImpl implements UserPrincipal {
                               final String accessToken) {
         this.username = username;
         this.email = email;
-        this.roles = Collections.synchronizedSet(new HashSet<>(roles));
+        this.roles = new HashSet<>(roles);
         this.identityToken = identityToken;
         this.accessToken = accessToken;
     }
--- a/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
@@ -29,6 +29,16 @@ public class UserPrincipalImpl implements UserPrincipal {
     private final Set<UserRole> roles;
 
     /**
+     * Identity token. It is used for logout
+     */
+    private final String identityToken;
+
+    /**
+     * Access token for http request authorisation
+     */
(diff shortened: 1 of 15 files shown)


wei-lee:
@aidenkeating @ziccardi @rachael-oregan @TommyJ1994 can you please review this.

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -37,10 +47,14 @@
      */
     protected UserPrincipalImpl(final String username,
                               final String email,
-                              final Set<UserRole> roles) {
+                              final Set<UserRole> roles,
+                              final String identityToken,
+                              final String accessToken) {
         this.username = username;
         this.email = email;
         this.roles = Collections.synchronizedSet(new HashSet<>(roles));
```
ziccardi: Why synchronizedSet?

example/src/main/java/org/aerogear/mobile/example/ui/AuthDetailsFragment.java
```diff
@@ -1,26 +1,25 @@
...
-import java.util.List;
-
 import butterknife.BindView;
 import butterknife.OnClick;
 
 public class AuthDetailsFragment extends BaseFragment {
     private static final String TAG = "AuthDetailsFragment";
 
-    public static final String PRINCIPAL = "USER-PRINCIPAL";
+    private UserPrincipalImpl currentUser;
```
ziccardi: "Impl" classes are not meant to be public.
If we want user to be able to use them, we should remove the 'impl' suffix.
Better would be to deal only with the interface.

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
rachael-oregan: I think I added this in cause I thought that a set needed to be synchronised when iterating over that set.  The set is being iterated over for the toString() function here -> https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java#L131

ziccardi: It is needed only if you need to handle concurrent modifications. Otherwise it does not need to be synchronized.
Better to remove that.

wei-lee: Agreed. Will remove it.

example/src/main/java/org/aerogear/mobile/example/ui/AuthDetailsFragment.java
wei-lee: addressed

PR #39, opened 2018-02-14
Agdroid 730

pb82:
## Motivation

Changes for extensible metrics. Allow other SDK modules to add their metrics.

JIRA: https://issues.jboss.org/browse/AGDROID-730

## Description

Allow other SDK modules to add their metrics by implementing `MetricsProvider` and adding it to the MetricsRegistry. Initially i wanted to make it work automagically by using an AnnotationProcessor that collects marked classes and exposes them to the metrics service. But that turned out to not work very well.

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -34,9 +34,14 @@ android {
 
 dependencies {
     implementation project(path: ':core')
+    implementation project(path: ':metrics')
+
     implementation 'net.openid:appauth'
     implementation 'org.bitbucket.b_c:jose4j'
+
     testImplementation "junit:junit"
     testImplementation 'org.mockito:mockito-core'
     testImplementation 'org.robolectric:robolectric'
 }
+
+apply from: '../gradle-mvn-push.gradle'
--- a/auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
@@ -8,12 +8,15 @@ import org.aerogear.mobile.auth.configuration.KeycloakConfiguration;
 import org.aerogear.mobile.auth.authenticator.OIDCAuthenticateOptions;
 import org.aerogear.mobile.auth.credentials.OIDCCredentials;
 import org.aerogear.mobile.auth.authenticator.OIDCAuthenticatorImpl;
+import org.aerogear.mobile.auth.metrics.AuthMetricsProvider;
 import org.aerogear.mobile.auth.user.UserPrincipal;
 import org.aerogear.mobile.auth.utils.UserIdentityParser;
 import org.aerogear.mobile.core.MobileCore;
 import org.aerogear.mobile.core.ServiceModule;
 import org.aerogear.mobile.core.configuration.ServiceConfiguration;
 import org.aerogear.mobile.core.logging.Logger;
+import org.aerogear.mobile.metrics.MetricsRegistry;
+import org.aerogear.mobile.metrics.MetricsService;
 
 import java.security.Principal;
 
@@ -106,6 +109,7 @@ public class AuthService implements ServiceModule {
         this.authStateManager = AuthStateManager.getInstance(context);
         this.authServiceConfiguration = authServiceConfiguration;
         this.oidcAuthenticatorImpl = new OIDCAuthenticatorImpl(this.serviceConfiguration, this.authServiceConfiguration, this.appContext, this.authStateManager);
+        MetricsRegistry.instance().registerProvider(new AuthMetricsProvider());
     }
 
     @Override
--- a/auth/src/main/java/org/aerogear/mobile/auth/metrics/AuthMetricsProvider.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/metrics/AuthMetricsProvider.java
@@ -1,0 +1,21 @@
+package org.aerogear.mobile.auth.metrics;
+
+import android.content.Context;
+
+import org.aerogear.mobile.metrics.interfaces.MetricsProvider;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class AuthMetricsProvider extends MetricsProvider {
+    @Override
+    public String namespace() {
+        return "auth";
+    }
+
+    @Override
+    public JSONObject metrics(Context context) throws JSONException {
+        JSONObject metrics = new JSONObject();
+        // TODO: add metrics
+        return metrics;
+    }
+}
--- a/core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
@@ -21,6 +21,8 @@ public interface HttpResponse {
     HttpResponse onComplete(Runnable runnable);
 
     int getStatus();
+    boolean isFailed();
+    Exception getRequestError();
 
     /**
      * This is a terminal method that will block the thread called on until the http request has been
--- a/metrics/src/main/java/org/aerogear/mobile/metrics/MetricsRegistry.java
+++ b/metrics/src/main/java/org/aerogear/mobile/metrics/MetricsRegistry.java
@@ -1,0 +1,35 @@
+package org.aerogear.mobile.metrics;
+
+import org.aerogear.mobile.metrics.interfaces.MetricsProvider;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * The metrics registry is the central storage for all metrics providers
+ */
+public final class MetricsRegistry {
+    private static MetricsRegistry INSTANCE;
+
+    private final Set<MetricsProvider> providers = new HashSet<>();
+
+    private MetricsRegistry() {
+        INSTANCE = this;
+    }
+
+    public void registerProvider(MetricsProvider provider) {
+        providers.add(provider);
+    }
+
+    public Set<MetricsProvider> getProviders() {
+        return this.providers;
+    }
+
+    public static MetricsRegistry instance() {
+        if (INSTANCE == null) {
+            new MetricsRegistry();
+        }
+
+        return INSTANCE;
+    }
+}
--- a/metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
+++ b/metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
@@ -1,7 +1,6 @@
 package org.aerogear.mobile.metrics;
 
 import android.content.Context;
-import android.content.SharedPreferences;
 
 import org.aerogear.mobile.core.MobileCore;
 import org.aerogear.mobile.core.ServiceModule;
@@ -10,67 +9,55 @@ import org.aerogear.mobile.core.http.HttpRequest;
 import org.aerogear.mobile.core.http.HttpResponse;
 import org.aerogear.mobile.core.http.HttpServiceModule;
 import org.aerogear.mobile.core.logging.Logger;
+import org.aerogear.mobile.metrics.interfaces.MetricsProvider;
+import org.aerogear.mobile.metrics.providers.DefaultMetricsProvider;
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import java.util.UUID;
-
 public class MetricsService implements ServiceModule {
-    public final static String STORAGE_NAME = "org.aerogear.mobile.metrics";
-    public final static String STORAGE_KEY = "metrics-sdk-installation-id";
-
     private final static String MODULE_NAME = "metrics";
-    private final static String TAG = "AEROGEAR/METRICS";
+    public final static String TAG = "AEROGEAR/METRICS";
 
     private HttpServiceModule httpService;
     private Logger logger;
 
-    private String appVersion = null;
     private String metricsUrl = null;
 
-    /**
-     * Get or create the client ID that identifies a device as long as the user doesn't
-     * reinstall the app or delete the app storage. A random UUID is created and stored in the
-     * application shared preferences.
-     *
-     * Can be overridden to provide a different implementation for identification.
-     *
-     * @param context Android app context
-     * @return String Client ID
-     */
-    protected String getOrCreateClientId(final Context context) {
-        final SharedPreferences preferences = context
-            .getSharedPreferences(STORAGE_NAME, Context.MODE_PRIVATE);
-
-        String clientId = preferences.getString(STORAGE_KEY, null);
-        if (clientId == null) {
-            clientId = UUID.randomUUID().toString();
-
-            logger.info(TAG, "Generated a new client ID: " + clientId);
-
-            SharedPreferences.Editor editor = preferences.edit();
-            editor.putString(STORAGE_KEY, clientId);
-            editor.commit();
+    // Copy all properties from the source object and add it to the top level
+    // of the root object
+    private void addToRoot(final JSONObject root, final JSONObject source) throws JSONException {
+        JSONArray names = source.names();
+        for (int i = 0; i < names.length(); ++i) {
+            String name = names.getString(i);
+            root.put(name, source.get(name));
         }
-
-        return clientId;
     }
 
-    /**
-     * This method is called to create the JSON object containing the metrics data.
-     * Can be overridden to add more data points.
-     *
-     * @param context Android app context
-     * @return JSONObject Metrics data
-     * @throws JSONException when any of the data results in invalid JSON
-     */
-    protected JSONObject metricsData(final Context context) throws JSONException {
-        final JSONObject result = new JSONObject();
-        result.put("clientId", getOrCreateClientId(context));
-        result.put("appId", context.getPackageName());
-        result.put("appVersion", appVersion);
-        result.put("sdkVersion", MobileCore.getSdkVersion());
-        return result;
+    private void aggregateSingleProvider(
+        final MetricsProvider provider,
+        final JSONObject target,
+        final Context context)
+        throws JSONException {
+
+        String namespace = provider.namespace();
+        JSONObject metrics = provider.metrics(context);
+
+        // If the namespace of a provider is null we add it's properties
+        // to the root object
+        if (namespace == null || namespace.length() == 0) {
+            addToRoot(target, metrics);
+        } else {
+            target.put(namespace, metrics);
+        }
+    }
+
+    // Iterate over all providers and add their JSON to the target metrics object
+    private void aggregateAllProviders(final JSONObject target, final Context context)
+        throws JSONException {
+        for (final MetricsProvider provider: MetricsRegistry.instance().getProviders()) {
+            aggregateSingleProvider(provider, target, context);
+        }
     }
 
     /**
@@ -83,19 +70,17 @@ public class MetricsService implements ServiceModule {
      */
     public final void init(final Context context) {
         try {
-            final JSONObject data = metricsData(context);
+            JSONObject metrics = new JSONObject();
+            aggregateAllProviders(metrics, context);
 
             // Send request to backend
             HttpRequest request = httpService.newRequest();
-            request.post(metricsUrl, data.toString().getBytes());
+            request.post(metricsUrl, metrics.toString().getBytes());
             HttpResponse response = request.execute();
 
             // Async response handling
-            response.onComplete(() -> {
-                if (response.getStatus() != 200) {
-                    logger.error(TAG, "Error sending metrics data");
-                }
-            });
+            response.onComplete(new MetricsResponseHandler(response));
+            response.waitForCompletionAndClose();
         } catch (JSONException e) {
             logger.error(TAG, e);
         }
@@ -104,9 +89,10 @@ public class MetricsService implements ServiceModule {
     @Override
     public void configure(final MobileCore core, final ServiceConfiguration serviceConfiguration) {
         metricsUrl = serviceConfiguration.getUri();
-        appVersion = core.getAppVersion();
         httpService = core.getHttpLayer();
         logger = MobileCore.getLogger();
+
+        MetricsRegistry.instance().registerProvider(new DefaultMetricsProvider());
     }
 
     @Override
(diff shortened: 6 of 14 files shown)


auth/src/main/java/org/aerogear/mobile/auth/metrics/AuthMetricsProvider.java
```diff
@@ -0,0 +1,21 @@
...
+import org.json.JSONObject;
+
+public class AuthMetricsProvider extends MetricsProvider {
+    @Override
+    public String namespace() {
+        return "auth";
+    }
+
+    @Override
+    public JSONObject metrics(Context context) throws JSONException {
```
david-martin: @pb82 is JSONObject the best choice here? It leaves it very open to the structure of the object returned. Is there some structure that should be enforced?

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
```diff
@@ -0,0 +1,40 @@
...
+import org.json.JSONObject;
+
+import java.util.Objects;
+
+public abstract class MetricsProvider {
+    public String namespace() {
+        return null;
+    }
+
+    public JSONObject metrics(final Context context) throws JSONException {
```
david-martin: does it make sense to have default implementations for these methods? Would it make sense for them to be interfaces instead and force the implementer to implement.

auth/src/main/java/org/aerogear/mobile/auth/metrics/AuthMetricsProvider.java
wtrocki: > Is there some structure that should be enforced?
That's just metrics data. JSONObject/HashMap will will work.
Structure will be enforced on metrics payload:
```
{
 "metricsName": { "JSONObject key values will be here" }
}
```

secondsun:
@pb82 Is there a Android Project that uses these features I can reference?

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsRegistry.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * The metrics registry is the central storage for all metrics providers
+ */
+public final class MetricsRegistry {
+    private static MetricsRegistry INSTANCE;
+
+    private final Set<MetricsProvider> providers = new HashSet<>();
```
wtrocki: [Optional] Set does not guarantee order. Not sure if order here will matter.

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
```diff
@@ -10,67 +9,58 @@
...
+        final JSONObject target,
+        final Context context)
+        throws JSONException {
+
+        String namespace = provider.namespace();
+        JSONObject metrics = provider.metrics(context);
+
+        // If the namespace of a provider is null we add it's properties
+        // to the root object
+        if (namespace == null || namespace.length() == 0) {
```
wtrocki: Not sure if we should allow this case and threat that as error or do it this way.
I have no preference for this but flagging that as possible problem.

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
wtrocki: +1 for making this abstract (with javadoc documentation)

metrics/src/main/java/org/aerogear/mobile/metrics/providers/DefaultMetricsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.aerogear.mobile.core.MobileCore;
+import org.aerogear.mobile.metrics.MetricsService;
+import org.aerogear.mobile.metrics.interfaces.MetricsProvider;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.UUID;
+
+public class DefaultMetricsProvider extends MetricsProvider {
```
wtrocki: Missing documentation.

metrics/src/main/java/org/aerogear/mobile/metrics/providers/DefaultMetricsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public JSONObject metrics(final Context context) throws JSONException {
+        final JSONObject result = super.metrics(context);
+        result.put("clientId", getOrCreateClientId(context));
+        result.put("appId", context.getPackageName());
+        result.put("appVersion", getAppVersion(context));
+        result.put("sdkVersion", MobileCore.getSdkVersion());
+        result.put("platform", "android");
```
wtrocki: [Optional] We can get also device name (mentioned in proposal)

auth/build.gradle
```diff
@@ -34,9 +34,14 @@ android {
 
 dependencies {
     implementation project(path: ':core')
+    implementation project(path: ':metrics')
```
wei-lee: this means the auth module will become dependent on the metrics module. I thought we want to avoid direct dependencies between each service modules?

wtrocki:
@secondsun Good point. We should add code to ./example project to showcase this.

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
```diff
@@ -106,6 +109,7 @@ public void init(final Context context, final AuthServiceConfiguration authServi
         this.authStateManager = AuthStateManager.getInstance(context);
         this.authServiceConfiguration = authServiceConfiguration;
         this.oidcAuthenticatorImpl = new OIDCAuthenticatorImpl(this.serviceConfiguration, this.authServiceConfiguration, this.appContext, this.authStateManager);
+        MetricsRegistry.instance().registerProvider(new AuthMetricsProvider());
```
wei-lee: I am not sure if it's just me, but `registerProvider` makes me think it is to register a provider for metrics service.  But it is really a provider to provide metrics data. So will it better to make it more clear?

auth/src/main/java/org/aerogear/mobile/auth/metrics/AuthMetricsProvider.java
```diff
@@ -0,0 +1,21 @@
+package org.aerogear.mobile.auth.metrics;
+
+import android.content.Context;
+
+import org.aerogear.mobile.metrics.interfaces.MetricsProvider;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class AuthMetricsProvider extends MetricsProvider {
```
wei-lee: same as my comment above. I think it is better to rename it `MetricsDataProvider` to make it clear that they are only to provide data for metrics, not the metrics service itself.

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
wtrocki: On IOS we call this `MetricsCollector`

auth/src/main/java/org/aerogear/mobile/auth/metrics/AuthMetricsProvider.java
pb82: good point. If we want to enforce a flat structure we could return `Map<String, String>` or `Map<String, ClassEnclosingAllowedTypes>`

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
pb82: The idea was that the default implementation provides everything that is needed to hook into the metrics service. The `namespace` for example is optional, i would prefer not to enforce implementation in an interface (if you don't override it your properties will be added to the root of the metrics structure). Also i'd like to have default implementations for equals and hashCode that take the namespace into account.

auth/src/main/java/org/aerogear/mobile/auth/metrics/AuthMetricsProvider.java
wtrocki: @pb82 Actually I want to revert my comment. JSONObject will be better as it will allow nested data.

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
wtrocki: @pb82 Yeah. We have abstract class with no abstract method. We could just make this method abstract and that will be all.

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsRegistry.java
pb82: don't think the order matters here, but i'm open for other opinions

metrics/src/main/java/org/aerogear/mobile/metrics/providers/DefaultMetricsProvider.java
pb82: thanks, will add

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
pb82: I added this mostly to support the current format (without nesting). But i can remove it if we decide to enforce namespacing.

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
pb82: Good point, will also rename it to `MetricsCollector`. Provider in the sense: it provider metrics FROM this module TO the metrics service.

auth/build.gradle
pb82: Yes, i'm also not 100% happy with this fact. Moving the interface to core might be an option because all modules depend on core anyway. But then core would have to keep track of the registered metric providers.

pb82:
@wei-lee That's right, at the moment we only support startup metrics. There is currently no proposal for 'custom' metrics (letting the module decide what and when to send). Currently we visualize things like SDK and App versions that are in use. Although at some point the metrics will have to be more flexible. You are right that this might not even be useful for other modules: not every module would have startup information to report. @david-martin what's your opinion on that? Will we stay with startup metrics for now or do we want to add support for custom metrics as soon as possible?

I agree that moving metrics into core would make sense. It would remove the need for additional dependencies in modules.

wtrocki:
Worth reading discussion in https://issues.jboss.org/browse/AGDROID-730

david-martin:
> So as the consumer of the metric service, I should be able to decide what data should be collected and when they should be collected.

@wei-lee Would you see the trello card https://trello.com/c/tnoYEHyE/26-8-as-an-android-mobile-developer-i-want-to-capture-metrics-on-device-self-defence-checks-so-that-i-know-when-a-check-has-failed being about 'enabling' the developer to capture metrics about self defence checks (by writing their own code)
*or*
automatically (if configured) sending metrics about self defence checks if the developer has both the auth & metrics SDK in their app?

wei-lee:
@david-martin My understanding is the latter, which means when we implement the self defence checks, we will invoke the metrics module and pass on the results automatically.  
However, it unlikely will happen at the app startup time. 
It will also make things a lot easier for us if we don't need to check if the backend metrics service is enabled or not.

pb82:
@wei-lee You don't have to check if the backend service is enabled or not. With the current implementation you only need to implement and register the metrics provider. If the metrics is not used (init not called) then nothing will be sent and the provider will be ignored.

wei-lee:
@pb82 yes, that's right. I think the main problem then is about only collect metrics at startup time. For the self defence checks, I think we will provide a method called `runSelfDefenceChecks` as part of either the IDM or security module, and developers can decide when they want to invoke them. We only need to send out the metrics data when that method is actually invoked.

pb82:
@wei-lee Yes that would be a problem with the current implementation: the user would have to call `metrics.init()` only after they ran the self defence checks. Thats not a good UX and we should not impose such a decision (the order of when to run init methods) on the user.

One possible solution would be to separate startup metrics from runtime metrics (in the backend as well) where startup metrics are unsurprisingly only sent on startup and runtime metrics are either sent whenever new data is collected or after some batch size is reached.

wtrocki:
@wei-lee I really liked your suggestion, but we got specific recommendation on top level card to have metrics as separate module. This caused some problems that were mentioned in the ticket etc. 

I believe that we just implemented pull metrics - so collectors are called at by metrics when it's needed (for example on app init). I believe that for IDM you will prefer to have push aproach where individual SDK's pushing metrics and there is this fixed time when metrics are sent to server. 
That's kinda standard way how metrics are collected, but from my point of view, we were asked to do something else. If we change requirements to have it core (which will be good IMHO) we could implement that the way it was suggested.

wei-lee:
@pb82 > One possible solution would be to separate startup metrics from runtime metrics (in the backend as well) where startup metrics are unsurprisingly only sent on startup and runtime metrics are either sent whenever new data is collected or after some batch size is reached.

Not sure what's the benefit of separating them TBH. 

Either way I think using the "Pulling" model is not a good idea. I think we far better go with the "Push" model. Both of them could work but I think the "Push" model is way simpler from an implementation point of view.

pb82:
@wei-lee @wtrocki Right, a pure push model might actually work better all scenarios considered. The metrics service could still send the startup metrics as soon as its initialized and other modules could send metrics whenever they are ready (and not depend on when metrics.init() is called).

wei-lee:
@wtrocki 

> I really liked your suggestion, but we got specific recommendation on top level card to have metrics as separate module. This caused some problems that were mentioned in the ticket etc.

I have left a comment in the trello card to explain why it is not a good approach. I think if there are more people echo my comment, we should be able to do something about it.

But even if that is still a requirement, we should still provide a `push` model.

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsRegistry.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public void registerProvider(MetricsProvider provider) {
+        providers.add(provider);
+    }
+
+    public Set<MetricsProvider> getProviders() {
+        return this.providers;
+    }
+
+    public static MetricsRegistry instance() {
```
secondsun: Can we get dependency injection over singletons?

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
```diff
@@ -83,19 +70,17 @@ protected JSONObject metricsData(final Context context) throws JSONException {
...
             HttpResponse response = request.execute();
 
             // Async response handling
-            response.onComplete(() -> {
-                if (response.getStatus() != 200) {
-                    logger.error(TAG, "Error sending metrics data");
-                }
-            });
+            response.onComplete(new MetricsResponseHandler(response));
+            response.waitForCompletionAndClose();
```
secondsun: This will block the main thread which means that init should be called from a background thread.  Is that the assumption?

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null)
+            return false;
+        if (getClass() != o.getClass())
+            return false;
+        MetricsProvider person = (MetricsProvider) o;
```
secondsun: person?

metrics/src/main/java/org/aerogear/mobile/metrics/providers/DefaultMetricsProvider.java
```diff
@@ -0,0 +1,82 @@
...
+                .versionName;
+        } catch (PackageManager.NameNotFoundException e) {
+            // Wrap in Initialization exception
+            throw new JSONException(e.getMessage());
+        }
+    }
+
+    @Override
+    public String namespace() {
+        return null;
```
secondsun: If namespace is null in the abstract and concrete classes, why do we have it?

metrics/src/test/java/org/aerogear/mobile/metrics/providers/DefaultMetricsProviderTest.java
```diff
@@ -0,0 +1,7 @@
+package org.aerogear.mobile.metrics.providers;
+
+import static org.junit.Assert.*;
+
+public class DefaultMetricsProviderTest {
+
```
secondsun: Empty class.

core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
```diff
@@ -21,6 +21,8 @@
     HttpResponse onComplete(Runnable runnable);
 
     int getStatus();
+    boolean isFailed();
+    Exception getRequestError();
```
secondsun: Request errors should be sent asynchronously and processed then.  I don't think it is correct to have them also be properties of the response.

auth/build.gradle
secondsun: So it might be worth it to have a utility library auth-metrics which provides auth metrics for auth.  
What metrics from the client side for auth are we wanting to capture?

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
pb82: No, better to remove that line and just let it do it's thing in the background.

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
pb82: Copy Pasta 🍝

metrics/src/main/java/org/aerogear/mobile/metrics/providers/DefaultMetricsProvider.java
pb82: In that case `null` has a specific meaning: the properties will be added to the root of the metrics object. You need to override it and return a value if you want namespacing.

core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
pb82: I think i added this because there was no way to know when a request has failed and trying to use `getStatus` on a failed requests results in a NPE. Maybe i was using the wrapper wrong.

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsRegistry.java
pb82: I'd prefer that too. But in that case we would need dependency injection across multiple modules. Will have a look at dagger2, maybe it can help us here.

metrics/src/main/java/org/aerogear/mobile/metrics/providers/DefaultMetricsProvider.java
pb82: Ah sorry, minunderstood the comment. Yes this can be removed.

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
```diff
@@ -0,0 +1,40 @@
...
+
+import android.content.Context;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.Objects;
+
+public abstract class MetricsProvider {
+    public String namespace() {
```
josemigallas: From what I've understand this method should be abstract. And I suggest it returns an enum `MetricsProviderNamespace` that references all existing namespaces instead of a hardcode string.

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null)
+            return false;
+        if (getClass() != o.getClass())
+            return false;
+        MetricsProvider person = (MetricsProvider) o;
+        return Objects.equals(namespace(), person.namespace());
```
josemigallas: The condition of equality is really that namespace is the same, isn't it?

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
josemigallas: For me, having a null or empty namespace would look like an error. When would this condition be met?

metrics/src/main/java/org/aerogear/mobile/metrics/interfaces/MetricsProvider.java
pb82: Yes, that and a few sanity checks (same Object type, etc.)

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
pb82: If you don't want to namespace your data and just append it to the root of the object

metrics/src/main/java/org/aerogear/mobile/metrics/providers/DefaultMetricsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public JSONObject metrics(final Context context) throws JSONException {
+        final JSONObject result = super.metrics(context);
```
josemigallas: I strongly suggest to make this data more robust. If DefaultMetrics values are fixed I would wrap them in a class or defining them as properties or something to avoid hardcoding. I definitely put those strings into constants too.

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsService.java
josemigallas: IMO all providers should have a namespace. Even the default ones could have it, for consistency and to eliminate this null/empty namespace approach.

metrics/src/main/java/org/aerogear/mobile/metrics/MetricsRegistry.java
wtrocki: We been looking at dagger2 - investigation was closed as it was not possible to use it on library level.

PR #41, opened 2018-02-14
Added sanity check

ziccardi:
## Motivation

JIRA: https://issues.jboss.org/browse/AGDROID-753

## Description

Added a SanityCheck object.
Added sanity check calls

## Progress

- [x] Add non-null checks
- [x] Add non-empty checks
- [x] Add unit tests
- [x] Move sanity check class into the core module
- [x] Fix failing existing tests
- [x] Add non-null annotations

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
@@ -2,6 +2,7 @@ package org.aerogear.mobile.auth;
 
 import android.content.Context;
 import android.content.Intent;
+import android.support.annotation.NonNull;
 
 import org.aerogear.mobile.auth.configuration.AuthServiceConfiguration;
 import org.aerogear.mobile.auth.configuration.KeycloakConfiguration;
@@ -15,7 +16,12 @@ import org.aerogear.mobile.core.ServiceModule;
 import org.aerogear.mobile.core.configuration.ServiceConfiguration;
 import org.aerogear.mobile.core.logging.Logger;
 
-import java.security.Principal;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
 
 /**
  * Entry point for authenticating users.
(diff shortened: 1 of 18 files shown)


auth/src/main/java/org/aerogear/mobile/auth/utils/SanityCheck.java
```diff
@@ -0,0 +1,17 @@
+package org.aerogear.mobile.auth.utils;
+
+public class SanityCheck {
```
wei-lee: actually, this is useful for all the other modules so I think we should move this class to the core module.

ziccardi: Agree. Will move it.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/OIDCAuthenticatorImpl.java
```diff
@@ -78,12 +80,14 @@ public OIDCAuthenticatorImpl(final ServiceConfiguration serviceConfiguration, fi
     @Override
     public void authenticate(final AuthenticateOptions authOptions, final Callback<UserPrincipal> callback) {
         this.authCallback = callback;
```
wei-lee: check the `callback` as well

ziccardi: Good catch!

ziccardi:
@aidenkeating I think it would be good. Going to add them.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/OIDCAuthenticatorImpl.java
ziccardi: Addressed

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
```diff
@@ -102,9 +102,9 @@ public void configure(final MobileCore core, final ServiceConfiguration serviceC
      * @param context
      */
     public void init(final Context context, final AuthServiceConfiguration authServiceConfiguration) {
```
wei-lee: so both `configure` and `init` methods need to be called to initialise the `AuthService` instance properly. I think we should add checks inside the public methods (`currentUser`, `login`, `logout`) to verify that the instance has been initialised properly. Otherwise it should throw an exception and explain that the instance hasn't been inited properly.

ziccardi:
@wei-lee would you mind to give another look?

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
```diff
@@ -32,6 +37,52 @@
...
+                methodsToBeInvoked.add(STEP.INITIALIZED.methodName);
+            }
+
+            if (!methodsToBeInvoked.isEmpty()) {
+                throw new IllegalStateException(
+                    String.format("The AuthService has not been correctly initialised. Following methods needs to be called: %s",
+                        Arrays.toString(methodsToBeInvoked.toArray())));
+            }
+        }
+    }
```
wei-lee: can we add some simple tests for this?

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
```diff
@@ -32,6 +37,52 @@
 
     private Context appContext;
     private Logger logger;
+
+    private final ReadynessStatus status = new ReadynessStatus();
+
+    private static class ReadynessStatus {
```
wei-lee: some javadocs will help others to read & understand the code.

ziccardi:
@wei-lee would you mind to give another look?

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
ziccardi: Addressed

secondsun:
@ziccardi I was actually in the middle of a review when you merged.  Please slow down next time

ziccardi:
@secondsun Sorry, I already had 2 LGTM, so I merged it

wei-lee:
@secondsun When this happens, you can continue your review and give your feedback. The author should still look at those comments and maybe create another PR to address your comments.

PR #49, opened 2018-02-16
Push based metrics

pb82:
## Motivation

Moving metrics into core and change to push based approach. Example usage in app:

```java
    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);
        // This will automatically initialize the metrics module and send the default metrics
        this.core = MobileCore.init(getApplicationContext());        

        // Can be used anywhere, anytime
        MetricsService metrics = core.getInstance(MetricsService.class);
        publisher.publish(new MyMetrics());
    }
```

JIRA: https://issues.jboss.org/browse/AGDROID-730

## Description

A pull based approach (where metrics asks the modules to provide their metrics at app startup) is not sufficient for our requirements, so i'm chaning it to a model where SDK modules or apps can send metrics whereever and whenever they want.

The metrics module is also in core now. When core is initialized the default metrics will be sent automatically.

Currently every ` push` call will result in a request. We could implement queing and pooling of requests in a future PR as well as backpressure handling if that becomes necessary.

@wei-lee Tests added through @danielpassos PR

ping @danielpassos @wtrocki updated description. I am not sure if we should add code to the example project yet, because:

1) Default metrics are sent automatically, no code to add (and we don't have a proposal for user defined metrics yet on the backend)
2) We don't exactly know what the backend URL looks like. I could put a dummy URL in the config but then you would get a failed request every time you try the example.

What do you think?


secondsun:
does this obsolete https://github.com/aerogear/aerogear-android-sdk/pull/39?

pb82:
@secondsun yes, closed #39 now

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -79,6 +81,18 @@ private MobileCore(Context context, Options options)
...
             this.httpLayer = options.httpServiceModule;
         }
+
+        // -- init and send default metrics
+        initMetrics();
+    }
+
+    private void initMetrics() {
+        MetricsService metrics = getInstance(MetricsService.class);
+        metrics.sendDefaultMetrics();
```
wtrocki: [Minor] In proposal this is linked as `app` group 
https://github.com/aerogear/proposals/blob/master/metrics/default-mobile-app-metrics.md

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -104,4 +104,16 @@ public String stringBody() {
             return "";
         }
     }
+
+    @Override
+    public boolean requestFailed() {
```
wtrocki: @pb82 Ignore this comment - I just want to flag that this http wrapper is not properly implemented.
@secondsun @danielpassos  Flagging this as it's kinda poor api. Any reason why we did not used lambdas here?

pb82: @wtrocki I added `requestFailed` and `getRequestError` because i couldn't see any other way know when a request failed. Trying to get the status of a failed request will result in a NPE. Happy to change this if there is a better solution.

core/src/main/java/org/aerogear/mobile/core/metrics/DefaultMetrics.java
```diff
@@ -0,0 +1,101 @@
...
+        this.appId = context.getPackageName();
+        this.appVersion = getAppVersion(context);
+        this.sdkVersion = MobileCore.getSdkVersion();
+        this.platform = "android";
+        this.platformVersion = String.valueOf(Build.VERSION.SDK_INT);
+    }
+
+    /**
+     * Return default metrics in JSON format
+     * @return JSONObject with metrics data
```
wtrocki: Outdated

core/src/main/java/org/aerogear/mobile/core/metrics/DefaultMetrics.java
```diff
@@ -0,0 +1,101 @@
...
+     * @throws JSONException
```
wtrocki: Outdated comment

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
wtrocki: Peter that's not problem. I just wanted to mention that http api did not provide that and needs to be redesigned. This fix works for now and it's not the problem with your PR.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -0,0 +1,59 @@
...
+     * Send default metrics including:
+     * - Client ID
+     * - App ID
+     * - App version
+     * - SDK version
+     * - Platform (always "android")
+     * - Platform version
+     */
+    public void sendDefaultMetrics() {
+        getPublisherForNamespace("default")
```
wtrocki: Proposal use `app`

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -0,0 +1,59 @@
+package org.aerogear.mobile.core.metrics;
+
+import org.aerogear.mobile.core.MobileCore;
+import org.aerogear.mobile.core.ServiceModule;
+import org.aerogear.mobile.core.configuration.ServiceConfiguration;
+import org.aerogear.mobile.core.metrics.interfaces.MetricsPublisher;
+import org.aerogear.mobile.core.metrics.interfaces.Observer;
+
+public class MetricsService implements ServiceModule {
```
wtrocki: [Optional] If something is in the core it may not need that interface. Not sure if that should be actual service if that's part of the core.

pb82:
@wtrocki I will add an example

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
pb82: @wtrocki I guess the nice thing with using that interface is that this way it behaves just like any other SDK module and `core.getInstance` can be used to retrieve the module instead us cluttering `MobileCore` with another `getXXXService` method. Sure, i added a convenience method to core to retrieve a metrics publisher but we could just as well remove it.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -79,6 +83,17 @@ private MobileCore(Context context, Options options)
...
         }
+
+        sendDefaultMetrics();
+    }
+
+    private void sendDefaultMetrics() {
+        try {
+            MetricsService metrics = getInstance(MetricsService.class);
+            metrics.sendDefaultMetrics();
+        } catch (ConfigurationNotFoundException e) {
```
pb82: @danielpassos I don't quite like this, but at the moment there is no way to check the configuration for a service exists without having an instance of this service (because we use the `type` method of the instance to get the name of the configuration). What do you think?

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private final Logger logger;
+
+    public LoggerMetricsPublisher(Logger logger) {
+        this.logger = logger;
+    }
+
+    @Override
+    public void publish(Metrics... metrics) {
+        logger.info("Metrics collected");
```
wei-lee: not sure if we should use `info` level here. The message is not very clear as it could be the metrics is being collected on the server side.

Also given that the real metrics data are only logged at `debug` level, which means they will not be printed for a release app,  not sure if there is any value to print the message here?

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -0,0 +1,55 @@
...
+
+            httpRequest.post(url, json.toString().getBytes());
+
+            MobileCore.getLogger().debug("Sending metrics");
+            HttpResponse httpResponse = httpRequest.execute();
+            httpResponse.onComplete(() -> {
+                if (httpResponse.getStatus() == HTTP_OK) {
+                    MobileCore.getLogger().debug("Metrics sent");
+                } else {
+                    MobileCore.getLogger().error(httpResponse.getRequestError().getMessage(),
```
wojta: There is really no way how consumer of the API can know if some error happened. Shouldn't be there some way, like optionally adding callback on error.
```
public void publish(Metrics... metrics, PushFailedCallback errorCallback) {
```

pb82: @wojta at the moment the metrics are automatically sent by the core sdk, but when we add custom user metrics we should provide such a callback.

wojta: I can live with that now :)

danielpassos:
Hi 5

PR #51, opened 2018-02-16
 ✨ initial implementation to load and manage json web key set for token [WIP]

wei-lee:
verification

## Motivation

JIRA: https://issues.jboss.org/browse/AGDROID-751

## Description

This PR will create a new JwksManager class to fetch and save the json web key set from the Keycloak server. 

The key set will be used to verify the local jwt token to make sure it is valid

## Progress

- [x] implement JwksManager
- [ ] use it to verify token
- [ ] unit tests
- [ ] integration tests using the sample app


auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -0,0 +1,155 @@
...
+import org.aerogear.mobile.core.logging.Logger;
+import org.jose4j.jwk.JsonWebKeySet;
+import org.jose4j.lang.JoseException;
+
+import java.util.Date;
+
+/**
+ * A class that is responsible for manage the Json Web Key Set(JWKS).
+ */
+class JwksManager {
```
secondsun: public?  If it is supposed to be package protected maybe @PublicForTesting?

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -0,0 +1,155 @@
...
+            HttpRequest getRequest = httpModule.newRequest();
+            getRequest.get(jwksUrl);
+            HttpResponse response = getRequest.execute();
+            response.onComplete(() -> {
+                //this is invoked on a background thread.
+                if (response.getStatus() == 200) {
+                    String jwksContent = response.stringBody();
+                    boolean isValidKeyset = true;
+                    try {
+                        new JsonWebKeySet(jwksContent);
```
secondsun: Missing assignment?

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
wei-lee: yes, it should be public

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
wei-lee: the original plan was just trying the validate the content and that's why it didn't need assignment. But now I have changed the API slightly and it will have the assignment statement.

auth/src/main/java/org/aerogear/mobile/auth/configuration/AuthServiceConfiguration.java
```diff
@@ -57,6 +65,11 @@ public AuthConfigurationBuilder allowSelfSignedCertificate(boolean allowSelfSign
             return this;
         }
 
+        public AuthConfigurationBuilder withMinTimeBetweenJwksRequests(int minTimeBetweenJwksRequests) {
+            this.minTimeBetweenJwksRequests = minTimeBetweenJwksRequests;
+            return this;
+        }
+
         public AuthServiceConfiguration build() {
             return new AuthServiceConfiguration(this);
```
ziccardi: If the `withMinTimeBetweenJwksRequests` is not called, minTimeBetweenJwksRequests will be 0 and the AuthServiceConfiguration will be built with a '0 min between jwks requests'.
I would suggest to make the default here in the builder, and make the minTimeBetweenJwksRequests final in the  AuthServiceConfiguration class

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -0,0 +1,155 @@
...
+class JwksManager {
+
+    private static final int MILLISECONDS_PER_MINUTE = 60*1000;
+
+    private static final String STORE_NAME = "org.aerogear.mobile.auth.JwksStore";
+
+    private static final String ENTRY_SUFFIX_FOR_KEY_CONTENT = "jwks_content";
+    private static final String ENTRY_SUFFIX_FOR_REQUEST_DATE = "requested_date";
+
+    private HttpServiceModule httpModule;
```
ziccardi: these should be final

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -0,0 +1,155 @@
...
+    private AuthServiceConfiguration authServiceConfiguration;
+    private SharedPreferences sharedPrefs;
+    private Logger logger;
+
+    JwksManager(Context context, HttpServiceModule httpModule, AuthServiceConfiguration authServiceConfiguration, Logger logger) {
```
ziccardi: It is a best practice to make the parameters final

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -0,0 +1,155 @@
...
+     * Check when the JWKS was requested last time and determine if a request should be sent again.
+     * @param keyCloakConfig the configuration of the Keycloak server
+     * @return true if the request should be triggered
+     */
+    private boolean shouldRequestJwks(KeycloakConfiguration keyCloakConfig) {
+        boolean shouldRequest = true;
+        String namespace = keyCloakConfig.getRealmName();
+        String requestedDateEntryName = buildEntryNameForQuestedDate(namespace);
+        long lastRequestDate = this.sharedPrefs.getLong(requestedDateEntryName, 0);
+        long currentTime = new Date().getTime();
```
ziccardi: Use System.currentTimeMillis() to avoid instantiating a Date object

wei-lee:
replaced by #60

PR #53, opened 2018-02-19
AEROGEAR-2177 - add asciibinder structure to sdk docs

finp:
## Motivation

want to set up the structure of docs in sdk and service repos, so that they build their own microsite (for all releases) using asciibinder

JIRA: https://issues.jboss.org/browse/AEROGEAR-2177

## Description

* add asciibinder files
* moved existing files (only one edit, make link to svg work)

## Progress
done

## Additional Notes

* feel free to reject, but want to show a proposal re. layout of files asap
* will edit adocs files in a follow up task


docs/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM centos/ruby-22-centos7
```
wtrocki: I'm curious how we can maintain this file for IOS, Android Cordova?
Do we have separate repository where common files like this one may land?

wtrocki: Since we will have couple platforms it will make sense to have one single place for automation (probably the target repository - github pages) where we could place this stuff.

wtrocki:
@finp I see no problem with the changes. We definitely could merge that. I have proposed alternative aproach - where target repository (global website etc. will have docker and html template files. Website could have scripts to fetch recent documentation from this repository and build themselves.

wtrocki:
PS: Digger will not be part of the SDK. It's part of the openshift extension.

wei-lee:
+1 on @wtrocki's suggestion. Not sure if the dockerfile and the templates need to live inside the SDK repo. That means you will have them duplicated across every SDKs. In the future if we want to change the templates it will require some extra work. Create a single repo to have the dockerfile and templates makes a lot sense to me.

docs/_stylesheets/asciibinder.css
```diff
@@ -0,0 +1,568 @@
+@import url(http://maxcdn.bootstrapcdn.com/font-awesome/4.1.0/css/font-awesome.min.css);
```
wtrocki: This seems to be generated by tooling as well. 
Let's remove it

wtrocki:
@finp Everything starts with `_` seems to be autogenerated by tooling.

wtrocki:
After moving docs main repository README has broken links.

finp:
@wtrocki @wei-lee  I've removed images and dockerfile, but would like to leave erb/js files from asciibinder because when asciibinder runs:
. it changes branches to produce html for each release
. it will expect these files (albeit maybe different versions) on each branch
 Currently we have only one distro in _distro.yml file, but as we release versions, this will expand and these files come into play

finp:
after discussing with @wtrocki , will remove asciibinder files

Issue #55, opened 2018-02-19
Application crashes on start when using the Android SDK


TommyJ1994:
## Expected Behavior
The Aerogear Android SDK can be added to my project by following the instructions. Following as per https://github.com/aerogear/aerogear-android-sdk/pull/52

## Actual Behavior
Application crashes on start when you reference the Aerogear Android SDK in your project.

```
Information:Gradle tasks [:app:assembleDebug]
Error:com.android.builder.dexing.DexArchiveBuilderException: Failed to process /home/tjackman/.gradle/caches/transforms-1/files-1.1/core-5.0.0-SNAPSHOT.aar/022238cf764ef72c2b4febd459b7ebbf/jars/classes.jar
Error:com.android.builder.dexing.DexArchiveBuilderException: Error while dexing org/aerogear/mobile/core/http/OkHttpResponse.class
Error:com.android.dx.cf.code.SimException: invalid opcode ba (invokedynamic requires --min-sdk-version >= 26)
Error:Execution failed for task ':app:transformClassesWithDexBuilderForDebug'.
> com.android.build.api.transform.TransformException: com.android.builder.dexing.DexArchiveBuilderException: com.android.builder.dexing.DexArchiveBuilderException: Failed to process /home/tjackman/.gradle/caches/transforms-1/files-1.1/core-5.0.0-SNAPSHOT.aar/022238cf764ef72c2b4febd459b7ebbf/jars/classes.jar
Information:BUILD FAILED in 3s
Information:4 errors
Information:0 warnings
Information:See complete output in console
```

## Steps to Reproduce the Problem

  1. Follow the getting started guide to add the aerogear android SDK to this project - https://github.com/feedhenry/mobile-security-android-template
  2. Try to run the application here using this PR https://github.com/feedhenry/mobile-security-android-template/pull/33/files
  

**Application build.gradle**
```
apply plugin: 'com.android.application'

android {
    compileSdkVersion 27
    buildToolsVersion '26.0.2'
    defaultConfig {
        applicationId "com.feedhenry.securenativeandroidtemplate"
        minSdkVersion 21
        targetSdkVersion 27
        versionCode 1
        versionName "1.0"
        testInstrumentationRunner "com.feedhenry.securenativeandroidtemplate.MockTestRunner"
        manifestPlaceholders = ['appAuthRedirectScheme': 'com.feedhenry.securenativeandroidtemplate']
        vectorDrawables {
            useSupportLibrary true
        }
    }
[...]

PR #56, opened 2018-02-19
Troubleshooting doc

danielpassos:


docs/troubleshooting.adoc
```diff
@@ -0,0 +1,19 @@
+== AeroGear Android Troubleshooting
+
+=== invokedynamic requires --min-sdk-version >= 26
```
wtrocki: I will change title to be less cryptic. `Problems with building on Android SDK lower than 26` ?

danielpassos: Not sure because that is **exactly** what logcat says. See: https://github.com/aerogear/aerogear-android-sdk/issues/55

docs/getting-started/getting-started.adoc
```diff
@@ -13,7 +14,8 @@ It is reccomended you have the following installed
  * The latest version of http://openjdk.java.net/install/index.html[OpenJDK]
 
 == Installing
-As of the writing of this guide, the Aerogear Android SDK is not in jcenter and needs to be installed manually.
+
+As of the writing of this guide, the Aerogear Android SDK is not publish anywhere yet and needs to be installed manually.
```
secondsun: How about "is not published in jcenter" instead of "is not published anywhere"

danielpassos: Back to the original one!

docs/troubleshooting.adoc
secondsun: Perhaps "Build error : invokedynamic requires --min-sdk-version >= 26" will be a better title?

PR #58, opened 2018-02-19
Flag required fields

wtrocki:
## Motivation

Allow to have app id on the top level so we will be able to retrieve metrics for specific application (if needed)
Add timestamp as in proposal to track when metrics were collected and send 

WIP idea for additional requirements.

## Tech notes

Two ways of doing this:
1. Change way we store metrics from `Metrics[]` to separate class containing it as field. 
1. Include that in publisher (probably worse) - implemented in this PR

## Tasks

- [x] Implementation
- [x] Adjust unit tests

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
@@ -7,7 +7,7 @@ import org.aerogear.mobile.core.http.HttpRequest;
 import org.aerogear.mobile.core.http.HttpResponse;
 import org.aerogear.mobile.core.metrics.Metrics;
 import org.aerogear.mobile.core.metrics.MetricsPublisher;
-import org.aerogear.mobile.core.utils.AppIdGenerator;
+import org.aerogear.mobile.core.utils.ClientIdGenerator;
 import org.json.JSONException;
 import org.json.JSONObject;
 
@@ -39,7 +39,7 @@ public class NetworkMetricsPublisher implements MetricsPublisher {
                 json.put(m.identifier(), new JSONObject(m.data()));
             }
             json.put("timestamp", System.currentTimeMillis());
-            json.put("clientId", AppIdGenerator.getOrCreateClientId(context));
+            json.put("clientId", ClientIdGenerator.getOrCreateClientId(context));
             httpRequest.post(url, json.toString().getBytes());
 
             MobileCore.getLogger().debug("Sending metrics");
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -32,14 +32,15 @@ public void publish(Metrics... metrics) {
             for (final Metrics m : metrics) {
                 json.put(m.identifier(), new JSONObject(m.data()));
             }
-
+            json.put("timestamp","TODO");
```
wtrocki: Ideally we will need to refactor from Metrics[] to individual object that will contain metrics. and wrap that into json object. 
Presenting this as idea as we may not need that. 

We could replicate this back to proposal to make sure that we have this covered for other platforms.

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -29,17 +34,19 @@ public void publish(Metrics... metrics) {
         try {
 
             JSONObject json = new JSONObject();
+            // Can we build DTO here instead of operating on JSONObject?
             for (final Metrics m : metrics) {
                 json.put(m.identifier(), new JSONObject(m.data()));
             }
-
+            json.put("timestamp", System.currentTimeMillis());
+            json.put("appId", AppIdGenerator.getOrCreateClientId(context));
```
wei-lee: not sure calling it `appId` is accurate as it is not an `app id`?

core/src/main/java/org/aerogear/mobile/core/utils/AppIdGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+import android.content.SharedPreferences;
+
+import org.aerogear.mobile.core.MobileCore;
+
+import java.util.UUID;
+
+/**
+ * Helper for generating ID's
+ */
+public class AppIdGenerator {
```
secondsun: -1 to the name.  Generally when I see appIds they are in the context of a identifier for an application for a service and is shared for all users of the application on that platform.

wtrocki: Yeah. Going to rename thay. . I will probably want to avoid static method as well

pb82: Actually, this is not about the App Id but the Client ID. We differentiate between them in the metrics (App Id is just the package name where the Client ID is the randomly generated and stored identifier). This could also be moved to a method in core, like `core.getDeviceId()`

pb82:
@wtrocki I think it's ok to implement that in the publisher. It's not super elegant and should we ever have more than two publishers we can always revise that. But it's just two additional lines and if we really want we can implement a Helper class with a method like `addIdAndTimestamp(metrics)` which pushes the top level object and adds id and timestamp.

wtrocki:
@danielpassos @secondsun Any comments?

PR #60, opened 2018-02-20
Add jwks manager

ziccardi:
## Motivation

JIRA: https://issues.jboss.org/browse/AGDROID-751

Taking over [PR-51](https://github.com/aerogear/aerogear-android-sdk/pull/51) and proceeding with the implementation

## Description

This PR will create a new JwksManager class to fetch and save the json web key set from the Keycloak server. 

The key set will be used to verify the local jwt token to make sure it is valid

## Progress

- [x] implement JwksManager
- [x] Fix existing unit tests and remove obsolete tests
- [x] use it to verify token
- [x] unit tests
- [x] integration tests using the sample app


auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -0,0 +1,184 @@
...
+
+    private final HttpServiceModule httpModule;
+    private final AuthServiceConfiguration authServiceConfiguration;
+    private final SharedPreferences sharedPrefs;
+    private static final Logger logger = MobileCore.getLogger();
+
+    public JwksManager(@NonNull final Context context,
+                       @NonNull final MobileCore mobileCore,
+                       @NonNull final AuthServiceConfiguration authServiceConfiguration) {
+        this.httpModule = nonNull(mobileCore, "mobileCore").getHttpLayer();
```
wei-lee: should we have another `nonNull` check on the httpModule as well?

ziccardi: it would be reduntant: getHttpLayer can't return null - https://github.com/ziccardi/aerogear-android-sdk/blob/d258418c93b74eaf5600a7bda4afc0f83b65192e/core/src/main/java/org/aerogear/mobile/core/MobileCore.java#L72-L72

auth/src/main/java/org/aerogear/mobile/auth/credentials/OIDCCredentials.java
```diff
@@ -67,71 +62,40 @@ public String getRefreshToken() {
...
-            throw new IllegalArgumentException(e);
-        } catch (InvalidKeySpecException e) {
-            Log.e(TAG, e.getMessage(), e);
-            // If the public key is invalid then we cannot determine the tokens integrity.
-            return false;
-        }
-
-
+    public boolean verifyClaims(final JsonWebKeySet jwks, final KeycloakConfiguration keycloakConfig) {
+        String issuer = keycloakConfig.getHostUrl() + "/realms/secure-app";
```
wei-lee: this should be returned from the `keycloakConfig` and it should be in the format of`{keycloakHost}/realms/{realmName}`

ziccardi:
@wei-lee done. Would you mind to give another look?

matzew:
+1

PR #61, opened 2018-02-20
AEROGEAR-2174 Initial interfaces and classes for SecurityService

aidenkeating:
## Motivation

Need initial interfaces to start development on implementation.

JIRA: https://issues.jboss.org/browse/AEROGEAR-2174

## Description

Add initial interfaces and classes

## Progress

- [x] Interfaces and classes
- [x] JavaDoc
- [x] Tests

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -30,6 +30,8 @@ android {
 dependencies {
     implementation project(path: ':core')
     implementation 'net.openid:appauth'
+    //noinspection GradleCompatible
+    implementation 'com.android.support:customtabs:26.1.0'
     implementation 'org.bitbucket.b_c:jose4j'
     implementation 'com.scottyab:rootbeer-lib:0.0.6'
     testImplementation "junit:junit"
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -31,6 +31,7 @@ dependencies {
     implementation project(path: ':core')
     implementation 'net.openid:appauth'
     implementation 'org.bitbucket.b_c:jose4j'
+    implementation 'com.scottyab:rootbeer-lib:0.0.6'
     testImplementation "junit:junit"
     testImplementation 'org.mockito:mockito-core'
     testImplementation 'org.robolectric:robolectric'
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -30,8 +30,8 @@ android {
 dependencies {
     implementation project(path: ':core')
     implementation 'net.openid:appauth'
-    //noinspection GradleCompatible
-    implementation 'com.android.support:customtabs:26.1.0'
+    implementation "com.android.support:customtabs:26.1.0"
+    implementation ('net.openid:appauth', { exclude group: 'com.android.support', module: 'customtabs' })
     implementation 'org.bitbucket.b_c:jose4j'
     implementation 'com.scottyab:rootbeer-lib:0.0.6'
     testImplementation "junit:junit"
--- a/auth/src/main/java/org/aerogear/mobile/security/Check.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/Check.java
@@ -5,8 +5,6 @@ import org.aerogear.mobile.security.checks.RootedCheck;
 public enum Check {
     /**
      *  Detect whether the device is rooted.
-     *
-     * @return <code>true</code> if the device is rooted.
      */
     IS_ROOTED(new RootedCheck());
 
--- a/auth/src/main/java/org/aerogear/mobile/security/Check.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/Check.java
@@ -1,0 +1,27 @@
+package org.aerogear.mobile.security;
+
+import org.aerogear.mobile.security.checks.RootedCheck;
+
+public enum Check {
+    /**
+     *  Detect whether the device is rooted.
+     *
+     * @return <code>true</code> if the device is rooted.
+     */
+    IS_ROOTED(new RootedCheck());
+
+    private SecurityCheck check;
+
+    Check(SecurityCheck check) {
+        this.check = check;
+    }
+
+    /**
+     * Return the {@link SecurityCheck} implementation for this check.
+     *
+     * @return
+     */
+    public SecurityCheck getSecurityCheck() {
+        return check;
+    }
+}
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheck.java
@@ -1,0 +1,13 @@
+package org.aerogear.mobile.security;
+
+import android.content.Context;
+
+public interface SecurityCheck {
+    /**
+     * Perform the check and return a result
+     *
+     * @param context Context to be used by the check.
+     * @return {@link SecurityCheckResult Result} of the test
+     */
+    SecurityCheckResult test(Context context);
+}
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheck.java
@@ -2,6 +2,12 @@ package org.aerogear.mobile.security;
 
 import android.content.Context;
 
+/**
+ * Interface for a single check to be executed.
+ *
+ * An example of a check would be detecting whether the device an application is running on is
+ * rooted.
+ */
 public interface SecurityCheck {
     /**
      * Perform the check and return a result
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
@@ -1,5 +1,7 @@
 package org.aerogear.mobile.security;
 
+import org.aerogear.mobile.core.metrics.MetricsService;
+
 public interface SecurityCheckExecutor {
     /**
      * Add a {@link Check check} to be executed on {@link #execute()}.
@@ -8,6 +10,13 @@ public interface SecurityCheckExecutor {
      * @return {@link SecurityCheckExecutor}
      */
     SecurityCheckExecutor addCheck(Check check);
+
+    /**
+     * Set that metrics should be sent on {@link #execute()} using the {@link MetricsService} provided.
+     *
+     * @return {@link SecurityCheckExecutor}
+     */
+    SecurityCheckExecutor sendMetrics(MetricsService metricsService);
 
     /**
      * Return the results of each test that was added to the executor.
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
@@ -1,0 +1,18 @@
+package org.aerogear.mobile.security;
+
+public interface SecurityCheckExecutor {
+    /**
+     * Add a {@link Check check} to be executed on {@link #execute()}.
+     *
+     * @param check The check to add.
+     * @return {@link SecurityCheckExecutor}
+     */
+    SecurityCheckExecutor addCheck(Check check);
+
+    /**
+     * Return the results of each test that was added to the executor.
+     *
+     * @return Array of {@link SecurityCheckResult results}
+     */
+    SecurityCheckResult[] execute();
+}
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
@@ -4,12 +4,20 @@ import org.aerogear.mobile.core.metrics.MetricsService;
 
 public interface SecurityCheckExecutor {
     /**
-     * Add a {@link Check check} to be executed on {@link #execute()}.
+     * Add a {@link SecurityCheckType check type} to be executed on {@link #execute()}.
      *
-     * @param check The check to add.
+     * @param securityCheckType The check type to add.
      * @return {@link SecurityCheckExecutor}
      */
-    SecurityCheckExecutor addCheck(Check check);
+    SecurityCheckExecutor addCheck(SecurityCheckType securityCheckType);
+
+    /**
+     * Add a {@link SecurityCheck} to be executed on {@link #execute()}.
+     *
+     * @param check The security check to add.
+     * @return {@link SecurityCheckExecutor}
+     */
+    SecurityCheckExecutor addCheck(SecurityCheck check);
 
     /**
      * Set that metrics should be sent on {@link #execute()} using the {@link MetricsService} provided.
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
@@ -2,6 +2,12 @@ package org.aerogear.mobile.security;
 
 import org.aerogear.mobile.core.metrics.MetricsService;
 
+/**
+ * Interface for performing multiple {@link SecurityCheck checks} and reporting their
+ * {@link SecurityCheckResult} results.
+ *
+ * Also allowing for the results of a check to be used as a metric.
+ */
 public interface SecurityCheckExecutor {
     /**
      * Add a {@link SecurityCheckType check type} to be executed on {@link #execute()}.
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckResult.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckResult.java
@@ -1,0 +1,18 @@
+package org.aerogear.mobile.security;
+
+public interface SecurityCheckResult {
+    /**
+     * Get the name of the security check that this is the result of. This can be used for
+     * reporting purposes.
+     *
+     * @return The name of the check
+     */
+    String getName();
+
+    /**
+     * Whether the check was successful.
+     *
+     * @return <code>true</code> if the check was successful.
+     */
+    boolean passed();
+}
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckResult.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckResult.java
@@ -1,5 +1,7 @@
 package org.aerogear.mobile.security;
 
+import org.aerogear.mobile.core.metrics.Metrics;
+
 public interface SecurityCheckResult {
     /**
      * Get the name of the security check that this is the result of. This can be used for
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckResult.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckResult.java
@@ -1,7 +1,8 @@
 package org.aerogear.mobile.security;
 
-import org.aerogear.mobile.core.metrics.Metrics;
-
+/**
+ * Interface for the results of a {@link SecurityCheck}.
+ */
 public interface SecurityCheckResult {
     /**
      * Get the name of the security check that this is the result of. This can be used for
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckType.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckType.java
@@ -2,7 +2,7 @@ package org.aerogear.mobile.security;
 
 import org.aerogear.mobile.security.checks.RootedCheck;
 
-public enum Check {
+public enum SecurityCheckType {
     /**
      *  Detect whether the device is rooted.
      */
@@ -10,7 +10,7 @@ public enum Check {
 
     private SecurityCheck check;
 
-    Check(SecurityCheck check) {
+    SecurityCheckType(SecurityCheck check) {
         this.check = check;
     }
 
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckType.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckType.java
@@ -2,6 +2,9 @@ package org.aerogear.mobile.security;
 
 import org.aerogear.mobile.security.checks.RootedCheck;
 
+/**
+ * Checks that can be performed.
+ */
 public enum SecurityCheckType {
     /**
      *  Detect whether the device is rooted.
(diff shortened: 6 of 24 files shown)


aidenkeating:
@wei-lee @ciaranRoche @rachael-oregan Would you mind taking a look? This is based on the comment on this Gist (https://gist.github.com/secondsun/2e674103c85363f0a602011d44b6b44a). Still requires some work but I wanted to get the concept reviewed quickly in case it's a no go.

This allows for chaining using the SecurityService.chain().method().method().run() syntax and singular tests using the SecurityService.method() syntax. There's a bit of repetition in repeating the method names though.

ziccardi:
Wouldn't be better to have something like this https://gist.github.com/ziccardi/5b7a4f8e1c4394b0f92ee9a5ae5450b2 ?

That is a kind of `command pattern`. That way adding and chaining new checks can be very easy and we won't need to add a new method to add a new check.

Moreover, if I didn't misunderstood, that would be closer to what has been discussed in https://issues.jboss.org/browse/AEROGEAR-2174.

@wei-lee @secondsun @aidenkeating WDYT?

rachael-oregan:
The approach so far looks good to me

wei-lee:
@ziccardi If you look at this gist: https://gist.github.com/secondsun/2e674103c85363f0a602011d44b6b44a.

I added a comment there and I think it will be good the have methods that explicitly state what it does. With the help of autocompletion it will be very easy for developers to find out what checks are available. Yes we will need to define an extra method but I am not too concerned about it.

When implementing the checks, I think we will certain follow the `command pattern`.

ziccardi:
@wei-lee I'm usually against classes with too many methods. If the only benefit we get is just for code completion, an Enum could do the trick as well. I updated my gist https://gist.github.com/ziccardi/5b7a4f8e1c4394b0f92ee9a5ae5450b2.

In that way, we would have code completion and chaining. Moreover, the user will be able to add custom security checks.

aidenkeating:
@ziccardi @wei-lee Made another branch based on that Gist (https://github.com/aerogear/aerogear-android-sdk/compare/master...aidenkeating:AEROGEAR-2174-alt). To me the approach in the Gist seems quite nice. So I'd be happy to go with that approach if you all are? (It also includes the rooted check as an example)

@rachael-oregan @ciaranRoche ^^

wei-lee:
@aidenkeating yes, I think it's quite nice. Nice work and good collaboration @aidenkeating @ziccardi

rachael-oregan:
+1 on the alternative approach.  It's that bit neater.  Nice work!

aidenkeating:
@wei-lee Based on this comment https://gist.github.com/secondsun/2e674103c85363f0a602011d44b6b44a . Is this kind of style what we want for `sendMetrics`?

We could also retrieve the metrics service from `MobileCore` inside the module instead of asking the user to provide it, or have both as an option.

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
```diff
@@ -0,0 +1,73 @@
...
+import org.aerogear.mobile.security.Check;
+import org.aerogear.mobile.security.SecurityCheck;
+import org.aerogear.mobile.security.SecurityCheckExecutor;
+import org.aerogear.mobile.security.SecurityCheckResult;
+import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
+
+import java.util.ArrayList;
+
+public class SecurityCheckExecutorImpl implements SecurityCheckExecutor {
+    private final ArrayList<SecurityCheck> checks;
```
ziccardi: Can we have the same check more than one time? If no, this should be a Set.
Moreover, it is a good practice to never declare a variable as ArrayList: better to declare it as List.
In our case, I think declaring it as Collection would be the best choice: that way we are able to change the real container without touching the rest of the code.

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private MetricsService metricsService;
+
+    public SecurityCheckExecutorImpl(Context context) {
+        this.checks = new ArrayList<>();
+        this.context = context;
+    }
+
+    @Override
+    public SecurityCheckExecutor addCheck(Check check) {
```
ziccardi: I think that we should add 

`public SecurityCheckExecutor addCheck(SecurityCheck check) {`

too. That way the user will be able to add custom checks as well

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    /**
+     * Return the {@link SecurityCheckResult results} for each of the tests added.
+     *
+     * @return Array of results.
+     */
+    private SecurityCheckResult[] getTestResults() {
+        ArrayList<SecurityCheckResult> results = new ArrayList<>();
```
ziccardi: Again here, better declare 
`List<SecurityCheckResult> results = new ArrayList<>();`

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        for (SecurityCheck check : checks) {
+            results.add(check.test(context));
```
ziccardi: Do we execute all the tests synchronously?

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckResultImpl.java
```diff
@@ -0,0 +1,24 @@
+package org.aerogear.mobile.security.impl;
+
+import org.aerogear.mobile.security.SecurityCheckResult;
+
+public class SecurityCheckResultImpl implements SecurityCheckResult {
+
+    private final String name;
+    private final boolean passed;
+
+    public SecurityCheckResultImpl(String name, boolean passed) {
```
ziccardi: Make parameter final

auth/src/main/java/org/aerogear/mobile/security/metrics/SecurityCheckResultMetric.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class SecurityCheckResultMetric implements Metrics {
+
+    private final String identifier;
+    private final Map<String, String> data;
+
+    public SecurityCheckResultMetric(SecurityCheckResult result) {
```
ziccardi: Make parameter final

auth/src/main/java/org/aerogear/mobile/security/metrics/SecurityCheckResultMetric.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Override
+    public String identifier() {
+        return identifier;
+    }
+
+    @Override
+    public Map<String, String> data() {
+        return data;
```
ziccardi: Never return reference to mutable internals.
Change this to:
`return Collections.unmodifiableMap(Data)`

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -145,7 +145,7 @@ public void destroy() {
             ServiceModule serviceModule = serviceClass.newInstance();
 
             if (serviceConfiguration == null) {
-                serviceConfiguration = getServiceConfiguration(serviceModule.type());
+                serviceConfiguration = getServiceConfiguration(serviceModule.type(), serviceModule.requiresConfiguration());
```
ziccardi: I think it would be better to leave the 
`getServiceConfiguration(serviceModule.type())` as it is, without the `requiresConfiguration` option and simply making it returns null if no configuration exists.
Then here we could do:
`if (config == null && serviceModule.requiresConfiguration()) {
    throw new ConfigurationNotFoundException(...)`
WDYT?

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -174,9 +173,9 @@ public Context getContext() {
      * @param type Service type/name
      * @return the configuration for this service from the JSON config file
      */
-    private ServiceConfiguration getServiceConfiguration(String type) {
+    private ServiceConfiguration getServiceConfiguration(String type, boolean requiresConfiguration) {
```
ziccardi: Make parameter final

example/src/main/java/org/aerogear/mobile/example/ui/SecurityServiceFragment.java
```diff
@@ -63,22 +68,18 @@
...
     }
 
-    @Nullable
     @Override
-    public View onCreateView(LayoutInflater inflater,@Nullable ViewGroup container,
-                              @Nullable Bundle savedInstanceState) {
-        view = inflater.inflate(R.layout.fragment_security_service, container, false);
-        ButterKnife.bind(this, view);
-
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
```
ziccardi: final parameter

auth/src/main/java/org/aerogear/mobile/security/Check.java
```diff
@@ -0,0 +1,25 @@
+package org.aerogear.mobile.security;
+
+import org.aerogear.mobile.security.checks.RootedCheck;
+
+public enum Check {
```
matzew: Check as in "security check" ?

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
matzew: and make it final :)

auth/src/main/java/org/aerogear/mobile/security/Check.java
aidenkeating: Yeah, I guess this could be `SecurityCheckType` or something like that

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
aidenkeating: Yep we do

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
aidenkeating: I'm happy to allow the service configuration to be null and have each service handle what happens in that scenario. @danielpassos @wei-lee Would you be fine with that?

ziccardi: @aidenkeating that's not exactly what I said. Only the service knows if they need or not a configuration, however, if they need it, it is managed centrally inside the `getInstance` method of the `MobileCore`. From the outside the result is the same, but IMO the code is cleaner this way (because from the POV of the `getConfiguration` method, not finding a configuration is not an exceptional case, so no exception should be raised).

aidenkeating: @ziccardi Apologies, misunderstanding, will make this change

aidenkeating:
@ziccardi Addressed comments, mind taking another quick look?

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private MetricsService metricsService;
+
+    public SecurityCheckExecutorImpl(Context context) {
+        this.checks = new HashSet<>();
+        this.context = context;
+    }
+
+    @Override
+    public SecurityCheckExecutor addCheck(SecurityCheckType securityCheckType) {
```
ziccardi: parameters should be final

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
```diff
@@ -0,0 +1,81 @@
...
+     * Return the {@link SecurityCheckResult results} for each of the tests added.
+     *
+     * @return Array of results.
+     */
+    private SecurityCheckResult[] getTestResults() {
+        final List<SecurityCheckResult> results = new ArrayList<>();
+        for (SecurityCheck check : checks) {
+            results.add(check.test(context));
+        }
+        return buildResultArray(results);
```
ziccardi: You know the number of results you are going to get since the beginning (`checks.size()`).
Using an `ArrayList` and then converting it into an array is useless here: use directly an array:

```java
SecurityCheckResult[] results = new SecurityCheckResults[checks.size()];
int i = 0;
for (Security check: checks) {
   results[i++] = check.test(context)
}
```

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private SecurityCheckResult[] buildResultArray(List<SecurityCheckResult> results) {
+        return results.toArray(new SecurityCheckResult[results.size()]);
```
ziccardi: Why a method for just one line? Better to inline this code.

auth/build.gradle
```diff
@@ -30,6 +30,8 @@ android {
 dependencies {
     implementation project(path: ':core')
     implementation 'net.openid:appauth'
+    //noinspection GradleCompatible
```
wojta: What about this?
`implementation "com.android.support:customtabs:26.1.0"
implementation ('net.openid:appauth', {
       exclude group: 'com.android.support', module: 'customtabs'
})`

aidenkeating: @wojta This seems to work too, would it be possible to get an example of what it does/why it helps? Not too familiar with this whole ecosystem

aidenkeating:
@ziccardi One more look when you get the chance?

@wojta Included your suggested change also, thanks 👍

auth/build.gradle
```diff
@@ -30,8 +30,8 @@ android {
 dependencies {
     implementation project(path: ':core')
     implementation 'net.openid:appauth'
-    //noinspection GradleCompatible
-    implementation 'com.android.support:customtabs:26.1.0'
+    implementation "com.android.support:customtabs:26.1.0"
```
wojta: I'm still actually not sure if this change is not dangerous. It simply replaces dependency, but `appauth` is still developed against 26.1.0.
What if we increase it to 27.0.2? 
```
implementation "com.android.support:customtabs:27.0.2"
    implementation "com.android.support:animated-vector-drawable:27.0.2"
    implementation "com.android.support:appcompat-v7:27.0.2"
    implementation('net.openid:appauth', {
        exclude group: 'com.android.support', module: 'customtabs'
    })
implementation('com.scottyab:rootbeer-lib:0.0.6', {
        exclude group: 'com.android.support', module: 'animated-vector-drawable'
        exclude group: 'com.android.support', module: 'appcompat-v7'
    })    
```
So far I tried it and app works still ok. Can we use it?

TommyJ1994: @wojta Might work because we have a hardcoded blacklist against the usage of custom tabs - see [here](https://github.com/aerogear/aerogear-android-sdk/blob/38a514725a31723d09038c7e59952cf1c857cae6/auth/src/main/java/org/aerogear/mobile/auth/authenticator/AuthorizationServiceFactory.java#L86-L88) (which needs to change). Have you tried without the blacklist?

ziccardi:
@aidenkeating just a couple of missing 'final' still there from my previous review.

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckType.java
```diff
@@ -0,0 +1,25 @@
+package org.aerogear.mobile.security;
+
+import org.aerogear.mobile.security.checks.RootedCheck;
+
+public enum SecurityCheckType {
```
matzew: let's have some javadoc

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckResult.java
```diff
@@ -0,0 +1,20 @@
+package org.aerogear.mobile.security;
+
+import org.aerogear.mobile.core.metrics.Metrics;
+
+public interface SecurityCheckResult {
```
matzew: let's have some javadoc

auth/src/main/java/org/aerogear/mobile/security/SecurityCheck.java
```diff
@@ -0,0 +1,13 @@
+package org.aerogear.mobile.security;
+
+import android.content.Context;
+
+public interface SecurityCheck {
```
matzew: let's have some javadoc

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutor.java
```diff
@@ -0,0 +1,35 @@
+package org.aerogear.mobile.security;
+
+import org.aerogear.mobile.core.metrics.MetricsService;
+
+public interface SecurityCheckExecutor {
```
matzew: let's have some javadoc

auth/build.gradle
aidenkeating: @danielpassos Any opinion on using this?

auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
```diff
@@ -0,0 +1,55 @@
+package org.aerogear.mobile.security;
+
+import org.aerogear.mobile.core.MobileCore;
+import org.aerogear.mobile.core.ServiceModule;
+import org.aerogear.mobile.core.configuration.ServiceConfiguration;
+import org.aerogear.mobile.core.metrics.MetricsService;
+import org.aerogear.mobile.security.impl.SecurityCheckExecutorImpl;
+import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
+
+public class SecurityService implements ServiceModule{
```
matzew: Service vs. Module ?

auth/src/main/java/org/aerogear/mobile/security/checks/RootedCheck.java
```diff
@@ -0,0 +1,25 @@
...
+
+import android.content.Context;
+
+import com.scottyab.rootbeer.RootBeer;
+
+import org.aerogear.mobile.security.SecurityCheck;
+import org.aerogear.mobile.security.SecurityCheckResult;
+import org.aerogear.mobile.security.impl.SecurityCheckResultImpl;
+
+public class RootedCheck implements SecurityCheck {
```
matzew: let's have some class level javadoc

auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
aidenkeating: With Auth and Metrics, the two most recent additions we went with Service. So I think it's service now

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import org.aerogear.mobile.security.SecurityCheckResult;
+import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+
+public class SecurityCheckExecutorImpl implements SecurityCheckExecutor {
+    private final Collection<SecurityCheck> checks;
```
matzew: why a `Collection`? instead of a `Set`? we can also initialise the thing on this line, instead of in the ctor

auth/src/main/java/org/aerogear/mobile/security/metrics/SecurityCheckResultMetric.java
```diff
@@ -0,0 +1,35 @@
...
+        return identifier;
+    }
+
+    @Override
+    public Map<String, String> data() {
+        return Collections.unmodifiableMap(data);
+    }
+
+    private Map<String, String> getDataFromResult(final SecurityCheckResult result) {
+        HashMap<String, String> data = new HashMap<>();
```
matzew: use `Map<String, String> data = ...` (and make it `final`)

auth/src/test/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+        executor.addCheck(securityCheckType).sendMetrics(metricsService).execute();
+
+        verify(metricsService, times(1)).publish(any());
+
+    }
+
+    @Test
+    public void testExecute() {
+        SecurityCheckResult[] results = executor.addCheck(securityCheckType).execute();
```
matzew: I think I generally don't understand why we do use that much arrays, instead of a generic collection (e.g. List/Set)

matzew: nothing on this PR - more a general thing :)

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
ziccardi: I suggested to use a `Collection` as interface and an `HashSet` as implementation.
The reason is that we need to have a set of unique objects, but we don't need any of the methods exposed by a `Set` that are not exposed in a `Collection`. So, I think the most generic interface should be used, so that if we should ever decide to change from HashSet to ArrayList, for example, the rest of the code will stay unchanged.

aidenkeating:
@ziccardi Yep I should probably make sure I addressed the comments before asking for another review 😛 

@ziccardi @matzew Would you mind taking another look?

auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
matzew: a service is not a module :)

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckExecutorImpl.java
matzew: fair enough

matzew:
@aidenkeating ship it !

auth/build.gradle
danielpassos: If we are not going to use this I see no problem

PR #70, opened 2018-02-22
Fixing broken toString() impl, and adding test

matzew:
## Motivation

The `toString(`) was buggy, not correctly printing the names, also the implemtation was doing String concatenation. it also lacked a unit test
 
## Description

fixed impl + test

## Progress

- [x] Done Task

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
@@ -157,21 +157,24 @@ public class UserPrincipalImpl implements UserPrincipal {
 
     @Override
     public String toString() {
-        String roleNames = "";
-        Iterator<UserRole> i = roles.iterator();
-        if (i.hasNext()) {
-            //first element
-            roleNames.concat("[").concat(i.next().getName());
-            while(i.hasNext()) {
-                roleNames.concat(", ").concat(i.next().getName());
+        final StringBuilder sb = new StringBuilder();
+        sb.append("[");
+
+        // TODO: use .stream().map(name -> name.getname())... when API 24
+        final Iterator<UserRole> iterator = roles.iterator();
+        if (iterator.hasNext()) {
+            sb.append(iterator.next().getName());
+
+            while (iterator.hasNext()) {
+                sb.append(", ").append(iterator.next().getName());
             }
         }
-        roleNames.concat("]");
+        sb.append("]");
 
         return "UserPrincipalImpl{" +
                 "username='" + username + '\'' +
                 ", email='" + email + '\'' +
-                ", roles=" + roleNames +
+                ", roles=" + sb.toString() +
                 '}';
     }
 
--- a/auth/src/test/java/org/aerogear/mobile/auth/user/UserPrincipalImplTest.java
+++ b/auth/src/test/java/org/aerogear/mobile/auth/user/UserPrincipalImplTest.java
@@ -1,10 +1,8 @@
 package org.aerogear.mobile.auth.user;
 
-import org.aerogear.mobile.auth.user.RoleType;
-import org.aerogear.mobile.auth.user.UserRole;
-import org.aerogear.mobile.auth.user.UserPrincipalImpl;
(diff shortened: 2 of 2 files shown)


auth/src/test/java/org/aerogear/mobile/auth/user/UserPrincipalImplTest.java
```diff
@@ -46,6 +43,14 @@ public void testHasRealmRoleSucceeds(){
         assertEquals(userPrincipalImpl.hasRealmRole("rRole"), true);
     }
 
+    @Test
+    public void testToStringImpl(){
+        assertEquals(
+            "UserPrincipalImpl{username='test-user', email='null', roles=[rRole, cRole]}",
```
matzew: @rachael-oregan when adding this test to the existing code base... the `roles=...` part was empty

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -157,21 +157,24 @@ public static Builder newUser() {
...
     public String toString() {
-        String roleNames = "";
-        Iterator<UserRole> i = roles.iterator();
-        if (i.hasNext()) {
-            //first element
-            roleNames.concat("[").concat(i.next().getName());
-            while(i.hasNext()) {
-                roleNames.concat(", ").concat(i.next().getName());
+        final StringBuffer sb = new StringBuffer();
+        sb.append("[");
```
matzew: when there are no roles... we want to render to following -> `[]`, right ?

auth/src/test/java/org/aerogear/mobile/auth/user/UserPrincipalImplTest.java
```diff
@@ -46,6 +43,11 @@ public void testHasRealmRoleSucceeds(){
         assertEquals(userPrincipalImpl.hasRealmRole("rRole"), true);
     }
 
+    @Test
+    public void testToStringImpl(){
+        assertEquals("UserPrincipalImpl{username='test-user', email='null', roles=[cRole, rRole]}", userPrincipalImpl.toString());
```
matzew: @rachael-oregan when adding this test to the existing code base... the roles=... part was empty

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
matzew: @rachael-oregan  ^

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -157,21 +157,24 @@ public static Builder newUser() {
...
     @Override
     public String toString() {
-        String roleNames = "";
-        Iterator<UserRole> i = roles.iterator();
-        if (i.hasNext()) {
-            //first element
-            roleNames.concat("[").concat(i.next().getName());
-            while(i.hasNext()) {
-                roleNames.concat(", ").concat(i.next().getName());
+        final StringBuffer sb = new StringBuffer();
```
ziccardi: Use `StringBuilder` instead of `StringBuffer` to avoid useless synchronization

matzew:
@ziccardi changed

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -157,21 +157,24 @@ public static Builder newUser() {
...
     public String toString() {
-        String roleNames = "";
-        Iterator<UserRole> i = roles.iterator();
-        if (i.hasNext()) {
-            //first element
-            roleNames.concat("[").concat(i.next().getName());
-            while(i.hasNext()) {
-                roleNames.concat(", ").concat(i.next().getName());
+        final StringBuilder sb = new StringBuilder();
+        sb.append("[");
```
matzew: when there are no roles... we want to render to following -> [], right ?

rachael-oregan: Yes that's correct

PR #72, opened 2018-02-22
Refactors tests and improves coverage

josemigallas:
## Motivation

Review test quality and coverage status of current projects and find improvements.

[No Jira at the moment]

## Description
This only affects `MobileCore`.
A bit of refactorization to `MobileCoreTest` class, to make it more extendable.
Also some new test cases to increase coverage 79% -> 90% .

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -28,6 +28,8 @@ import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
  */
 public final class MobileCore {
 
+    public static final String DEFAULT_CONFIG_FILE_NAME = "mobile-services.json";
+
     private static final String TAG = "AEROGEAR/CORE";
     private static Logger logger = new LoggerAdapter();
     private static String appVersion;
@@ -221,8 +223,6 @@ public final class MobileCore {
 
     @SuppressWarnings({"UnusedReturnValue", "WeakerAccess"})
     public static final class Options {
-
-        public static final String DEFAULT_CONFIG_FILE_NAME = "mobile-services.json";
 
         private String configFileName = DEFAULT_CONFIG_FILE_NAME;
         // Don't have a default implementation because it should use configuration
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -3,6 +3,7 @@ package org.aerogear.mobile.core;
 import android.content.Context;
 import android.content.pm.PackageManager;
 import android.support.annotation.NonNull;
+import android.support.annotation.VisibleForTesting;
 
 import org.aerogear.android.core.BuildConfig;
 import org.aerogear.mobile.core.configuration.MobileCoreJsonParser;
@@ -199,6 +200,7 @@ public final class MobileCore {
         return logger;
     }
 
+    @VisibleForTesting()
     public String getConfigFileName() {
         return configFileName;
     }
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -197,6 +197,10 @@ public final class MobileCore {
         return logger;
     }
 
+    public String getConfigFileName() {
+        return configFileName;
+    }
+
     /**
      * Get the version name of the SDK itself
      *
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -207,6 +206,10 @@ public static Logger getLogger() {
         return logger;
     }
 
+    public String getConfigFileName() {
```
danielpassos: Dunno if I wanna expose it only to be visible in the tests. Maybe you should use [@VisibleForTesting](https://developer.android.com/reference/android/support/annotation/VisibleForTesting.html)

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -225,9 +228,12 @@ public static String getAppVersion() {
         return appVersion;
     }
 
+    @SuppressWarnings({"UnusedReturnValue", "WeakerAccess"})
     public static final class Options {
 
-        private String configFileName = "mobile-services.json";
+        public static final String DEFAULT_CONFIG_FILE_NAME = "mobile-services.json";
+
+        private String configFileName = DEFAULT_CONFIG_FILE_NAME;
```
danielpassos: Do we really need this redundant?

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
josemigallas: Didn't know that existed, it looks just like what I wanted

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
josemigallas: I agree it is a bit redundant in this case. However I think it is a good practice to extract constants out of the code, always. In this case it is importante to make clear this string is a default value for the entire application, that is not going to change ever. Actually I would put it out of MobileCore even. It seems to me like a too relevant piece of string to be just hardcoded.

josemigallas: Plus, this way it can be tested more robustly.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -225,9 +228,12 @@ public static String getAppVersion() {
         return appVersion;
     }
 
+    @SuppressWarnings({"UnusedReturnValue", "WeakerAccess"})
     public static final class Options {
 
-        private String configFileName = "mobile-services.json";
+        public static final String DEFAULT_CONFIG_FILE_NAME = "mobile-services.json";
```
matzew: perhaps move up? our of the nested `Options` ?

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
danielpassos: That is why I'm here! :bowtie:

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
josemigallas: Yes it would be better out. In `MobileCore`? Or some other class, maybe `MobileCoreContants` or something like that. WDYT?

pb82: I think moving it to `MobileCore` is fine. Just to keep all the public static's at the same place.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
pb82: But why not just use `DEFAULT_CONFIG_FILE_NAME` in the code instead of `configFileName`?

core/src/test/java/org/aerogear/mobile/core/MobileCoreTest.java
```diff
@@ -26,12 +28,25 @@
 @SmallTest
 public class MobileCoreTest {
 
-    @Test
-    public void testInit() {
+    private static final String DUMMY_MOBILE_SERVICES_JSON = "dummy-mobile-services.json";
+    private static final String EMPTY_MOBILE_SERVICES_JSON = "empty-mobile-services.json";
+
+    private MobileCore core;
+    private MobileCore dummyCore;
```
pb82: why `core` and `dummyCore`?

josemigallas: Core is using default config file with the default services (the production-like version if you will) and dummycore is for testing weird cases, dummy services and so on, it has a custom config file made only for testing.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -234,11 +234,6 @@ public static String getAppVersion() {
         public Options() {
         }
 
-        public Options(@NonNull final String configFileName, @NonNull final HttpServiceModule httpServiceModule) {
```
josemigallas: @danielpassos I wasn't sure of removing this but I can see that Options is meant to be used like fluent api
```java
new Options()
    .setConfigFileName()
    .setFoo()
```
Plus is not being used used anywhere. WDYT?

danielpassos: If it's not used... 

Kill it with fire

secondsun:
@josemigallas can you rebase on master and I will merge

josemigallas:
@secondsun done

PR #73, opened 2018-02-22
Core fixes

matzew:
## Motivation

:lipstick:  a bit of code cosmetics

## Progress

- [x] Done Task

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -128,8 +128,9 @@ public final class MobileCore {
 
     @SuppressWarnings("unchecked")
     public <T extends ServiceModule> T getInstance(final Class<T> serviceClass,
-                                                   ServiceConfiguration serviceConfiguration)
+                                                   final ServiceConfiguration serviceConfiguration)
         throws InitializationException {
+        nonNull(serviceClass, "serviceClass");
 
         if (services.containsKey(serviceClass)) {
             return (T) services.get(serviceClass);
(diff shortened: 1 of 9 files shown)


matzew:
@ziccardi sounds good to me - not that familiar w/ the codebase. perhaps worth a jira ? @pb82  ?

pb82:
@ziccardi with sanity checks, do you mean assertions?

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -17,7 +17,7 @@ public MetricsPublisher getPublisher() {
         return publisher;
     }
 
-    public MetricsService setPublisher(MetricsPublisher publisher) {
+    public MetricsService setPublisher(final MetricsPublisher publisher) {
         if (publisher == null) {
```
ziccardi: @pb82 This could be changed to use the `SanityCheck` class

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -31,13 +31,13 @@ public String type() {
     }
 
     @Override
-    public void configure(MobileCore core, ServiceConfiguration serviceConfiguration) {
+    public void configure(final MobileCore core, final ServiceConfiguration serviceConfiguration) {
         defaultMetrics = new Metrics[]{
             new AppMetrics(core.getContext()),
```
ziccardi: @pb82 This will lead to a NullPointerException if `core` os `serviceConfiguration` is null

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
```diff
@@ -11,13 +11,13 @@
...
-    public LoggerMetricsPublisher(Logger logger) {
+    public LoggerMetricsPublisher(final Logger logger) {
         this.logger = logger;
     }
 
     @Override
-    public void publish(Metrics... metrics) {
-        for (Metrics m : metrics) {
+    public void publish(final Metrics... metrics) {
+        for (final Metrics m : metrics) {
```
ziccardi: @pb82 and here another example, if metrics are null

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -142,7 +136,7 @@ public void destroy() {
         }
 
         try {
-            ServiceModule serviceModule = serviceClass.newInstance();
+            final ServiceModule serviceModule = serviceClass.newInstance();
 
             if (serviceConfiguration == null) {
                 serviceConfiguration = getServiceConfiguration(serviceModule.type());
```
ziccardi: Changing parameters value is a bad practice

matzew: yeah - for a quick fix, I am doing a local var

PR #74, opened 2018-02-22
Clean up

josemigallas:
## Motivation

Maintain code base as tidy and consistent as possible.

## Description

Cleanup consist of: reformat code and remove unused imports.
Cleanup do not consist of: refactor code or change logic in any way.

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/AuthStateManager.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/AuthStateManager.java
@@ -24,6 +24,7 @@ public class AuthStateManager {
 
     /**
      * Reads credentials from storage.
+     *
      * @return OIDCCredentials
      */
     public OIDCCredentials load() {
@@ -36,6 +37,7 @@ public class AuthStateManager {
 
     /**
      * Saves a token
+     *
      * @param authState token to be saved
      * @throws IllegalStateException
      */
@@ -44,7 +46,7 @@ public class AuthStateManager {
             clear();
         } else {
             SharedPreferences.Editor e = prefs.edit().putString(KEY_STATE, authState.serialize());
-            if(!e.commit()) {
+            if (!e.commit()) {
                 throw new IllegalStateException("Failed to update state from shared preferences");
             }
         }
@@ -52,6 +54,7 @@ public class AuthStateManager {
 
     /**
      * Deletes a token
+     *
      * @throws IllegalArgumentException
      */
     public synchronized void clear() {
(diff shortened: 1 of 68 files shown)


wtrocki:
> Maintain code base as tidy and consistent as possible.
-1 for doing that this way. Let's talk about standards on the list. Get required tooling into gradle and extend build to fail when changes do not conform with the standard. Making some manual changes are not improving overall state of the project because they are really presenting your point of view that most of the SDK developers will not be aware of and still push changes the old way.

danielpassos:
@josemigallas Just wondering if you should not kill/merge too much cleanup commits in only one!

wei-lee:
agree with @wtrocki. It's ok to merge this PR but I think it's more important to create a style/editor configuration to make sure we all use the same settings.

danielpassos:
@wei-lee @wtrocki Why not use the default one provided by Android Studio/Google?

josemigallas:
@wtrocki I see your point however this is not my point of view but the default Android Studio formatting rules, which I believe meets the most extended styling rules for Java.

@wei-lee That's true and also my intention. In the meanwhile I though why not updating the code for now.

wei-lee:
@danielpassos I am using the default one in my Android Studio. So why do we still have the problem with formatting, are we missing a check/build step somewhere?

danielpassos:
@wei-lee Because people forgot to apply it :bowtie:

My point is... we don't need to **create a style/editor configuration** we need to make sure it was applied when running ci

auth/src/main/java/org/aerogear/mobile/auth/AuthStateManager.java
```diff
@@ -44,14 +46,15 @@ public synchronized void save(final OIDCCredentials authState) {
             clear();
         } else {
             SharedPreferences.Editor e = prefs.edit().putString(KEY_STATE, authState.serialize());
```
matzew: make `e` final

auth/src/main/java/org/aerogear/mobile/auth/Callback.java
```diff
@@ -1,6 +1,8 @@
 package org.aerogear.mobile.auth;
 
 public interface Callback<T extends Object> {
```
matzew: Let's add some javadoc

josemigallas:
@wei-lee @wtrocki @danielpassos If everyone agrees, I propose closing this PR and opening a new one, with all the necessary changes to maintain code clean, that is: cleanup following just the default IDE formatting rules and necessary additions to CI tool to check lint and also some gradle task that formats code while building, if possible.

matzew:
> also some gradle task that formats code while building

@josemigallas let's please not do that. This will cause large PRs, where lot's of distraction is on awfully formatted code.  The build should fail if things are not correctly formatted

wei-lee:
@josemigallas +1. Let's run the checks during build time and fail it if there is code is not formatted/linted correctly.

josemigallas:
OK then. Follow up task https://issues.jboss.org/browse/AGDROID-763

PR #75, opened 2018-02-22
Reimplemented check for readiness state of the AuthService service by…

ziccardi:
## Motivation

The current implementation of the check readiness for the AuthState service was cumbersome.

## Description

Changed the implementation of the way we check for readiness removing bitwise operation and replacing with EnumSet

## Progress

- [x] Replace bitwise with EnumSet
- [x] Check that tests still pass
- [x] Verify the example app still works

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
@@ -21,6 +21,7 @@ import org.jose4j.jwk.JsonWebKeySet;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.List;
 
 import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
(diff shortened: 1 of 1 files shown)


auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
```diff
@@ -207,15 +172,16 @@ public void configure(final MobileCore core, final ServiceConfiguration serviceC
         this.serviceConfiguration = nonNull(serviceConfiguration, "serviceConfiguration");
         this.keycloakConfiguration = new KeycloakConfiguration(serviceConfiguration);
 
-        status.updateStatus(ReadynessStatus.STEP.CONFIGURED);
+        //status.updateStatus(ReadynessStatus.STEP.CONFIGURED);
```
wei-lee: remove the comment

ziccardi: good catch. I missed that! Thanks!

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
```diff
@@ -45,104 +46,68 @@
...
 
-        /**
-         * Invoked to check if the service is ready. If it is not ready, an IllegalStateException is
-         * thrown with an appropriate error message.
-         */
-        public void checkReadyness() {
+    /**
+     * Stores the list of executed initialisation steps. When started, all steps must be executed.
+     */
+    private EnumSet<STEP> initialisationStatus = EnumSet.noneOf(STEP.class);
```
matzew: wanna make it `final` ? :smile:

PR #76, opened 2018-02-22
non null for passed in arguments should throw IllegalArgumentException

matzew:
## Motivation

If a null object is passed in, as an _argument_, and `IllegalArgumentException` needs to be thrown. NPE is only good when you invoke a null objects. But for checks on passed in objects, it's `IllegalArgumentException`

see: https://stackoverflow.com/questions/3881/illegalargumentexception-or-nullpointerexception-for-a-null-parameter

## Progress

- [x] Done Task


danielpassos:
👍

PR #78, opened 2018-02-22
 ✨  add instructions on how to support self-signed certs. Update the

wei-lee:
example app to make sure it can work with self signed certs

## Motivation

Make sure the example app can work with self-signed certs for local development and testing

## Description

* Add instructions on how to support self-signed certs
* Update the example app to load additional user added CAs
* Remove `ConnectionBuilderForTesting` as it is not safe to have it in the lib

## Progress

- [x] Done Task

## Additional Notes

To test:
* Provision Keycloak using mobile core. It should be running with https and use self-signed certs
* Follow the instructions in this PR to support the self-signed cert
* Update the configuration file of the example app to use the local keycloak instance
* Try user authentication from the example app. It should continue to work.


auth/src/main/java/org/aerogear/mobile/auth/ConnectionBuilderForTesting.java
```diff
@@ -1,112 +0,0 @@
...
-import javax.net.ssl.SSLSession;
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.X509TrustManager;
-/**
- * An example implementation of {@link ConnectionBuilder} that permits connecting to http
- * links, and ignores certificates for https connections. *THIS SHOULD NOT BE USED IN PRODUCTION
- * CODE*. It is intended to facilitate easier testing of AppAuth against development servers
- * only.
- */
-public class ConnectionBuilderForTesting implements ConnectionBuilder {
```
matzew: I was wondering about removing that thing, too 

why was it added ? should it live in tests ?  @ziccardi ?

ziccardi: It was there because, without that, untrusted certificates are... untrusted ;)
I don't think we need it in the tests. It was just meant to make life easier in development environments.

PR #82, opened 2018-02-22
Some fixes on OIDCCredential clazz

matzew:
## Motivation

throwing `UnsupportedOperationException` in `renew` and adding a test

## Progress

- [x] Done Task

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/credentials/OIDCCredentials.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/credentials/OIDCCredentials.java
@@ -26,8 +26,6 @@ import static org.aerogear.mobile.core.utils.SanityCheck.nonEmpty;
 public class OIDCCredentials {
 
     private static final String TAG = "OIDCCredentials";
-    private static final String beginPublicKey = "-----BEGIN PUBLIC KEY-----";
-    private static final String endPublicKey = "-----END PUBLIC KEY-----";
 
     private final AuthState authState;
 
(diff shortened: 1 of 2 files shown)


auth/src/main/java/org/aerogear/mobile/auth/credentials/OIDCCredentials.java
```diff
@@ -26,8 +26,6 @@
 public class OIDCCredentials {
 
     private static final String TAG = "OIDCCredentials";
-    private static final String beginPublicKey = "-----BEGIN PUBLIC KEY-----";
-    private static final String endPublicKey = "-----END PUBLIC KEY-----";
```
matzew: these two were not used ...

auth/src/test/java/org/aerogear/mobile/auth/credentials/OIDCCredentialsTest.java
```diff
@@ -71,7 +72,13 @@ public void testSerialize() throws JSONException {
         assertEquals(serializedCredential.get("authState"), CREDENTIAL_AUTH_STATE);
     }
 
-    @Test
+    @Test(expected = UnsupportedOperationException.class)
+    public void testRenew() throws AuthenticationException {
```
ziccardi: Does it make sense to create a test for a 'not yet implemented' method?

matzew: IMO this enforces us to write a test, once implemented :-)

PR #83, opened 2018-02-22
Renaming OICDAuthOptions to DefaultAuthOptions

matzew:
## Motivation

there was no OIDC on that class

## Progress

- [x] Done Task


matzew:
@wei-lee as discussed on IRC 

I am keeping the marker interface... just to be safe

PR #85, opened 2018-02-22
correct function name

TommyJ1994:
## Motivation
Tiny update to correct the function name - I think it is supposed to be `handleAuthResult()`?


TommyJ1994:
@aidenkeating Mind taking a quick look?

PR #86, opened 2018-02-22
AEROGEAR-1978 - Add a Check to Detect Developer Mode Enabled

rachael-oregan:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-1978

## Description

Add another check `DeveloperModeCheck` for the Security Service 

## Progress

- [x] Developer Mode Security Check
- [x] Java Docs
- [x] Unit Tests

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
@@ -1,0 +1,29 @@
+package org.aerogear.mobile.security.checks;
+
+import android.content.Context;
+import android.provider.Settings;
+
+import org.aerogear.mobile.security.SecurityCheck;
+import org.aerogear.mobile.security.SecurityCheckResult;
+import org.aerogear.mobile.security.impl.SecurityCheckResultImpl;
+
+
+/**
+ * Security check that detects if developer mode is enabled in the device.
+ */
+public class DeveloperModeCheck implements SecurityCheck {
+    private static final String NAME = "detectDeveloperMode";
+
+    /**
+     * Check if developer mode has been enabled in the device.
+     *
+     * @param context Context to be used by the check.
+     * @return <code>true</code> if the developer options have been enabled on the device.
+     */
+    @Override
+    public SecurityCheckResult test(final Context context) {
+        int devOptions = Settings.Secure.getInt(context.getContentResolver(), Settings.Global.DEVELOPMENT_SETTINGS_ENABLED, 0);
+        boolean devOptionsEnabled = devOptions > 0 ? true : false;
+        return new SecurityCheckResultImpl(NAME, devOptionsEnabled);
+    }
+}
--- a/auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
@@ -15,8 +15,6 @@ import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
  * Security check that detects if developer mode is enabled in the device.
  */
 public class DeveloperModeCheck implements SecurityCheck {
-    private static final String NAME = "detectDeveloperMode";
-
     /**
      * Check if developer mode has been enabled in the device.
      *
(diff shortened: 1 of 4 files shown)


auth/src/test/java/org/aerogear/mobile/security/checks/DeveloperModeCheckTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+        DeveloperModeCheck developerModeCheck = new DeveloperModeCheck();
+
+        SecurityCheckResult actual = developerModeCheck.test(context);
+
+        assertEquals(expected.getName(), actual.getName());
+        assertEquals(expected.passed(), actual.passed());
+    }
+
+    @Test(expected = NullPointerException.class)
```
rachael-oregan: @wei-lee just refactored this test.  Not sure if this test adds much value, WDYT?

rachael-oregan: @aidenkeating ^^

wei-lee: yeah, I don't think so, but you already have it, so let's keep it. I think it's better to have a sanity check in the implementation to make sure `context` can not be null use the `SanityCheck` class from core.

auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Check if developer mode has been enabled in the device.
+     *
+     * @param context Context to be used by the check.
+     * @return <code>true</code> if the developer options have been enabled on the device.
+     */
+    @Override
+    public SecurityCheckResult test(final Context context) {
+        int devOptions = Settings.Secure.getInt(context.getContentResolver(), Settings.Global.DEVELOPMENT_SETTINGS_ENABLED, 0);
```
matzew: I'd annotated the passed in `Context` with `@NonNull`, and than also do a sanity check. E.g. replace the line with:

```
int devOptions = Settings.Secure.getInt(nonNull(context, "context").getContentResolver(), Settings.Global.DE......
```

auth/src/test/java/org/aerogear/mobile/security/checks/DeveloperModeCheckTest.java
matzew: The SantiyCheck (see my comment above), returns an `IllegalArgumentException`, since passing in null, is an illegal argument :-)

auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
matzew: @rachael-oregan  ^

auth/src/test/java/org/aerogear/mobile/security/checks/DeveloperModeCheckTest.java
matzew: @rachael-oregan  ^

auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
```diff
@@ -0,0 +1,29 @@
...
+        boolean devOptionsEnabled = devOptions > 0 ? true : false;
+        return new SecurityCheckResultImpl(NAME, devOptionsEnabled);
```
ziccardi: These 2 lines can be changed to 
`return new SecurityCheckResultImpl(NAME, devOptions > 0);`

rachael-oregan:
@wei-lee @matzew @ziccardi thanks for the review guys. I've addressed your comments if ye have time for another look thanks

wei-lee:
@rachael-oregan I think you need to rebase this PR. You will need to implement `getName` method in the check class, it should return the full class name. See #90

ziccardi:
@rachael-oregan Wait for [PR-94](https://github.com/aerogear/aerogear-android-sdk/pull/94) to ibe merged, then rebase. That PR will provide a default `getName` implementation and will force a small change on how the `ServiceResultImpl` must be instantiated (a reference to the ServiceCheck must be passed - ie `this` - instead of a free string).

rachael-oregan:
@wei-lee @ziccardi mind having another look please

wei-lee:
LGTM

PR #87, opened 2018-02-23
Added Async executor for SecurityChecks

ziccardi:
Refactored SecurityCheckExecutor class

## Motivation

Current code allows only synchronous security check execution.

## Description

Added a way to execute the checks asynchronously

## Progress

- [x] Add AsyncSecurityCheckExecutor
- [x] Refactor current SecurityCheckExecutor
- [x] Update current unit tests
- [x] Add unit tests
- [x] Add Sanity Check
- [x] Add JavaDocs


auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor {
+
+    private final ExecutorService pool = Executors.newFixedThreadPool(10);
+
+    AsyncSecurityCheckExecutor(final Context context, final Collection<SecurityCheck> checks, final MetricsService metricsService) {
+        super(context, checks, metricsService);
+    }
+
+    public Future<SecurityCheckResult>[] execute() {
```
matzew: can we change the interface ?

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -0,0 +1,41 @@
...
+import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
+
+import java.util.Collection;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor {
+
+    private final ExecutorService pool = Executors.newFixedThreadPool(10);
```
wei-lee: the core module provides AppExecutor, it's best to keep using it.

wei-lee: Also not sure if we actually need a pool of executors. For the security checks, I think it's ok for them to all happen on one background thread. Using a pool seems a little bit too much.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -0,0 +1,41 @@
...
+
+        final Future[] res = new Future[checks.size()];
+
+        int i = 0;
+        for (final SecurityCheck check : checks) {
+            res[i++] = (pool.submit(() -> {
+                final SecurityCheckResult result =  check.test(getContext());
+                if (metricsService != null) {
+                    metricsService.publish(new SecurityCheckResultMetric(result));
+                }
```
wei-lee: not sure i like to public a metric every time when a test is done. I prefer to send them in one go

ziccardi: I think it would be a bit overkill to wait for all checks to be executed just to send the metric in one go. Do we really want that even for asynchronous execution?

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
ziccardi: Yeah, was going to add a method to set the executor in the builder

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
ziccardi: @wei-lee @aidenkeating What do you think? I'm ok with changing the interface to return a `Collection` instead of an Array

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -0,0 +1,64 @@
...
+                return result;
+            }));
+        }
+
+        return res;
+    }
+
+    public void execute(final Callback<SecurityCheckResult> callback) {
```
wei-lee: is this method really needed?

ziccardi: I added this in case the user wants to receive the results as they get produced instead of waiting on the future. We can of course remove it, but I think it can be useful, especially for UI.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
wei-lee: why it is overkill?

ziccardi:
@wei-lee Yes, but I think it is better to give the user the chance to change the executor as well. So, if you agree, I could use the AppExecutor as default, unless the user specifies a custom executor.

WDYT?

wei-lee:
@ziccardi can you think any reason why they may want to change the executor at all? Also I think if that is the case, that should be configured for the whole lib (thus it should be the job of the core module). I don't think each individual module should allow that.

ziccardi:
@wei-lee Sorry I read too fast. You want to get rid of the AsyncExecutor class. 
Well, I think sync executor and async executor are two different objects, so I prefer to model them on two different classes, both inheriting from the same base class.

ziccardi:
@wei-lee there are three main reasons I think it could be useful:

1. You can never know what is going to happen in the future
2. You can never know what the user will need
3. The user could implement a custom SecurityCheck that needs very long time, so he wants to use a custom executor to avoid blocking other checks.

matzew:
>  I think sync executor and async executor are two different objects

I'd agree here, that decoupling them does make sense, instead of smacking new methods on different concerns to the orignal clazz

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
wei-lee: I can see your point, given how quickly those tests will run, I honest don't think this is necessary. Plus it actually make things a little bit tricker for developers to consume: we need to tell developers that this callback will be invoked every time when a test is finished, and they can't control the order, and they can't know when all the tests finished running easily.

If we are going to add callback for each test, at least add another callback when all the tests finished running.

But as I said, I think this is overkill and over-engineered.

matzew: > we need to tell developers that this callback will be invoked every time when a test is finished

I actually like this, and pretty much matches also the world of reactive programming, for each 'item' an onNext() is invoked

> If we are going to add callback for each test, at least add another callback when all the tests finished running.

we could do something like "onComplete", similar to the pattern above 


> I think this is overkill and over-engineered.

do you mean the async handling? or the new class structure?

auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
```diff
@@ -39,12 +38,13 @@ public void configure(final MobileCore core, final ServiceConfiguration serviceC
...
 
     /**
-     * Retrieve a {@link SecurityCheckExecutor} to run multiple {@link SecurityCheckType checks} chained.
+     * Retrieve a {@link SyncSecurityCheckExecutor} to run multiple {@link SecurityCheckType checks} chained.
      *
      * @return A new executor.
      */
-    public SecurityCheckExecutor getCheckExecutor() {
-        return new SecurityCheckExecutorImpl(core.getContext());
+    public SyncSecurityCheckExecutor getCheckExecutor() {
```
aidenkeating: Where does the async executor get used?

ziccardi: Just added the method to return it ;)

auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
```diff
@@ -39,12 +38,23 @@ public void configure(final MobileCore core, final ServiceConfiguration serviceC
...
+        return SecurityCheckExecutor.Builder
+            .newSyncExecutor(core.getContext()).build();
+    }
+
+    /**
+     * Retrieve a {@link AsyncSecurityCheckExecutor} to asynchronously run multiple {@link SecurityCheckType checks} chained.
+     *
+     * @return A new async executor.
+     */
+    public AsyncSecurityCheckExecutor getCheckExecutorAsync() {
```
wei-lee: should it be `getAsyncCheckExecutor`?

core/src/main/java/org/aerogear/mobile/core/executor/AppExecutors.java
```diff
@@ -35,5 +38,6 @@ public Executor networkThread() {
         return networkExecutor;
     }
 
+    public ExecutorService service() { return serviceThreadExecutor; }
```
wei-lee: can we name it to `singleThreadService` to make it more obvious?

PR #88, opened 2018-02-23
Update Check Authenticated Example

TommyJ1994:
## Motivation
Small update to the docs to prevent an Incompatible types error in Android Studio.


TommyJ1994:
@aidenkeating Small one if you can take a look?

PR #90, opened 2018-02-23
Added back the 'addCheck' methods/Changed return value from array to Map

ziccardi:
## Motivation

Added back the 'addCheck' method that was erroneously removed
Changed the return value of the check executors: now a Map is returned (key: check name, value check result).

## Description

Re-added the addCheckMethods

## Progress
- [x] restore the addCheck method
- [x] change the return value of the executors to Map
- [x] update javadocs
- [x] update unit tests

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -11,6 +11,8 @@ import org.aerogear.mobile.core.metrics.MetricsService;
 import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
 
 import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 
@@ -64,24 +66,26 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
 
     /**
      * Executes the checks asynchronously and returns an array of {@link Future}
-     * @return an array of {@link Future} representing executed checks
+     *
+     * Returns a {@link Map} containing the results of each executed test (a {@link Future}).
+     * The key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
+     * a {@link Future} with the result of the check.
+     *
+     * @return a {@link Map} containing the results of all the executed checks
      */
-    public Future<SecurityCheckResult>[] execute() {
+    public Map<String, Future<SecurityCheckResult>> execute() {
 
-        final Collection<SecurityCheck> checks = getChecks();
         final MetricsService metricsService = getMetricsService();
+        final Map<String, Future<SecurityCheckResult>> res = new HashMap<>();
 
-        final Future[] res = new Future[checks.size()];
-
-        int i = 0;
-        for (final SecurityCheck check : checks) {
-            res[i++] = (executorService.submit(() -> {
+        for (final SecurityCheck check : getChecks()) {
+            res.put(check.getName(), (executorService.submit(() -> {
                 final SecurityCheckResult result =  check.test(getContext());
                 if (metricsService != null) {
                     metricsService.publish(new SecurityCheckResultMetric(result));
                 }
                 return result;
-            }));
+            })));
         }
 
         return res;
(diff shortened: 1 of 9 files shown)


ziccardi:
Since with this change the NAME of the check becomes very important, I would suggest using `getClass().getName` as name.
This way we will be sure that each check has a **unique** name.
Moreover, in the docs, we should outline to the user that he should use the same technique when creating a custom security check.
@wei-lee @aidenkeating @matzew WDYT?

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -64,24 +66,26 @@ public AsyncSecurityCheckExecutor build() {
...
-     * @return an array of {@link Future} representing executed checks
+     *
+     * Returns a {@link Map} containing the results of each executed test (a {@link Future}).
+     * The key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
+     * a {@link Future} with the result of the check.
+     *
+     * @return a {@link Map} containing the results of all the executed checks
      */
-    public Future<SecurityCheckResult>[] execute() {
+    public Map<String, Future<SecurityCheckResult>> execute() {
```
matzew: :heart_eyes_cat:

auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
```diff
@@ -39,40 +41,44 @@ public SyncSecurityCheckExecutor build() {
...
-     * Return the {@link SecurityCheckResult results} for each of the tests added.
+     * Returns a {@link Map} containing the results of each executed test.
+     * The key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
+     * the result of the check.
      *
-     * @return Array of results.
+     * @return a {@link Map} containing the results of all the executed checks
      */
-    private SecurityCheckResult[] getTestResults() {
+    private Map<String, SecurityCheckResult> getTestResults() {
```
matzew: I don't like the `getTestResults` name - it returns a collection of _checks_ - not tests (as in (software) testing) 

How about `readSecurityChecks()` ?

auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
```diff
@@ -39,40 +41,44 @@ public SyncSecurityCheckExecutor build() {
...
 
-        Collection<SecurityCheck> checks = getChecks();
+        final Map<String, SecurityCheckResult> results = new HashMap<>();
 
-        SecurityCheckResult[] results = new SecurityCheckResult[checks.size()];
-        int i = 0;
-        for (SecurityCheck check : checks) {
-            results[i++] = check.test(getContext());
+        for (SecurityCheck check : getChecks()) {
+            results.put(check.getName(), check.test(getContext()));
```
matzew: does a `SecurityCheck` have a name? or is the (simple)clazz name efficent enough ?

ziccardi: @matzew The issue here is that the user must have an easy way to map the received result to the check that produced it. In my previous comment I suggest that the check name is the 'class name' to avoid name clash when the user implements a custom check.

auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
ziccardi: agree

auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
```diff
@@ -48,11 +48,18 @@ public SyncSecurityCheckExecutor build() {
...
      */
     public Map<String, SecurityCheckResult> execute() {
-        final Map<String, SecurityCheckResult> results = getTestResults();
-        if (getMetricsService() != null) {
-            publishResultMetrics(results.values());
+        final Map<String, SecurityCheckResult> results = new HashMap<>();
+
+        for (SecurityCheck check : getChecks()) {
+            SecurityCheckResult result = check.test(getContext());
+            results.put(check.getName(), result);
```
matzew: even better this way, instead of the other method

wei-lee:
> Since with this change the NAME of the check becomes very important, I would suggest using getClass().getName as name.
This way we will be sure that each check has a unique name.

agreed. Let's do that.

matzew:
> I am happy to merge this PR and you can create another to change the name. Or you can do it as part of this PR. 

+1 for a new PR on this!

PR #94, opened 2018-02-26
Changed getname implementation

ziccardi:
## Motivation

Now that the return value for the security check executor is a `Map<String, SecurityCheckResult>` object instead of a simple array, the value returned as NAME by each check is very important to allow the caller to map the result with the check that generated it.
Moreover, that NAME must be unique to avoid NAME clash when populating the map with the results.

## Description

With this PR a new `getName` method with default implementation has been added to the `SecurityCheck` interface. That method returns, by default, the result of `getClass().getName()`.

## Progress

- [x] Add the new getName method
- [x] Update existing checks where needed
- [x] Update unit tests
- [x] Update javadocs


danielpassos:
@ziccardi should we really have a _default_ implementation for `SecurityCheck` name? It means now the `RootCheck` is going to send:

```
{"org.aerogear.mobile.security.checks.RootedCheck":{"passed":"false"}}
```

instead of:

```
{"detectRooted":{"passed":"false"}}
```

/cc @david-martin

ziccardi:
@danielpassos Yes, that is exactly what I wanted to achieve. Now that the 'name' is the key to the resulting map, we need it to be unique, even for custom checks implemented by the user.

danielpassos:
Hrm, I really don't like to have the identity with the name of the class.

PR #107, opened 2018-02-27
backports module initial

pb82:
## Motivation

WIP: don't merge

Our Cordova and react-native SDKs require native code for some use cases. Wrapping the Android SDK has proven to be difficult because it's using Java8 features and has rather high minimum SDK requirements (too high for react-native).

The idea here is to provide a `backport` module that does not depend on any other SDK modules. It's only purpose is to provide native implementations that can be consumed by Cordova and react-native.

Not sure if this should even be a module in the native SKDs or standalone.

I've tested this with react-native and will also test it with a Cordova plugin. There are still open questions about the structure of our Cordova/react-native/Web SDK.

Let me know what you think.


secondsun:
Right now I am at a big -1.

From my understanding this module isn't meant to be used by native developers and therefore shouldn't be in the SDK.  

Furthermore, native modules for javascript platforms should probably be kept together in a repository for those platforms.  For example last summer we produces a UPS module for react native here : https://github.com/emilioicai/react-native-unified-push.  As you can see it has an ios and android library.  Putting related native plugins for js SDKs together makes a lot more sense than sprinkling them among native SDKs.

wei-lee:
agree with @secondsun. It's better to create another repo host the the code that is required to get the android/ios SDK work with react-native. It can then consume the android & ios SDKs as dependencies.

pb82:
@secondsun I'm ok with moving this into it's own repository. About bundling this into the react native library: the idea is that this code can be used from react native and Cordova. That's why it should not be part of the react-native/Cordova libs.

wtrocki:
@pb82 Current proposal is to write as much JS as possible and use just native helpers - do not wrap entire SDK's etc - just exact cases where native code is needed. When we go this way we will most likely use existing native plugins or even  hold native plugins in the actual js-sdk repo. I think it's best to move this conversation to the list. 

I have started proposal by suggesting full native aproach, but after moving to the list this aproach was opposed by some team members - with really valid points. So then we moved to aproach where we write more JavaScript code, basing on the challenges required with full native aproach and problems with code reusability.
Everything is on the list and in proposal - including problems and different ideas. 

@david-martin Done really great job on summarizing that in the last email.

pb82:
@wtrocki @paolobueno @secondsun @wei-lee i've summarized the current concerns in a google groups post: https://groups.google.com/forum/#!topic/aerogear/-C1eMiOLnfQ

secondsun:
I think that we've decided we aren't doing backports per https://groups.google.com/forum/#!topic/aerogear/-C1eMiOLnfQ so I am closing this PR

PR #111, opened 2018-02-28
Example app crashing fix.

ciaranRoche:
## Motivation

The example app was crashing when the self defense checks where loaded.

## Description

The crash was caused by a cast in the SecurityService class. We where trying to cast SecurityCheckResult to a SecurityCheck, by removing the cast and returning a SecurityCheckResult we stop the example app from crashing

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
@@ -58,18 +58,18 @@ public class SecurityService implements ServiceModule{
     }
 
     /**
-     * Perform a single {@link SecurityCheckType} and get the {@link SecurityCheckResult result} for it.
+     * Used with enumeration to perform a single {@link SecurityCheckType} and get the {@link SecurityCheckResult result} for it.
      *
      * @param securityCheckType The type of check to execute
      * @return {@link SecurityCheckResult}
      * @throws IllegalArgumentException if {@param securityCheckType} is null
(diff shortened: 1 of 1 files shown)


auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
```diff
@@ -65,7 +65,7 @@ public AsyncSecurityCheckExecutor getAsyncCheckExecutor() {
      * @throws IllegalArgumentException if {@param securityCheckType} is null
      */
     public SecurityCheckResult check(@NonNull final SecurityCheckType securityCheckType) {
-        return (SecurityCheckResult) nonNull(securityCheckType, "securityCheckType").getSecurityCheck();
+        return check(nonNull(securityCheckType, "securityCheckType").getSecurityCheck());
```
ziccardi: I don't think this works.
This should be 
`return check(nonNull(securityCheckType, securityCheckType").getSecurityCheck()).test(context);`

aidenkeating: @ziccardi Why call test here instead of providing a SecurityCheck to the other check method which calls test anyway? Are we not duplicating the test call if we do this?

ziccardi: @aidenkeating @ciaranRoche Forget my comment. I misread the code. Sorry.

PR #112, opened 2018-02-28
Added base class to remove duplicate code

ziccardi:
## Motivation

Each security check marks the `context` as nonnull and sanity checks are duplicated in every security check implementation.
Moreover, each security check implementation is polluted with code to build the `SecurityCheckResult` object just to return a boolean.

## Description

Created a base class that delegates to subclasses the implementation of the check by simply implementing the `execute` method and just returning a `boolean`.

The `execute` method is guaranteed to receive a nonnull context, so no check need to be performed there.

The `SecurityResult` object is built in the base class automatically, so no need to worry about that in the real check implementation

## Progress

- [x] Add AbstractSecurityCheck class
- [x] Ensure tests still works

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/checks/AbstractSecurityCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/checks/AbstractSecurityCheck.java
@@ -1,0 +1,34 @@
+package org.aerogear.mobile.security.checks;
+
+import android.content.Context;
+import android.support.annotation.NonNull;
+
+import org.aerogear.mobile.security.SecurityCheck;
+import org.aerogear.mobile.security.SecurityCheckResult;
+import org.aerogear.mobile.security.impl.SecurityCheckResultImpl;
+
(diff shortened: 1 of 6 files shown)


auth/src/main/java/org/aerogear/mobile/security/checks/AbstractSecurityCheck.java
```diff
@@ -0,0 +1,34 @@
...
+import org.aerogear.mobile.security.impl.SecurityCheckResultImpl;
+
+import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
+
+/**
+ * Base class for security checks.
+ */
+public abstract class AbstractSecurityCheck implements SecurityCheck {
+    /**
+     * Check whether a debugger is attached to the current application.
```
wei-lee: is this comment right?

ziccardi: Ops... Copy and paste error! Thanks!

PR #113, opened 2018-02-28
Split metrics to user and contributor guide

wtrocki:
## Motivation

Adjust documentation to suggestion made by @david-martin in https://github.com/aerogear/aerogear-ios-sdk/pull/21


wtrocki:
Leaving if someone else want to comment and going to merge that later.

PR #114, opened 2018-02-28
AEROGEAR-2208 Add allow backup flag check

aidenkeating:
## Motivation

Add a check for `allowBackup` flag.

JIRA: https://issues.jboss.org/browse/AEROGEAR-2208

## Description

Checks whether the `allowBackup` flag is enabled.

Also, addressing errors from linting.
## Progress

- [x] Check
- [x] Tests
- [x] JavaDoc

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/checks/AllowBackupFlagCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/checks/AllowBackupFlagCheck.java
@@ -1,0 +1,32 @@
+package org.aerogear.mobile.security.checks;
+
+import android.content.Context;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager.NameNotFoundException;
+import android.support.annotation.NonNull;
+
+/**
+ * Check to determine whether the allowBackup flag is enabled for the application.
+ * The allowBackup flag determines whether to allow the application to participate in the backup
+ * and restore infrastructure.
+ */
+public class AllowBackupFlagCheck extends AbstractSecurityCheck{
+    /**
+     * Check whether the allowBackup flag is enabled.
+     *
+     * @param context Context to be used by the check.
+     * @return true if allowBackup is enabled.
+     * @throws IllegalStateException Will be thrown if package information can not be found.
+     */
+    @Override
+    public boolean execute(@NonNull final Context context) {
+        PackageInfo packageInfo;
+        try {
+            packageInfo = context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
+        } catch (NameNotFoundException e) {
+            throw new IllegalStateException("Could not retrieve package information from provided context", e);
+        }
+        return (packageInfo.applicationInfo.flags & ApplicationInfo.FLAG_ALLOW_BACKUP) != 0;
+    }
+}
--- a/auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
@@ -4,12 +4,6 @@ import android.content.Context;
 import android.provider.Settings;
 import android.support.annotation.NonNull;
 
-import org.aerogear.mobile.security.SecurityCheck;
-import org.aerogear.mobile.security.SecurityCheckResult;
-import org.aerogear.mobile.security.impl.SecurityCheckResultImpl;
-
-import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
-
 
 /**
  * Security check that detects if developer mode is enabled in the device.
--- a/auth/src/main/java/org/aerogear/mobile/security/checks/RootedCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/checks/RootedCheck.java
@@ -5,12 +5,6 @@ import android.support.annotation.NonNull;
 
 import com.scottyab.rootbeer.RootBeer;
 
-import org.aerogear.mobile.security.SecurityCheck;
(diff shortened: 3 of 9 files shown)


aidenkeating:
@ziccardi @secondsun @wei-lee @ciaranRoche Mind taking a look?

auth/src/main/java/org/aerogear/mobile/security/util/BuildReader.java
```diff
@@ -0,0 +1,80 @@
+package org.aerogear.mobile.security.util;
+
+import android.os.Build;
+
+/**
+ * Class to allow simpler testing of classes that use Build.
+ */
+public class BuildReader {
```
wei-lee: is this class only used for testing? If this is the case should it be moved to the test package?

aidenkeating: This isn't meant to be here! 😅

aidenkeating:
Oops, still need to add the check to the example app

auth/src/main/java/org/aerogear/mobile/security/checks/AllowBackupFlagCheck.java
```diff
@@ -0,0 +1,34 @@
...
+public class AllowBackupFlagCheck extends AbstractSecurityCheck{
+    /**
+     * Check whether the allowBackup flag is enabled.
+     *
+     * @param context Context to be used by the check.
+     * @return true if allowBackup is enabled.
+     */
+    @Override
+    public boolean execute(@NonNull final Context context) {
+        SanityCheck.nonNull(context, "context");
```
ziccardi: Remove this. Context is guaranteed to be non null

auth/src/main/java/org/aerogear/mobile/security/checks/AllowBackupFlagCheck.java
```diff
@@ -0,0 +1,34 @@
...
+        PackageInfo packageInfo;
+        try {
+            packageInfo = context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
+        } catch (NameNotFoundException e) {
+            throw new IllegalStateException("Could not retrieve package information from provided context", e);
```
ziccardi: Add javadocs for this

auth/src/main/java/org/aerogear/mobile/security/checks/DeveloperModeCheck.java
```diff
@@ -21,6 +15,7 @@
      *
      * @param context Context to be used by the check.
      * @return <code>true</code> if the developer options have been enabled on the device.
+     * @throws IllegalArgumentException if {@code context} is null
```
ziccardi: Remove this: it is wrong. Context is guaranteed to be nonnull here

auth/src/main/java/org/aerogear/mobile/security/checks/RootedCheck.java
```diff
@@ -21,6 +15,7 @@
      *
      * @param context Context to be used by the check.
      * @return <code>true</code> if the device is rooted.
+     * @throws IllegalArgumentException if {@code context} is null
```
ziccardi: Remove this. Context is guaranteed to be non null here.

auth/src/main/java/org/aerogear/mobile/security/checks/ScreenLockCheck.java
```diff
@@ -22,6 +15,7 @@
      *
      * @param context Context to be used by the check.
      * @return <code>true</code> if the device has a screen lock enabled.
+     * @throws IllegalArgumentException if {@code context} is null
```
ziccardi: Remove this. Context is guaranteed to be non null here.

danielpassos:
Not directly related to this PR, but testing this I was wondering if `securityService.checkAndSendMetric()` and `securityService.check()` should not accept more than one check at the same time. 

If I need to check and send the metrics for 5 different checks it will fire 5 request :pensive:

wei-lee:
@danielpassos  see https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java#L45-L55. You can use executors to chain the checks together.

ziccardi:
@danielpassos TBH I was thinking if it is really useful to have those checkXXX method, since the same can be obtained with the getCheckExecutor and getAsyncCheckExecutor methods with no added complexity.

danielpassos:
@wei-lee Yeap, but it still doing 1 request per check, what is "the" problem. 

* [SyncSecurityCheckExecutor.java#L56](https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java#L56)
* [AsyncSecurityCheckExecutor.java#L85](https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java#L85)

ziccardi:
@danielpassos @wei-lee That's something that can be addressed (very easily on the SyncSecurityCheckExecutor class, a bit harder on the AsyncSecurityCheckExecutor). If you believe it is something that must be done, please create a Jira.

secondsun:
@aidenkeating  Can you rebase on master and fix the merge conflicts?

wei-lee:
@danielpassos The way I see it is that the request to metrics module should be quick and light weight. The consumer should be allowed to call `metrics.publish` as many as it likes, after all, it's about metrics and there could be a lot of metrics. So I think it should be the metrics module's job to maybe batch the data and decide when to send the metrics. I don't think it should set restrictions on how to use the module.

aidenkeating:
@secondsun Mind taking a look?

danielpassos:
@aidenkeating 👍

PR #77, opened 2018-02-22
AEROGEAR-1979 - Self Defence Check API Docs

rachael-oregan:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-1979

## Description

API docs for the self defence check feature

PR #116, opened 2018-02-28
Add metrics executor example

paolobueno:
## Motivation

Current example in `SecurityServiceFragment` doesn't utilize the Executors and their integration with the metrics service.

This could be relevant to match the documentation being written in #77 and also for https://issues.jboss.org/browse/AGDROID-742

## Description

- Run examples in `SecurityServiceFragment` via an Executor instead of individually.
- Use the `Builder` interface to hook up the metrics service

## Progress

- [x] Use executor
- [x] Hook up metrics
- [x] Update `detect*()` examples to grab results from result map

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -32,7 +32,7 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
 
         private ExecutorService executorService;
 
-        public Builder(final Context ctx) {
+        Builder(final Context ctx) {
             super(ctx);
         }
 
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -32,7 +32,7 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
 
         private ExecutorService executorService;
 
-        Builder(final Context ctx) {
+        public Builder(final Context ctx) {
             super(ctx);
         }
 
--- a/auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
@@ -21,7 +21,7 @@ import java.util.Map;
 public class SyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<SyncSecurityCheckExecutor> {
 
     public static class Builder extends SecurityCheckExecutor.Builder.AbstractBuilder<Builder, SyncSecurityCheckExecutor> {
-        public Builder(final Context ctx) {
+        Builder(final Context ctx) {
             super(ctx);
         }
 
--- a/auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
@@ -21,7 +21,7 @@ import java.util.Map;
 public class SyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<SyncSecurityCheckExecutor> {
 
     public static class Builder extends SecurityCheckExecutor.Builder.AbstractBuilder<Builder, SyncSecurityCheckExecutor> {
-        Builder(final Context ctx) {
+        public Builder(final Context ctx) {
             super(ctx);
         }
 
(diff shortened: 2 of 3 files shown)


auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -32,7 +32,7 @@
 
         private ExecutorService executorService;
 
-        Builder(final Context ctx) {
+        public Builder(final Context ctx) {
```
ziccardi: This is not meant to be public

auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
```diff
@@ -21,7 +21,7 @@
 public class SyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<SyncSecurityCheckExecutor> {
 
     public static class Builder extends SecurityCheckExecutor.Builder.AbstractBuilder<Builder, SyncSecurityCheckExecutor> {
-        Builder(final Context ctx) {
+        public Builder(final Context ctx) {
```
ziccardi: This is not meant to be public

paolobueno: I did this because this useful chaining interface wasn't exposed.

Currently https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java#L45 allows supplying only the `Context` and there's no way to give it the metrics service.
We could improve the `SecurityService` interface to allow supplying the other builder params instead, but that defeats the purpose of having a builder.

example/src/main/java/org/aerogear/mobile/example/ui/SecurityServiceFragment.java
```diff
@@ -81,15 +86,25 @@ public void onActivityCreated(@Nullable final Bundle savedInstanceState) {
      * Executes all test and calls setTrustScore() to calculate an average score
      */
     public void runTests() {
+        SyncSecurityCheckExecutor executor = (new SyncSecurityCheckExecutor.Builder(this.getContext()))
```
ziccardi: Change this to 
`SecurityCheckExecutor.newSyncExecutor(this.getContext())`

auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
ziccardi: It is meant to be retrieved from the SecurityCheckExecutor class through the `newSyncExecutor` and `newAsyncExecutor` methods.

example/src/main/java/org/aerogear/mobile/example/ui/SecurityServiceFragment.java
paolobueno: @ziccardi Is there a way to provide a reference to the metrics service this way?

See https://github.com/aerogear/aerogear-android-sdk/pull/116#discussion_r171285424.

auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
ziccardi: An Example usage can be found here https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/test/java/org/aerogear/mobile/security/SecurityCheckExecutorTest.java#L53

example/src/main/java/org/aerogear/mobile/example/ui/SecurityServiceFragment.java
ziccardi: @paolobueno yes See here https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/test/java/org/aerogear/mobile/security/SecurityCheckExecutorTest.java#L53

paolobueno: Got it, changing. Any reason for the extra method hop though?

example/src/main/java/org/aerogear/mobile/example/ui/SecurityServiceFragment.java
```diff
@@ -81,15 +87,26 @@ public void onActivityCreated(@Nullable final Bundle savedInstanceState) {
      * Executes all test and calls setTrustScore() to calculate an average score
      */
     public void runTests() {
+
+        SyncSecurityCheckExecutor executor = SecurityCheckExecutor.Builder.newSyncExecutor(this.getContext())
```
ziccardi: The idea behind this builder structure is that you don really need to reference the Sync/Async executor. 
You can avoid saving the instance to a SyncSecurityCheckExecutor variable, and directly do:
`
Map<String, SecurityCheckResult> results = SecurityCheckExecutor.Builder.newSyncExecutor(this.getContext())
...
.build().execute();
`

paolobueno: I see, so avoiding the explicit import of `SyncSecurityCheckExecutor` :+1:

example/src/main/java/org/aerogear/mobile/example/ui/SecurityServiceFragment.java
ziccardi: Not really. Feel free to fix that.

Issue #96, opened 2018-02-26
Enhance Logout Functionality


TommyJ1994:
## Expected Behavior
When I attempt to logout I should have a callback on whether or not the logout was successful or not so I can handle the result how I want.

## Actual Behavior
The current `logout()` function is send and pray and doesn't currently use callbacks. The user might not be logged out on the keycloak server, but the authstate will be deleted. The keycloak session is still intact though. Its difficult to write an application and provide UI feedback when you always have to presume the logout succeeds.

## Steps to Reproduce the Problem

  1. `authService.logout(currentUser);`

## Specifications

  - Core SDK Version: 5.0.0.SNAPSHOT
  - Relevant Service Modules: auth

## Suggestion
Introduce callbacks to check if the keycloak server is contactable and if logout has succeeded or not. I think the deletion of the auth tokens should be split from the logout function. Maybe a `deleteAuthTokens()` or a similar function so a developer could use these for offline logouts or other use cases, rather than being baked into the logout function.

TommyJ1994:
Integrated in https://github.com/aerogear/aerogear-android-sdk/pull/118

PR #118, opened 2018-02-28
Improve Logout Functionality

TommyJ1994:
## Motivation
Detail in https://github.com/aerogear/aerogear-android-sdk/issues/96
Improves usage experience for - https://github.com/feedhenry/mobile-security-android-template/pull/36

## Description
Adding a callback for the logout function so it lets the user know if the logout has succeeded or not. 

## Progress
- [x] Add Logout Callbacks https://github.com/aerogear/aerogear-android-sdk/pull/118/commits/8b08cbe4201d0f65fd5dca69f420a55a9852b6a3
- [x] Update Example App to use Logout Callbacks https://github.com/aerogear/aerogear-android-sdk/pull/118/commits/8b08cbe4201d0f65fd5dca69f420a55a9852b6a3
- [x] Expose a `deleteTokens()` function to handle offline logouts or other. https://github.com/aerogear/aerogear-android-sdk/pull/118/commits/7bf77a429580c2553a9c197244985e0665328237
- [x] Update Docs https://github.com/aerogear/aerogear-android-sdk/pull/118/commits/bbbebe60de80b03fd2953ea46879833c38dcbd75
- [x] Create logout test. https://github.com/aerogear/aerogear-android-sdk/pull/118/commits/62ca6bbdad72aba2e530f56062adaad87993ae8e

## Additional Notes
Updated the Callback interface so that onSuccess can be called without the requirement to pass some data.

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/authenticator/AbstractAuthenticator.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/authenticator/AbstractAuthenticator.java
@@ -36,8 +36,9 @@ public abstract class AbstractAuthenticator {
      * Logout the given principal
      *
      * @param principal principal to be log out
+     * @param callback the callback function to be invoked
      */
-    public abstract void logout(final UserPrincipal principal);
+    public abstract void logout(final UserPrincipal principal, final Callback<UserPrincipal> callback);
 
     /**
      * Returns the authentication singleThreadService configuration
--- a/auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
@@ -173,15 +173,20 @@ public class OIDCAuthenticatorImpl extends AbstractAuthenticator {
 
         final HttpResponse httpResponse = request.execute();
 
-        httpResponse.onComplete(() -> {
+        httpResponse.onSuccess(() -> {
             if (httpResponse.getStatus() == HTTP_OK || httpResponse.getStatus() == HTTP_MOVED_TEMP) {
                 // delete the local tokens when the session with the OIDC has been terminated
                 authStateManager.save(null);
                 logoutCallback.onSuccess();
             } else {
-                MobileCore.getLogger().error("Error Performing a Logout on the Remote OIDC Server: ", httpResponse.getRequestError());
-                logoutCallback.onError(httpResponse.getRequestError());
+                // Non HTTP 200 or 302 Status Code Returned
+                Exception error = httpResponse.getError() != null ? httpResponse.getError() : new Exception("Non HTTP 200 or 302 Status Code.");
+                MobileCore.getLogger().error("Error Performing a Logout on the Remote OIDC Server: ", error);
+                logoutCallback.onError(error);
             }
+        }).onError(() -> {
+            MobileCore.getLogger().error("Error Performing a Logout on the Remote OIDC Server: ", httpResponse.getError());
+            logoutCallback.onError(httpResponse.getError());
         });
     }
 
--- a/auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
@@ -21,16 +21,21 @@ import org.aerogear.mobile.auth.credentials.OIDCCredentials;
 import org.aerogear.mobile.auth.user.UserPrincipal;
 import org.aerogear.mobile.auth.user.UserPrincipalImpl;
 import org.aerogear.mobile.auth.utils.UserIdentityParser;
+import org.aerogear.mobile.core.MobileCore;
 import org.aerogear.mobile.core.configuration.ServiceConfiguration;
 import org.aerogear.mobile.core.executor.AppExecutors;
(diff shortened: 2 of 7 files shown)


auth/src/main/java/org/aerogear/mobile/auth/authenticator/AbstractAuthenticator.java
```diff
@@ -36,8 +36,9 @@ public AbstractAuthenticator(@NonNull final ServiceConfiguration serviceConfig)
      * Logout the given principal
      *
      * @param principal principal to be log out
+     * @param callback the callback function to be invoked
      */
-    public abstract void logout(final UserPrincipal principal);
+    public abstract void logout(final UserPrincipal principal, final Callback<UserPrincipal> callback);
```
secondsun: I would prefer of instead of a call back you use a promise style API.  Callbacks are often written as anonymous inner classes in Activities which can leak references and cause crashes.

Ultimately we will probably want to have our exposed APIs wrap something like RxJava behind the scenes

wei-lee: @secondsun I would agree, but I think it should be consistent across all the modules. Perhaps something should be provided by the core module?

For now, let's use callback for this one as login method is already using callbacks.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
```diff
@@ -169,15 +171,22 @@ private void performLogout(final URL logoutUrl) {
...
-        authStateManager.save(null);
+        final HttpResponse httpResponse = request.execute();
+
+        httpResponse.onComplete(() -> {
+            if (httpResponse.getStatus() == HTTP_OK || httpResponse.getStatus() == HTTP_MOVED_TEMP) {
+                // delete the local tokens when the session with the OIDC has been terminated
+                authStateManager.save(null);
+                logoutCallback.onSuccess();
+            } else {
+                MobileCore.getLogger().error("Error Performing a Logout on the Remote OIDC Server: ", httpResponse.getRequestError());
```
wei-lee: should we fail the logout action even if we can't log out the current session? I thought the session will be expired eventually?

auth/src/main/java/org/aerogear/mobile/auth/authenticator/AbstractAuthenticator.java
TommyJ1994: @secondsun Sounds like a plan. Do you mind opening an issue in this repo to discuss a bit more widely? For this PR I am following the current Callback style that is used currently by other functions in the SDK.  I think it would be good to have some alignment of the best approach for this as it will affect the example/mobile-security apps too with how they do callbacks.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
TommyJ1994: I mentioned briefly about this in https://github.com/aerogear/aerogear-android-sdk/issues/96, but I think it is better to give the end developer more control or how they handle logouts when they can't contact the server. If we always the delete the local tokens when calling logout(), it will be impossible for a user to do a true logout against the server when they come online, as we have deleted the identity token. Simply deleting the tokens locally will still mean that all the tokens and the keycloak session are still active, we just don't have the tokens anymore (that doesn't stop someone else from using them though). With SSO in gerneral, the expiry times could be really long and once we delete the tokens locally, they are gone - but the session/tokens are still valid unless we notify the server.

I was planning on exposing a `deleteTokens()` function so the developer can choose how they handle online/offline logouts. All tokens/sessions will expire eventually, whether they click a logout button or not, but the use of callbacks plus exposing some token deletion methods can help them reduce the timelines to revoke sessions/tokens. 

On a minor side not, it *may* help reduce the number of concurrent sessions for a single user and have less of an impact on things like active session metrics.

Let me know which approach you think is best.

wei-lee: Yeah, I think you are right. Let the developers decide what they want to do.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/AbstractAuthenticator.java
secondsun: If you missed the issue further down : https://github.com/aerogear/aerogear-android-sdk/issues/122

auth/src/test/java/org/aerogear/mobile/auth/authenticator/OIDCAuthenticatorImplTest.java
```diff
@@ -205,4 +234,20 @@ public void onError(Throwable error) {
         authenticator.handleAuthResult(intent);
     }
 
+    @Test
+    public void testLogout() {
+        authenticator.logout(userPrincipalImpl, new Callback<UserPrincipal>() {
```
secondsun: This is an asynchronous method so the callback isn't being called.  Therefor this test isn't testing anything.  You can verify by adding a fails in the success block.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
secondsun: getRequestError is now just getError

TommyJ1994:
@secondsun Updated the test. Ok to merge?

PR #119, opened 2018-03-01
implement http error handling

pb82:
## Motivation

Allow better error handling when using the http module and fix a bug in the metrics module. When the return code from a request was 500 the `onComplete` handler was triggered. The metrics module tried to evaluate the `requestError` but it was null. The current implementation does set the `requestError` only when the request itself failed, not when the response returned an invalid status code.

Example:

```java
final HttpResponse resp = httpRequest.execute();
resp.onSuccess(() -> {
  // Do something with the result           
}).onError(() -> {
  // Print an error message, you can safely access resp.getRequestError()
}).onComplete() -> {
  // Close any resources, popups... This will always be called
});
```

## Description

Adds `onError` and `onSuccess` handlers to the `HttpResponse` objects. I've looked at the OkHttp recipes (https://github.com/square/okhttp/wiki/Recipes) and they seem to suggest a similar pattern for error handling.

Tested with:

- successful request
- request that returns status 500
- request to invalid URL

## Progress

- [x] Done Task

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
@@ -9,7 +9,9 @@ package org.aerogear.mobile.core.http;
 public interface HttpResponse {
 
     /**
-     * Creates a callback to be called when the response is finished successfully.
+     * Creates a callback to be called when the response is finished (successful or unsuccessful).
+     * This will be called, even if success and error handlers are registered. onComplete will
+     * be called last.
      *
      * The response is delivered immediately if the request has already been finished.
      *
@@ -28,6 +30,16 @@ public interface HttpResponse {
      * @return the instance of the HttpResponse for API chaining
      */
     HttpResponse onError(Runnable runnable);
+
+    /**
+     * Creates a callback to be called when the request has succeeded.
+     *
+     * This will be triggered when the request succeeded and received a non-error status code.
+     *
+     * @param runnable a callback method
+     * @return the instance of the HttpResponse for API chaining
+     */
+    HttpResponse onSuccess(Runnable runnable);
 
     /**
      * Returns the HTTP status code of the response.
@@ -50,16 +62,8 @@ public interface HttpResponse {
     String stringBody();
 
     /**
-     * Returns true if the requests failed. Use getStatus to identify
-     * the reason
-     *
-     * @return true if request failed
-     */
-    boolean requestFailed();
-
-    /**
      * Returns the request error if it failed
      * @return Exception request error or null
      */
-    Exception getRequestError();
+    Exception getError();
 }
--- a/core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
@@ -19,6 +19,17 @@ public interface HttpResponse {
     HttpResponse onComplete(Runnable runnable);
 
     /**
+     * Creates a callback to be called when the request has failed.
+     *
+     * This will be triggered when the request could not be sent or no response can be
+     * received.
+     *
+     * @param runnable a callback method
+     * @return the instance of the HttpResponse for API chaining
+     */
+    HttpResponse onError(Runnable runnable);
+
+    /**
      * Returns the HTTP status code of the response.
      *
      * @return the status code.
--- a/core/src/main/java/org/aerogear/mobile/core/http/InvalidResponseException.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/InvalidResponseException.java
@@ -1,0 +1,18 @@
+package org.aerogear.mobile.core.http;
+
+/**
+ * Custom exception thrown when a http request returns an unenxpected
+ * result or result code
+ */
+public class InvalidResponseException extends RuntimeException {
+    private final int statusCode;
+
+    public InvalidResponseException(int statusCode) {
+        super("Unexpected response from server. Status code was: " + statusCode);
+        this.statusCode = statusCode;
+    }
+
+    public int getStatusCode() {
+        return statusCode;
+    }
+}
--- a/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
@@ -15,6 +15,7 @@ class OkHttpResponse implements HttpResponse {
     private final CountDownLatch requestCompleteLatch = new CountDownLatch(1);
     private Response response;
     private Runnable completionHandler;
+    private Runnable errorHandler;
     private Exception requestError;
     private boolean closed = false;
 
@@ -23,12 +24,16 @@ class OkHttpResponse implements HttpResponse {
             try {
                 response = okHttpCall.execute();
                 requestCompleteLatch.countDown();
-                if (completionHandler != null) {
+
+                if (!response.isSuccessful()) {
+                    throw new IOException("Request failed with status code " + response.code());
+                } else {
                     runCompletionHandler();
                 }
             } catch (IOException e) {
                 requestError = e;
                 requestCompleteLatch.countDown();
+                runErrorHandler();
             }
         });
     }
@@ -46,11 +51,27 @@ class OkHttpResponse implements HttpResponse {
         }
     }
 
+    private synchronized void runErrorHandler() {
+        if (errorHandler != null) {
+            errorHandler.run();
+            errorHandler = null;
+        }
+    }
+
     @Override
     public HttpResponse onComplete(Runnable completionHandler) {
         this.completionHandler = completionHandler;
-        if (response != null) {
+        if (response != null && response.isSuccessful()) {
             runCompletionHandler();
+        }
+        return this;
+    }
+
+    @Override
+    public HttpResponse onError(Runnable errorHandler) {
+        this.errorHandler = errorHandler;
+        if (response != null && !response.isSuccessful()) {
+            runErrorHandler();
         }
         return this;
     }
@@ -66,8 +87,10 @@ class OkHttpResponse implements HttpResponse {
             requestCompleteLatch.await(DEFAULT_TIMEOUT, TimeUnit.SECONDS);
             //If a completion Handler was set before this wait then we need to make sure that
             // gets called before we free these resources.
-            if (completionHandler != null) {
+            if (completionHandler != null && requestError == null) {
                 runCompletionHandler();
+            } else {
+                runErrorHandler();
             }
         } catch (InterruptedException interruptedException) {
             /*If we have an error the work of the thread is already done and the thread was exiting
--- a/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
@@ -16,7 +16,8 @@ class OkHttpResponse implements HttpResponse {
     private Response response;
     private Runnable completionHandler;
     private Runnable errorHandler;
-    private Exception requestError;
+    private Runnable successHandler;
+    private Exception error;
     private boolean closed = false;
 
     public OkHttpResponse(final Call okHttpCall, AppExecutors appExecutors) {
@@ -26,14 +27,16 @@ class OkHttpResponse implements HttpResponse {
                 requestCompleteLatch.countDown();
 
                 if (!response.isSuccessful()) {
-                    throw new IOException("Request failed with status code " + response.code());
+                    throw new InvalidResponseException(response.code());
                 } else {
-                    runCompletionHandler();
+                    runSuccessHandler();
                 }
-            } catch (IOException e) {
-                requestError = e;
+            } catch (IOException | InvalidResponseException e) {
+                error = e;
                 requestCompleteLatch.countDown();
                 runErrorHandler();
+            } finally {
+                runCompletionHandler();
             }
         });
     }
@@ -58,10 +61,17 @@ class OkHttpResponse implements HttpResponse {
         }
     }
 
+    private synchronized  void runSuccessHandler() {
+        if (successHandler != null) {
+            successHandler.run();
+            successHandler = null;
+        }
+    }
+
     @Override
     public HttpResponse onComplete(Runnable completionHandler) {
         this.completionHandler = completionHandler;
-        if (response != null && response.isSuccessful()) {
+        if (response != null) {
             runCompletionHandler();
         }
         return this;
@@ -77,6 +87,15 @@ class OkHttpResponse implements HttpResponse {
     }
 
     @Override
+    public HttpResponse onSuccess(Runnable successHandler) {
+        this.successHandler = successHandler;
+        if (response != null && response.isSuccessful()) {
+            runSuccessHandler();
+        }
+        return this;
+    }
+
+    @Override
     public int getStatus() {
         return response.code();
     }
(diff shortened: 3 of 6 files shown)


david-martin:
@pb82 Looks good. Do we use this pattern elsewhere?
Will leave approval to @danielpassos & @secondsun

pb82:
@david-martin it's the success/error callback pattern that's widely used in async languages. Not sure about Android, but since we already have `onComplete` i think it makes sense to also have `onError`.

The problem with `onComplete` was that you can't be sure that the `requestError` object was set and before using it you always had to check if it is null (If the server returned 500 the error was not set, only when the request failed completely).

Now when `onError` is called there will always be a `requestError` that you can inspect. In the case of a 500 it will look like:

```
E/AeroGear: Metrics request error java.io.IOException: Request failed with status code 500
```

There's another use in the auth module where the error handler also could be used: https://github.com/aerogear/aerogear-android-sdk/blob/master/auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java#L101

(But the old method also still works)

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -54,12 +54,10 @@ public void publish(final Metrics... metrics) {
...
             httpResponse.onComplete(() -> {
-                if (httpResponse.getStatus() == HTTP_OK) {
-                    MobileCore.getLogger().debug("Metrics sent: " + json.toString());
-                } else {
-                    MobileCore.getLogger().error(httpResponse.getRequestError().getMessage(),
-                        httpResponse.getRequestError());
-                }
+                MobileCore.getLogger().debug("Metrics sent: " + json.toString());
+            }).onError(() -> {
+                MobileCore.getLogger().error("Metrics request error",
```
pb82: @secondsun @danielpassos @wtrocki Should we log anything here? This will happen on devices out in the field, while the logs are only useful when plugged into Android Studio. Are you aware of any conventions about logging in Android apps?

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -54,12 +54,10 @@ public void publish(final Metrics... metrics) {
 
             final HttpResponse httpResponse = httpRequest.execute();
             httpResponse.onComplete(() -> {
-                if (httpResponse.getStatus() == HTTP_OK) {
-                    MobileCore.getLogger().debug("Metrics sent: " + json.toString());
-                } else {
-                    MobileCore.getLogger().error(httpResponse.getRequestError().getMessage(),
-                        httpResponse.getRequestError());
-                }
+                MobileCore.getLogger().debug("Metrics sent: " + json.toString());
```
wtrocki: Great to see this aligned with IOS!

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
wtrocki: Great to see this example aligned with IOS:
https://github.com/aerogear/aerogear-ios-sdk/blob/master/modules/core/metrics/engine/MetricsNetworkPublisher.swift#L20-L26

This can be ok for now. For the cases when network connection is down error will be printed but that's not a big problem IMHO.
I have no strong opinion about it, but would like to keep that aligned with IOS.

core/src/test/java/org/aerogear/mobile/core/http/OkHttpServiceModuleTest.java
```diff
@@ -43,4 +43,26 @@ public void testGet() {
         response.waitForCompletionAndClose();
     }
 
+    @Test
+    public void testErrorHandler() {
```
josemigallas: Please @pb82 rename this method accordingly with what the test case is about

pb82: But it is about testing the error handler?

core/src/test/java/org/aerogear/mobile/core/http/OkHttpServiceModuleTest.java
```diff
@@ -43,4 +43,26 @@ public void testGet() {
...
+        HttpServiceModule module = new OkHttpServiceModule();
+
+        HttpRequest request = module.newRequest();
+        request.get("http://does.not.exist.com");
+
+        final HttpResponse response = request.execute();
+        assertNotNull(response);
+
+        response.onComplete(() -> {
+            // The complete handler must not be called here
```
josemigallas: Call `Assert.fail();` instead of comment plus forced bad assertion

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
wtrocki: Actually this area will need more work anyway.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -46,15 +51,31 @@ private synchronized void runCompletionHandler() {
...
+        if (response != null && response.isSuccessful()) {
             runCompletionHandler();
         }
         return this;
     }
 
+    @Override
+    public HttpResponse onError(Runnable errorHandler) {
+        this.errorHandler = errorHandler;
+        if (response != null && !response.isSuccessful()) {
```
danielpassos: I'm not sure if we need this kind of check here. IMO we should check only if `errorHandler` is null to avoid NPE

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -66,8 +87,10 @@ public void waitForCompletionAndClose() {
             requestCompleteLatch.await(DEFAULT_TIMEOUT, TimeUnit.SECONDS);
             //If a completion Handler was set before this wait then we need to make sure that
             // gets called before we free these resources.
-            if (completionHandler != null) {
+            if (completionHandler != null && requestError == null) {
```
danielpassos: Like I said before if we have a `completionHandler` it should be always called, not only on success

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
pb82: This check is required because otherwise it could happen that you attach the errorHandler when the network thread has not yet finished and response is still null. In that case we can not yet check if the response was successful and have to wait for the network thread callback to handle that.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
pb82: Hmm, wouldn't that make the error handler pointless? Because then you would have to check for errors in your completion handler.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -46,15 +51,31 @@ private synchronized void runCompletionHandler() {
...
+            errorHandler.run();
+            errorHandler = null;
+        }
+    }
+
     @Override
     public HttpResponse onComplete(Runnable completionHandler) {
         this.completionHandler = completionHandler;
-        if (response != null) {
+        if (response != null && response.isSuccessful()) {
```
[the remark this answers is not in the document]
pb82: Ok, understand. Then you would see the completion handler more as soemthing like a `finally`? I'm ok with handling that in the completion handler but then we should probably remove the error handler because you would have to handle a possible error in your completion handler anyway.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
wei-lee: @danielpassos I was thinking about the same, but if you read the javadoc of the `onComplete` method:

> /**
     * Creates a callback to be called when the response is finished successfully.
     *
     * The response is delivered immediately if the request has already been finished.
     */

It says that it is only called when the response is finished successfully.  

If this is still the case, then I think it shouldn't be called when there is an error.

I think the name of the method is a bit confusing. I agree that if it is called `onComplete`, then it should be called all the time. Otherwise it's better rename it to `onSuccess`.

pb82: @wei-lee i'll update this PR soon to include `onSuccess`, `onError` and `onComplete`.

wei-lee: @pb82 👍

core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
```diff
@@ -18,6 +20,27 @@
...
+
+    /**
+     * Creates a callback to be called when the request has succeeded.
+     *
+     * This will be triggered when the request succeeded and received a non-error status code.
+     *
+     * @param runnable a callback method
+     * @return the instance of the HttpResponse for API chaining
+     */
+    HttpResponse onSuccess(Runnable runnable);
```
matzew: doesn't this pretty much make the `onComplete` redundant ? 

I'd assume it's no a "long-running" response, right ?

pb82: @matzew it's like `try`, `catch`, `finally`. onComplete can always be used to clean things up even when there's an error. And you would use onSuccess only if the request succeeded to do things that rely on the data.

danielpassos: @matzew Nopz, it's the same way RxJava handler stuffs

http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observer.html

matzew: @danielpassos actually, nope - `Flowable`/`Observable` are emitting `n` objects - while `Single` just emits one. See the `subscribe(Consumer<? super T> onSuccess, Consumer<? super Throwable> onError)` 

that's why I asked if these are long-running requests, where something like `onNext(ByteBuffer)` would make sense.  


--

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -23,12 +25,18 @@ public OkHttpResponse(final Call okHttpCall, AppExecutors appExecutors) {
             try {
                 response = okHttpCall.execute();
                 requestCompleteLatch.countDown();
-                if (completionHandler != null) {
-                    runCompletionHandler();
+
+                if (!response.isSuccessful()) {
+                    throw new IOException("Request failed with status code " + response.code());
```
danielpassos: I'm in doubt what we should do here. The _request_ was **success** called, but the response was not (range [200..300)). This should be a dev responsibility to handle that. IMO we should always call the success expect when the HTTP layer raises a problem. That is how retrofit handles it.

@secondsun Can you give you opinion here?

secondsun: So first a bit of bikeshedding, IOException is the wrong exception to use here.  IOException is more of a "the network quit working" type of failure.

I think that success and error handlers make a lot more sense when we start thinking about ways for modules to "rescue" a http request.  If auth can refresh a token then it should be able to register as an error handler and then this distinction makes a lot of sense.

However this kind of changes the semantics a bit.  A request is made.  A response is returned.  success and error handlers are called (which may or may not generate other network activity as appropriate) and when the whole mess has settled the completion handler has finally been called.  Completion therefore stops being about a single HTTP request and is instead about a user's action in the application.

Sorry I don't think I made anything more clear LOL

pb82:
@secondsun I agree with you that `IOException` is the wrong type for a invalid response even though the OkHttp guys seem to bee happy with it: https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/SynchronousGet.java#L33

I've changed that to a custom Exception.

@secondsun @matzew @wei-lee  I was discussing this change with @danielpassos and there is one open question about the onSuccess/onError handlers:

Should a response with an error code (e.g. 500) be treated as an error (onError) or treated as a successful request & response (onSuccess) and we only call onError for Network errors?

I'd prefer to call `onError` for a 500 response but i'm happy to change it if you think that is wrong.

What's your Opinion on this?

wei-lee:
@pb82 I agree with you and I think it should call `onError` for 5xx responses. In this case, there should be a response object and the developer can use the status code to determine the type of the error.

core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
secondsun: Right now this API doesn't support that type of request.  When the request gets done things get rolled into memory for better or worse.

core/src/main/java/org/aerogear/mobile/core/http/HttpResponse.java
```diff
@@ -38,17 +61,9 @@
...
-     * @return true if request failed
-     */
-    boolean requestFailed();
-
     /**
      * Returns the request error if it failed
      * @return Exception request error or null
      */
-    Exception getRequestError();
+    Exception getHttpError();
```
secondsun: How about just getError?

core/src/main/java/org/aerogear/mobile/core/http/InvalidResponseException.java
```diff
@@ -0,0 +1,18 @@
+package org.aerogear.mobile.core.http;
+
+/**
+ * Custom exception thrown when a http request returns an unenxpected
+ * result or result code
+ */
+public class InvalidResponseException extends Exception {
```
secondsun: I would prefer extending runtimeexception.

TommyJ1994:
@pb82 Are HTTP 302 Redirects considered errors in a response? I've updated my PR to use the changes here, but it's causing issues with Keycloak, specifically on Keycloak logouts where the result is a HTTP 302 for a successful logout.

```
E/AeroGear: Error Performing a Logout on the Remote OIDC Server: 
            org.aerogear.mobile.core.http.InvalidResponseException: Unexpected response from server. Status code was: 302
                at org.aerogear.mobile.core.http.OkHttpResponse.lambda$new$0$OkHttpResponse(OkHttpResponse.java:30)
```

pb82:
@TommyJ1994 Apparently yes: https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Response.java#L106

pb82:
@TommyJ1994 There are three possible ways to resolve this:

1. we provide our own `isSuccesful` that includes redirects
2. we go with @danielpassos approach and consider every response a success and only network errors as failures
3. we add a `onRedirect` handler. Probably not a good idea.

secondsun:
I vote for 2

TommyJ1994:
2 for me too for flexibility.

pb82:
@secondsun @TommyJ1994 Ok, preparing a PR.

TommyJ1994:
@pb82 Created an issue for this so you can have a reference for the PR. https://github.com/aerogear/aerogear-android-sdk/issues/123

danielpassos:
2 is exactly what Okhttp do, they do this because there is a difference between request success and response success. We should handle only request success because there is the lib responsibility. If we could fire the request without any problem we should call success, if not we should call error. Handle the response error is the dev responsibility.

PR #120, opened 2018-03-01
Adjust default network timeouts

wtrocki:
## Motivation

Adjust default network timeouts.
Issue logged by @TommyJ1994

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -21,6 +21,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import okhttp3.OkHttpClient;
 
 import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -65,7 +68,11 @@ private MobileCore(final Context context, final Options options) throws Initiali
 
         // -- Setting default http layer
         if (options.httpServiceModule == null) {
-            final OkHttpServiceModule httpServiceModule = new OkHttpServiceModule();
+            OkHttpClient.Builder builder = new OkHttpClient.Builder();
+            builder.connectTimeout(10, TimeUnit.SECONDS)
+                .writeTimeout(10, TimeUnit.SECONDS)
+                .readTimeout(30, TimeUnit.SECONDS);
```
wei-lee: magic numbers! 

Also I think these value should be configurable by the developers, but it probably can be done in a different PR.

wtrocki: if (options.httpServiceModule == null)

wei-lee: Personally I think it's not very good user experience to ask the developers to provide a whole new http implementation just to change the default timeout values.  Considering there will be more features will be added in the future to the http module (like automatically do certificate pinning for AG services), I think it's better to give some configuration options to the developers so that they can customise some aspects of the http module. 

But my main concern with this PR is the usage of magic numbers. The configuration can be dealt with in a separate PR.

danielpassos: I agree with @wei-lee but we can create a new Jira/PR for it

danielpassos: I agree with @wei-lee but we can create a new Jira/PR for it

PR #126, opened 2018-03-02
[AGDROID-771] Batch security metrics and update json schema

paolobueno:
## Motivation

JIRA: https://issues.jboss.org/browse/AGDROID-771

## Description

Make both Sync and Async executors send a single metrics request upon finishing execution.
Update data being sent to match the server-side implementation at https://github.com/aerogear/aerogear-app-metrics/pull/33

## Progress

- [x] Add support for multiple results
- [x] Update call from sync executor
- [x] Update call from async executor
- [x] Update unit tests
- [x] Add `type` field
- [x] Change syncing mechanism to `syncroninzed` keyword
- [x] Make metrics interface generic

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -23,7 +23,10 @@ android {
     }
 
     testOptions {
-        unitTests.includeAndroidResources = true
+        unitTests {
+            includeAndroidResources = true
+            returnDefaultValues = true
+        }
     }
 }
 
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -4,9 +4,7 @@ import android.content.Context;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
-import org.aerogear.mobile.core.MobileCore;
 import org.aerogear.mobile.core.executor.AppExecutors;
-import org.aerogear.mobile.core.logging.Logger;
 import org.aerogear.mobile.core.metrics.MetricsService;
 
 import java.util.ArrayList;
@@ -132,7 +130,7 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
                     listener.onExecuted(result);
 
                     if (remaining <= 0) {
-                        listener.onFinished();
+                        listener.onComplete();
                     }
                 }
 
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -116,18 +116,18 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
         listeners.add(getMetricServicePublisher());
 
         final Map<String, Future<SecurityCheckResult>> res = new HashMap<>();
-        AtomicInteger count = new AtomicInteger(checks.size());
+        final AtomicInteger count = new AtomicInteger(checks.size());
 
         for (final SecurityCheck check : getChecks()) {
             res.put(check.getName(), (executorService.submit(() -> {
                 final SecurityCheckResult result =  check.test(getContext());
 
+                final int remaining = count.decrementAndGet();
+
                 for (SecurityCheckExecutorListener listener : listeners) {
                     listener.onExecuted(result);
-                }
 
-                if (count.decrementAndGet() == 0) {
-                    for (SecurityCheckExecutorListener listener : listeners) {
+                    if (remaining <= 0) {
                         listener.onFinished();
                     }
                 }
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -98,10 +98,23 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
      * The key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
      * a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
      *
+     * @return {@link Map}
+     */
+    public Map<String, Future<SecurityCheckResult>> execute() {
+        return execute(new SecurityCheckExecutorListener[0]);
+    }
+
+    /**
+     * Executes the checks asynchronously.
+     *
+     * Returns a {@link Map} containing the results of each executed test (a {@link Future}).
+     * The key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
+     * a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
+     *
      * @param securityCheckExecutorListeners list of listeners that will receive events about checks execution
      * @return {@link Map}
      */
-    public Map<String, Future<SecurityCheckResult>> execute(SecurityCheckExecutorListener... securityCheckExecutorListeners) {
+    private Map<String, Future<SecurityCheckResult>> execute(SecurityCheckExecutorListener... securityCheckExecutorListeners) {
 
         final List<SecurityCheckExecutorListener> listeners =
             securityCheckExecutorListeners == null ? new ArrayList<>(1) : new ArrayList(Arrays.asList(securityCheckExecutorListeners));
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -10,11 +10,15 @@ import org.aerogear.mobile.core.logging.Logger;
 import org.aerogear.mobile.core.metrics.MetricsService;
 import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Executor used to asynchronously execute checks.
@@ -94,19 +98,40 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
      * The key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
      * a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
      *
+     * @param securityCheckExecutorListeners list of listeners that will receive events about checks execution
      * @return {@link Map}
      */
-    public Map<String, Future<SecurityCheckResult>> execute() {
+    public Map<String, Future<SecurityCheckResult>> execute(SecurityCheckExecutorListener... securityCheckExecutorListeners) {
 
-        final MetricsService metricsService = getMetricsService();
+        final List<SecurityCheckExecutorListener> listeners;
+        final Collection<SecurityCheck> checks = getChecks();
+
+        // Initializes listener list: the metric publisher will be added to passed in listeners
+        if (securityCheckExecutorListeners == null) {
+            listeners = new ArrayList<>(1);
+        } else {
+            listeners = new ArrayList(Arrays.asList(securityCheckExecutorListeners));
+        }
+
+        listeners.add(getMetricServicePublisher());
+
         final Map<String, Future<SecurityCheckResult>> res = new HashMap<>();
+        AtomicInteger count = new AtomicInteger(checks.size());
 
         for (final SecurityCheck check : getChecks()) {
             res.put(check.getName(), (executorService.submit(() -> {
                 final SecurityCheckResult result =  check.test(getContext());
-                if (metricsService != null) {
-                    metricsService.publish(new SecurityCheckResultMetric(result));
+
+                for (SecurityCheckExecutorListener listener : listeners) {
+                    listener.onExecuted(result);
                 }
+
+                if (count.decrementAndGet() == 0) {
+                    for (SecurityCheckExecutorListener listener : listeners) {
+                        listener.onFinished();
+                    }
+                }
+
                 return result;
             })));
         }
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -8,7 +8,6 @@ import org.aerogear.mobile.core.MobileCore;
 import org.aerogear.mobile.core.executor.AppExecutors;
 import org.aerogear.mobile.core.logging.Logger;
 import org.aerogear.mobile.core.metrics.MetricsService;
-import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -25,9 +24,6 @@ import java.util.concurrent.atomic.AtomicInteger;
  * Checks are executed by using {@link AppExecutors#singleThreadService()} if no custom executor is configured.
  */
 public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<AsyncSecurityCheckExecutor> {
-
-    private final static String TAG = "AsyncSecurityCheckExecutor";
-    private final static Logger LOG = MobileCore.getLogger();
 
     private final ExecutorService executorService;
 
@@ -117,7 +113,7 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
     private Map<String, Future<SecurityCheckResult>> execute(SecurityCheckExecutorListener... securityCheckExecutorListeners) {
 
         final List<SecurityCheckExecutorListener> listeners =
-            securityCheckExecutorListeners == null ? new ArrayList<>(1) : new ArrayList(Arrays.asList(securityCheckExecutorListeners));
+            securityCheckExecutorListeners == null ? new ArrayList<>(1) : new ArrayList<>(Arrays.asList(securityCheckExecutorListeners));
         final Collection<SecurityCheck> checks = getChecks();
 
         // Adds the metric publisher to the passed in listeners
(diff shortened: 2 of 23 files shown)


auth/build.gradle
```diff
@@ -23,7 +23,10 @@ android {
     }
 
     testOptions {
-        unitTests.includeAndroidResources = true
+        unitTests {
+            includeAndroidResources = true
+            returnDefaultValues = true
```
paolobueno: I did this because of http://tools.android.com/tech-docs/unit-testing-support#TOC-Method-...-not-mocked.- errors.

Not sure if it's okay but seemed easier than mocking and DI-ing JSONObject and Looper.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -101,16 +105,37 @@ public AsyncSecurityCheckExecutor build() {
...
 
         return res;
     }
+
+    private void publishMetricsAsync(final int howMany) {
+        final CompletionService<SecurityCheckResult> cs = new ExecutorCompletionService<>(executorService);
+        final ArrayList<SecurityCheckResult> results = new ArrayList<>(howMany);
+        for (int i = 0; i < howMany; i++) {
+            try {
+                SecurityCheckResult result = cs.take().get();
```
paolobueno: This seems to be the recommended way to consume these but I'm not sure if it's very solid.

@secondsun Is there a risk these could get mixed up with other tasks in the same `ExecutorService`?
It seems that by default it's the `AppExecutors#singleThreadService()` singleton and this method is set to be run in the `networkThread()`.

core/src/main/java/org/aerogear/mobile/core/metrics/Metrics.java
```diff
@@ -1,11 +1,9 @@
 package org.aerogear.mobile.core.metrics;
 
-import java.util.Map;
-
 public interface Metrics {
 
     String identifier();
 
-    Map<String, String> data();
+    Object data();
```
paolobueno: Had to do this to support `JSONArray`. `JSONArray` and `JSONObject` don't share any ancestors.

The goal is to be able to have an entry with the shape: 
```json
{
  {{identifier}}: [data]
}
```

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
secondsun: What do you mean by "mixed up"?

core/src/main/java/org/aerogear/mobile/core/metrics/Metrics.java
secondsun: Both implementations we provide use JSONObject, so why not a JSONObject?  In the future if we need this then we can raise the interface's type without breaking compatibility.

Alternatively we can define the the Metrics data as a generic type.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
paolobueno: @secondsun maybe the race condition is only in my head.

IIUC the executorService is a queue so I thought that maybe if somewhere else in the main thread other kinds of jobs are queued into the `singleThreadService` the `cs.take()` call could end up picking a result that's meant for another async operation.

secondsun:
I'm interested in seeing how this works.  Mind tossing in some steps for how to test it using openshift and the mobile features?

core/src/main/java/org/aerogear/mobile/core/metrics/Metrics.java
paolobueno: We want to reach the following shape from https://github.com/aerogear/aerogear-app-metrics/pull/33:
```
"security": [
      {
        "type": "org.aerogear.mobile.security.checks.DeveloperModeCheck",
        "passed": true
      },
     {
        "type": "org.aerogear.mobile.security.checks.EmulatorCheck",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.DebuggerCheck",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.RootedCheck",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.ScreenLockCheck",
        "passed": false
      }
    ]
```

So I changed `SecurityCheckResultMetric` to be `{identifier: "security", data: JSONArray()}`.
I guess we could somehow change `NetworkMetricsPublisher#publish` instead.

Do you see any benefit of having `Metrics<T>` without a type constraint over `Object`?

I didn't see much point doing that because even https://developer.android.com/reference/org/json/JSONObject.html#put(java.lang.String,java.lang.Object) receives `Object` over having overloads for `JSONObject|Array`

darahayes:
@paolobueno can you please revisit the server side PR here: https://github.com/aerogear/aerogear-app-metrics/pull/33

The server now expects the security checks to have a `name` field for displaying human readable labels in graphs/charts.

The format should now look like this:

```
    "security": [
      {
        "type": "org.aerogear.mobile.security.checks.DeveloperModeCheck",
        "name": "Developer Mode Check",
        "passed": true
      },
     {
        "type": "org.aerogear.mobile.security.checks.EmulatorCheck",
        "name": "Emulator Check",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.DebuggerCheck",
        "name": "Debugger Check",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.RootedCheck",
        "name": "Rooted Check",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.ScreenLockCheck",
        "name": "Screen Lock Check",
        "passed": false
      }
    ]
  }
```

auth/src/main/java/org/aerogear/mobile/security/metrics/SecurityCheckResultMetric.java
```diff
@@ -48,19 +66,32 @@ public String identifier() {
      * the value is <code>true</code> if the check result passed
      */
     @Override
-    public Map<String, String> data() {
-        return Collections.unmodifiableMap(data);
+    public JSONArray data() {
```
josemigallas: Please update doc comments.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -101,16 +105,37 @@ public AsyncSecurityCheckExecutor build() {
         final MetricsService metricsService = getMetricsService();
         final Map<String, Future<SecurityCheckResult>> res = new HashMap<>();
 
-        for (final SecurityCheck check : getChecks()) {
-            res.put(check.getName(), (executorService.submit(() -> {
-                final SecurityCheckResult result =  check.test(getContext());
-                if (metricsService != null) {
-                    metricsService.publish(new SecurityCheckResultMetric(result));
-                }
+        final Collection<SecurityCheck> checks = getChecks();
```
ziccardi: Wouldn't be easier like this:
```java
public Map<String, Future<SecurityCheckResult>> execute() {

        final MetricsService metricsService = getMetricsService();
        final Map<String, Future<SecurityCheckResult>> res = new HashMap<>();

        final Set<SecurityCheckResult> results = new HashSet<>();
        final Collection<SecurityCheck> checks = getChecks();

            res.put(check.getName(), (executorService.submit(() -> {
                final SecurityCheckResult result =  check.test(getContext());
                results.add(result);

                synchronized (results) {
                    if (metricsService != null && results.size() == checks.size()) {
                        metricsService.publish(new SecurityCheckResultMetric(result));
                    }
                }

                return result;
            })));
        }

        return res;
    }
```

auth/src/main/java/org/aerogear/mobile/security/metrics/SecurityCheckResultMetric.java
```diff
@@ -48,19 +66,32 @@ public String identifier() {
...
+        // TODO: consider returning a deep clone
```
josemigallas: I would not create the JSONArray during instantiation but rather store the results as class field and return a new JSONArray each time:
```java
    public SecurityCheckResultMetric(@NonNull final Iterable<SecurityCheckResult> results) {
        this.results = results;
    }

    public SecurityCheckResultMetric(@NonNull final SecurityCheckResult... results) {
        this.results = Arrays.asList(results);
    }

    public JSONArray data() {
        return getDataFromResult(this.results);
    }
```
Also I'd add a test to ensure data cannot be changed if this would be a problem.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
secondsun: Ahhh right.  So 1) The network thread is single threaded.  That isn't guaranteed, but it is right now.  2) If the order is important there are options for enforcing the order of the jobs, but I will have to refresh myself.  (Or we add a epic for weaving RxJava into the SDK implementation)

core/src/main/java/org/aerogear/mobile/core/metrics/Metrics.java
secondsun: The best thing that a generic type parameter gets us is a little bit of developer ease at code writing time.  Depending on how much people will put in their Metrics implementations it may or may not be beneficial.

What the generic type does is make explicit the data type that is being wrapped by the metrics class.  Functionality wise it won't behave differently than the current proposal, but it does make the down typing you can do on the data method more explicit.

ziccardi:
@josemigallas as you can see here, the `SyncSecurityCheckExecutor` executes the checks one after the other.
If you would wrap it in a `Future`:

1. All the checks would be executed in just one thread
2. The code will be less customizable

While, on the other hand, the code to be written would be almost the same.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -101,16 +105,37 @@ public AsyncSecurityCheckExecutor build() {
...
+
+        if(metricsService != null) {
+            (new AppExecutors()).networkThread()
+                .submit(() -> publishMetricsAsync(checkCount));
         }
 
         return res;
     }
+
+    private void publishMetricsAsync(final int howMany) {
```
ziccardi: With the suggested implementation, this method is useless

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
ziccardi: @paolobueno With the suggested implementation, this problem does not exists

auth/src/main/java/org/aerogear/mobile/security/SyncSecurityCheckExecutor.java
```diff
@@ -73,23 +71,23 @@ public SyncSecurityCheckExecutor build() {
...
-     * @param result {@link SecurityCheckResult} to be published
+     * @param results {@link SecurityCheckResult} to be published
      * @throws IllegalArgumentException if result is null
      */
-    private void publishResultMetrics(@NonNull SecurityCheckResult result) {
-        MetricsService metricsService = getMetricsService();
-
-        if (metricsService != null) {
-            metricsService.publish(new SecurityCheckResultMetric(nonNull(result, "result")));
+    private void publishResultMetrics(final @NonNull Map<String, SecurityCheckResult> results) {
```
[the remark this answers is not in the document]
ziccardi: 👍

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
paolobueno: I see, wasn't familiar with synchronized methods and blocks.

I was, however, thinking about ways that users could do this kind of queuing of actions themselves. Both of these approaches are only possible because we're in control of the Callables or aware of the ExecutorService.

I guess we can postpone these improvements for Rx :pray:

paolobueno:
I think having a single sync Executor and running it in a background thread could be enough. Paralellizing all the self-defence checks seems like a premature optimization, though the code is mostly in place.

Do you guys think there'll be any long-running self-defence check that would warrant being run in parallel?

Wasn't planning on doing this change as part of this PR though.

ziccardi:
@paolobueno Not sure I get it. The `AsyncSecurityCheckExecutor` is able to execute checks asynchronously using a passed in ExecutorService, so if a multithread executor is passed, checks are automatically executed in parallel.

I thought this PR was just about batching the metrics: even about that, I think we are making the code unnecessarily complex. 
If we want the metrics to be batched, that should be done by the metric service object.
If we want the user to be able to control the batch size (not sure we want that), then instead of passing a `metric service` object to the executor, we should pass a `metric service publisher` object which can be configured with the batch size.

```java
SecurityCheckExecutor.newAsyncCheckExecutor(context)
...
...
.withMetricsPublisher(new MetricPublisher(10))
.build().execute();
```
Then the publisher could (in a background thread):

1. publish the metrics whenever the list of received metrics gets bigger than X
2. publish the metrics every X seconds since the last time metrics have been published

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
ziccardi: @paolobueno TBH I think we shouldn't change the executor to batch the metrics, but batching should be done by the metric service (see my comment at the end of the PR)

paolobueno:
Making the publisher smarter was mentioned in https://issues.jboss.org/browse/AGDROID-771?focusedCommentId=13540352&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13540352 too.

Agree it'd be nice to have but I'd like to leave this out of the scope of this PR if possible.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -91,26 +91,41 @@ public AsyncSecurityCheckExecutor build() {
...
+
+    private synchronized void tallyResultsAndPublish(SecurityCheckResult result) {
+        MetricsService metricsService = this.getMetricsService();
+        if (metricsService == null) {
+            return;
+        }
+
+        int checkCount = getChecks().size();
+        if (results == null) {
+            results = new ArrayList<>(checkCount);
```
ziccardi: This way if the called will call 2 times the execute method, metrics will be messed up and some metric will be lost.
With the proposed approach, we would not have this issue.

paolobueno: I thought the method being `synchronized` would prevent this.

ziccardi: Nope. The result array is a member variable. If I call the execute two times, I will add the results twice in the same array

paolobueno: Ah, I see what you mean, the executor interface doesn't prevent it being executed multiple times. :+1:

danielpassos:
@paolobueno If I do execute from `securityService.getCheckExecutor()` the `metricsService` is always be `null` on `publishResultMetrics`. So how I suppose to use that?

Here is an example what I did:

```java 
securityService.getCheckExecutor()
    .addCheck(SecurityCheckType.ALLOW_BACKUP_ENABLED)
    .addCheck(SecurityCheckType.IS_EMULATOR)
    .addCheck(SecurityCheckType.IS_DEBUGGER)
    .addCheck(SecurityCheckType.IS_DEVELOPER_MODE)
    .addCheck(SecurityCheckType.IS_ROOTED)
    .addCheck(SecurityCheckType.SCREEN_LOCK_ENABLED)
    .addCheck(SecurityCheckType.HAS_ENCRYPTION_ENABLED)
    .execute();
```

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -91,26 +90,39 @@ public AsyncSecurityCheckExecutor build() {
...
                 return result;
-            })));
+            });
+            res.put(check.getType(), future);
         }
 
         return res;
     }
+
+    private synchronized void tallyResultsAndPublish(SecurityCheckResult result, List<SecurityCheckResult> results) {
```
ziccardi: @paolobueno This way it would work, however synchronising the whole method, you synchronize parts of the code that doesn't needs to be synchronized (thus slowing the execution).

I have created a PR to propose a different implementation for the batching. Would you mind giving that a look?
https://github.com/aerogear/aerogear-android-sdk/pull/129

ziccardi:
@danielpassos That's true. IMHO I think that the `getCheckExecutor`, `getAsyncCheckExecutor`, `checkAndSendMetric`(s) methods are confusing and makes the code less dynamic/maintainable.

To do what you want we should add a `setMetricService` method to the executor

However `addCheck` and eventually the `setMetricService` will create RACE conditions if the same executor instance is used in two different threads.

That does not happen if we use directly the builders since they were designed to produce immutable objects. Using the builders, your code would be:

```java
SecurityCheckExecutor
    .newSyncExecutor(getContext())  // or newAsyncExecutor(getContext()) if you need an async
    .withSecurityCheck(SecurityCheckType.ALLOW_BACKUP_ENABLED)
    .withSecurityCheck(SecurityCheckType.IS_EMULATOR)
    .withSecurityCheck(SecurityCheckType.IS_DEBUGGER)
    .withSecurityCheck(SecurityCheckType.IS_DEVELOPER_MODE)
    .withSecurityCheck(SecurityCheckType.IS_ROOTED)
    .withSecurityCheck(SecurityCheckType.SCREEN_LOCK_ENABLED)
    .withSecurityCheck(SecurityCheckType.HAS_ENCRYPTION_ENABLED)
    .withMetricService(metricService)
    .build()
    .execute();
```

wei-lee:
@paolobueno can you take a look at PR #129. I think it is a much simpler and cleaner way to implement this function.

paolobueno:
@ziccardi rebased this on #129 and removed the java.util.concurrent-based approach, only change I needed on top of it was ed523c6 in order to put the metrics inside a `"security":` key as per the schema from the metrics server.

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutorListener.java
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * Called after each check is executed
+     * @param result the result of the check
+     */
+    void onExecuted(SecurityCheckResult result);
+
+    /**
+     * Called when all submitted checks has been executed.
+     */
+    void onFinished();
```
ziccardi: as suggested by @matzew in the other PR, rename this to `onComplete`

paolobueno: done!

PR #127, opened 2018-03-05
Metrics integration test

psturc:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2237

## Description

* added deploying of [app metrics (metrics api)](https://github.com/aerogear/aerogear-app-metrics) as a prerequisite of metrics integration test
* TBD: load url of metrics api and run metrics integration test (@josemigallas)
* this test will be triggered if PR is containing label "test/integration"

## Progress

- [x] Deploy app-metrics to Wendy, get its URL and write to file
- [x] Load metrics URL from file, run metrics integration test

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,67 @@
+#!groovy
+
+// https://github.com/feedhenry/fh-pipeline-library
+@Library('fh-pipeline-library') _
+
+def prLabels = getPullRequestLabels {}
+
+if ( prLabels.contains("test/integration") ) {
+    def metricsApiHost
+    def repositoryName = "android-sdk"
+    def projectName = "test-${repositoryName}-${currentBuild.number}-${currentBuild.startTimeInMillis}"
+    def linkToApiMetricsTemplate = "https://raw.githubusercontent.com/aerogear/aerogear-app-metrics/master/openshift-template.yml"
+    def apiMetricsTemplateFilename = "api-metrics-template.yml"
+
+    stage('Trust') {
+      enforceTrustedApproval('aerogear')
+    }
+
+    openshift.withCluster() {
+        node ('jenkins-openshift') {
+            stage ('Deploy app metrics service') {
+
+                sh "curl ${linkToApiMetricsTemplate} > ${apiMetricsTemplateFilename}"
+
+                openshift.newProject(projectName)
+                openshift.withProject(projectName) {
+                    def routeSelector
+
+                    openshift.newApp( "--file", "${apiMetricsTemplateFilename}" )
+                    // Get URL of deployed metricsApi
+                    routeSelector = openshift.selector("route", "aerogear-app-metrics")
+                    metricsApiHost = "https://" + routeSelector.object().spec.host
+                }
+            }
+
+            node ('osx') {
+                stage('Cleanup osx workspace') {                    
+                    // Clean workspace
+                    deleteDir()
+                }
+    
+                stage ('Checkout to osx slave') {
+                    checkout scm
+                }
+    
+                stage ('Prepare config file') {
+                    def servicesConfigJsonPath = "./core/src/test/assets/integration-test-mobile-services.json"
+                    def servicesConfigJson = readJSON file: servicesConfigJsonPath
+
+                    // Update URL for metrics in mobile-services.json file
+                    servicesConfigJson.services.each {
+                        if ( it.id == "metrics" ) {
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -1,12 +1,52 @@
...
+
+// https://github.com/feedhenry/fh-pipeline-library
+@Library('fh-pipeline-library') _
+
+stage('Trust') {
+    enforceTrustedApproval('aerogear')
+}
+
+def prLabels = getPullRequestLabels {}
+def metricsApiTemplateUrl = "https://gist.githubusercontent.com/psturc/a2a7150b66d48369e5fa62ab72352330/raw/91acde6f764e9bc59a84e3e34fa694206eb0a877/aerogear-app-metrics.yml"
```
danielpassos: Should this come from a gist and from your account? Is not possible to put it in this repo?

darahayes: +1 on placing it in the repo. I would like to review the contents of the gist. Also, the app metrics service is still potentially subject to config and startup changes which would break this gist.

psturc: Template added

secondsun:
I feel like the yml file should be in a subdirectory and not in the root of the project, but that may just me be bikeshedding.

I'm not familiar enough with Jenkins to give a great review, @psturc  can you give a demo sometime[on BJ or just link a video here]?  (I see Jenkins failed the build, but that tends to happen with good code.)

psturc:
@secondsun what would you like to see in the demo?
Moving the folder to subdirectory is actually good idea (done now). With more integration tests we might need more templates and test-related things and having them in root folder would be quite messy.

Jenkinsfile
```diff
@@ -1,12 +1,52 @@
...
   args: '${computer.jnlpmac} ${computer.name}',
-  ttyEnabled: false) 
+  ttyEnabled: false)
+
+def runIntegrationTest() {
+  def repositoryName = "android-sdk"
+  def projectName = "test-${repositoryName}-${currentBuild.number}-${currentBuild.startTimeInMillis}"
+  def metricsApiTemplatePath = "./ci/aerogear-app-metrics.yml"
+  
+    openshift.withCluster() {
```
wojta: Does it deploy to Wendy's OpenShift or somewhere else?

psturc: Currently yes. We will probably want to make a use of the slaves behind VPN later.
Related ticket: https://issues.jboss.org/browse/AEROGEAR-2292

secondsun:
@psturc Basically a here's a jenkins running in openshift, here's how you make the test run.  Here's how you know the test ran.  I suppose I can do that with a doc too.  I just haven't touched our current stack of Jenkins testing stuff and I need to learn more :)

secondsun:
But things look good as far as I can tell, feel free to merge :)

psturc:
@secondsun some things you've mentioned are already contained in [this video](https://youtu.be/ByDW8a9bhaw) I made after we implemented PR tests for APBs, but once we're done with this integration test for all platforms, I think it would be worth to do a little demo too. Stay tuned :P

Jenkinsfile
```diff
@@ -0,0 +1,62 @@
...
+      openshift.newProject(projectName)
+      openshift.withProject(projectName) {
+          openshift.newApp( "--file", "${metricsApiTemplatePath}" )
+          // Get URL of deployed metricsApi
+          def routeSelector = openshift.selector("route", "aerogear-app-metrics")
+          def metricsApiHost = "http://" + routeSelector.object().spec.host
+          def servicesConfigJsonPath = "./core/src/debug/assets/integration-test-mobile-services.json"
```
secondsun: we are moving the mobile-services files from debug/assets to test/assets

josemigallas: @psturc @danielpassos 
Heads up, the folder where mobile-services JSON files is going to change from `debug` to `test` #167

josemigallas: 🥇

psturc: thanks for heads up! fixed

secondsun:
@psturc Looks like Jenkins is still having trouble building, I checked the logs and it is claiming it couldn't download a dependency.  The dependency is live as far as I can tell, but it did take a few seconds for the download to begin.  If there a way to kick Jenkins back off?

psturc:
@secondsun hey, so now the builds should be stable (and a lot faster) since we're using `osx` slave in our lab.

PR #128, opened 2018-03-05
Adding Spotless to enforce style

secondsun:
## Motivation

There have been several comments about code style, and there has been a PR for style fixes already.

This enforces and autofixes style at the gradle level so this will NEVER BE A PROBLEM AGAIN!

## Discussion

I have defaulted to the google java code guidelines.  We may want to pick another.  

https://google.github.io/styleguide/javaguide.html

## Additional Notes

Do not review the autofix commit.  It is insanity.  Instead if you see a style you don't like let's discuss it.


paolobueno:
:+1: Will leave it to SMEs approve but let's get this merged soon so the rebasefest can start.

josemigallas:
@secondsun 2 spaces indentation?

wtrocki:
If you format this using Android Studio does it show some changes?
I think it's we could align profile with the default android studio profile. 
This will minimize amount of changes and make developers life less difficult.

secondsun:
@wtrocki  probably should update .editorconfig in this PR to make the styles align so that autoformat doesn't make a lot of work.

wtrocki:
@secondsun That will resolve the problem!

secondsun:
@josemigallas It is the google style.  At least it isn't tabs ;)

secondsun:
https://docs.google.com/forms/d/1Sfpzxmbgi2nh278LXKHntEGdqyxlx-HKXTV6n0NCWjc/edit?usp=sharing

A form for people to put their ideas into if we want to make changes.

wei-lee:
@secondsun Personally I think we should treat styles like Golang - just set one and go with it. I think it is a waste of time to debate code styles...

secondsun:
@wei-lee I'm fine with giving people this one opportunity as an illusion of choice ;)

secondsun:
@wtrocki So it looks like Google doesn't provide an editor config, but they do provide https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml which is the intellij settings.

We may just want to flag that in the README or contributing sections.

CONTRIBUTING.md
```diff
@@ -46,3 +46,7 @@ Run the following command to install the dependencies for the Android SDK:
 In Android Studio, go to `Tools > Android > Sync Project With Gradle Files`.
 
 See the [Gradle Documentation](https://docs.gradle.org/current/userguide/pt02.html) for more information on Gradle and the Gradle Wrapper
+
+## Adding Our Style Guide to Android Studio
+
+We are using spotless to enforce our style guidelines on the project.  You may invoke spotless manually by running `./gradlew check`.  If there are errors spotless will prompt you to run `./gradlew spotlessApply` and auto fix them.  If your code fails to meet the style guidelines it will fail.  However, if you are using Android Studio you may import our style guidelines into your IDE using our [intellij-code-style.xml](codequalite/intellij-code-style.xml) file.
```
danielpassos: Hrm, `./gradlew check` will cause an error because of the debug JSON file problem. 

```
Caused by: java.io.FileNotFoundException at MobileCoreTest.java:42
```

Maybe we should replace it with `./gradlew spotlessCheck`?

danielpassos:
@secondsun [intellij-java-google-style.xml](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) and [googleJavaFormat](https://github.com/diffplug/spotless/blob/master/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java) are really weird. I couldn't find any Android project on Google and Firebase repo using 2 spaces for _INDENT_SIZE_. I really don't like this, the code seems to strange.

danielpassos:
@secondsun is possible to we (easy) go with 4 spaces instead of 2?

secondsun:
So Right now I am trying to sort out the differences between telling Android studio to format the code and telling spotless to format the code.

You would think having the same rules format the same input would yield the same output.  You would be wrong.

secondsun:
Ok, so I've replaced the intelliJ formatter with a plugin that uses the google java formatter.  If you apply both spotless and intelliJ (with the plugin) formatting you will get code that is mostly the same.  There is an open issue in intelliJ to package its code formatter as a library so we can use it in spotless. (https://youtrack.jetbrains.com/issue/IDEA-180965).  Once that issue is closed we should revisit the autoformatted and use an intelliJ file.

secondsun:
I"m taking a look at https://github.com/krasa/EclipseCodeFormatter#instructions.  We might be able to get a style file, ci validation, and ide integration, and have everything be perfect!

PR #129, opened 2018-03-06
Added metric publishing batching

ziccardi:
Added event listener

## Motivation

Implementation proposal for batch metric publishing that avoids polluting the code with logic to batch the metrics.

## Description

We currently want to avoid modifying the metric service to add batch received metrics, so we have to add some batching logic to the async/sync security check executors.

This implementation adds the ability to pass `listeners` to the execute method, so that events can be received as the checks are executed (`onExecuted` and `onFinished`).
A `SecurityCheckMetricPublisher` class implementing the listener interface is provided to batch the metrics and publish them when `onFinished` is received.

With this approach we have two benefits:
1. The code to batch the metrics is outside the executor class
2. We provide the user another useful interface

## Improvements

We could change the publisher so that after a number of metrics have been received, the metrics get published anyway, even before `onFinish` is called.

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -116,18 +116,18 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
         listeners.add(getMetricServicePublisher());
 
         final Map<String, Future<SecurityCheckResult>> res = new HashMap<>();
-        AtomicInteger count = new AtomicInteger(checks.size());
+        final AtomicInteger count = new AtomicInteger(checks.size());
 
         for (final SecurityCheck check : getChecks()) {
             res.put(check.getName(), (executorService.submit(() -> {
                 final SecurityCheckResult result =  check.test(getContext());
 
+                final int remaining = count.decrementAndGet();
+
                 for (SecurityCheckExecutorListener listener : listeners) {
                     listener.onExecuted(result);
-                }
 
-                if (count.decrementAndGet() == 0) {
-                    for (SecurityCheckExecutorListener listener : listeners) {
+                    if (remaining <= 0) {
                         listener.onFinished();
                     }
                 }
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -98,10 +98,23 @@ public class AsyncSecurityCheckExecutor extends AbstractSecurityCheckExecutor<As
      * The key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
      * a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
      *
+     * @return {@link Map}
+     */
+    public Map<String, Future<SecurityCheckResult>> execute() {
+        return execute(new SecurityCheckExecutorListener[0]);
+    }
+
+    /**
+     * Executes the checks asynchronously.
+     *
+     * Returns a {@link Map} containing the results of each executed test (a {@link Future}).
+     * The key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
+     * a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
+     *
      * @param securityCheckExecutorListeners list of listeners that will receive events about checks execution
      * @return {@link Map}
      */
-    public Map<String, Future<SecurityCheckResult>> execute(SecurityCheckExecutorListener... securityCheckExecutorListeners) {
+    private Map<String, Future<SecurityCheckResult>> execute(SecurityCheckExecutorListener... securityCheckExecutorListeners) {
 
         final List<SecurityCheckExecutorListener> listeners =
             securityCheckExecutorListeners == null ? new ArrayList<>(1) : new ArrayList(Arrays.asList(securityCheckExecutorListeners));
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -10,11 +10,15 @@ import org.aerogear.mobile.core.logging.Logger;
 import org.aerogear.mobile.core.metrics.MetricsService;
 import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private final MetricsService metricsService;
+    private final List<SecurityCheckResultMetric> metrics = new ArrayList<>();
+
+    /**
+     * Builds the object.
+     *
+     * @param metricsService metric service to be used to publish the metrics
+     */
+    SecurityCheckMetricPublisher(final MetricsService metricsService) {
```
ziccardi: We could pass an executor here, to publish the metrics asynchronously. WDYT?

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -94,19 +98,40 @@ public AsyncSecurityCheckExecutor build() {
...
+                for (SecurityCheckExecutorListener listener : listeners) {
+                    listener.onExecuted(result);
+                }
+
+                if (count.decrementAndGet() == 0) {
+                    for (SecurityCheckExecutorListener listener : listeners) {
+                        listener.onFinished();
+                    }
                 }
+
```
wei-lee: this definitely is a clean way to supporting the batch metrics. However, for now, I don't think we want to expose this method to developers as it is making the API quite complicated. This is already beyond the original requirement.

So can I suggest let's make this method private, and create a new public `execute` method that takes no parameters. In the implementation of that method, it will add a metrics listener and call the private method.

This way if this feature proves useful in the future, we can make it public at that point.

WDYT?

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
wei-lee: I think the metrics service itself should choose the right executor. For now, let's not worry about it.

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
ziccardi: We already had a similar chat about this ;) I don't think it really adds complexity since Listeners are pretty standard in Java programming, however it won't be a big deal to hide it, so I can certainly do that.

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutorListener.java
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * Called after each check is executed
+     * @param result the result of the check
+     */
+    void onExecuted(SecurityCheckResult result);
+
+    /**
+     * Called when all submitted checks has been executed.
+     */
+    void onFinished();
```
matzew: I'd call it `onComplete` to be compliant with RxJava style programming

ziccardi: 👍

matzew: what's happening onError ?

matzew:
I wonder if this should be in the example app, for showing the value of this ?

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutorListener.java
ziccardi: Currently checks return just `passed` (true/false) and don't throw any exception.

matzew:
My main concern here is really, when we implement our own listeners and their handlings... why not leveraging something like RXJava for this? I think that we should at some point start looking to integrate that library into our SDK. The lib is pretty popular w/ Android folks, so I'd (as a developer) kinda expect new SDKs/API directly leveraging this

ziccardi:
@matzew We are going to hide this functionality from the end user (see other comments from @wei-lee)

darahayes:
@ziccardi Quick question, does this result in a payload similar to the following?

```
{
  "clientId": "9796fb69-a566-43fb-b932-03951d82437f",
  "timestamp": 1519989859342,
  "data": {
    "security": [
      {
        "type": "org.aerogear.mobile.security.checks.DeveloperModeCheck",
        "name": "Developer Mode Check",
        "passed": true
      },
     {
        "type": "org.aerogear.mobile.security.checks.EmulatorCheck",
        "name": "Emulator Check",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.DebuggerCheck",
        "name": "Debugger Check",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.RootedCheck",
        "name": "Rooted Check",
        "passed": false
      },
      {
        "type": "org.aerogear.mobile.security.checks.ScreenLockCheck",
        "name": "Screen Lock Check",
        "passed": false
      }
    ]
  }
}
```

ziccardi:
@darahayes I guess so: all results from all the executed checks in a single call are sent to the metric service at the same time.

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutorListener.java
paolobueno: Would it be too much trouble passing all the `results` here?

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
paolobueno: :+1:  I think this would be done at the same time as the metrics service being improved to batch sequential requests by itself.

paolobueno:
@ziccardi Thanks for the PR, love event listeners myself.

In order to reach the goals for https://issues.jboss.org/browse/AGDROID-771 as @darahayes pointed I still need to 'merge' this with some of the changes in #126 for the SecurityCheckResult and the SecurityCheckResultMetric in order to match the shape expected by the back-end.

How do you suggest I do this? I could 'rebase' those changes on this branch PR against your personal fork.

paolobueno:
@ziccardi Actually I'm having quite a bit of success rebasing my branch on this one, will force push to #126 when successful.

wei-lee:
@matzew +1 on leveraging RxJava. I think it worth creating a Github issue so we can track it.

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckExecutorListener.java
ziccardi: @paolobueno It wouldn't, but I don't like the idea of caching all the results in the executor. Currently the results are very small objects and the number of checks is small, but we can never know what is going to happen in the future.
Better to pass them one by one and let the receiver decide what to do with them

ziccardi:
@paolobueno I'm just going to make a small change on this PR to hide the new method from the end user.

Issue #89, opened 2018-02-23
Expose Users First/Last Name


TommyJ1994:
## Expected Behavior
The currentUser.getName() returns the users First Name + Last Name.

## Actual Behavior
The users username is returned instead.

I think it might be better to call this getUsername() instead. I think we should also expose a getFirstName() and a getLastName() as these will be used to show the users name in the UI/profile. Optionally, the getName() could returns the first name and last name together. Username are often used as identifiers, and the users actual name can be used in the UI.

PR #134, opened 2018-03-06
Give users first and last names

aidenkeating:
## Motivation

We only currently expose the users username. This is not the same as a full name, it's a unique identifier that could be anything. We should also expose the first and last name of the user.

## Description

Updating the user principal to allow for a first name and a last name to be provided to it and retrieved from it. Update the example app to use `getUsername` instead of `getName`.

Resolves https://github.com/aerogear/aerogear-android-sdk/issues/89

## Progress

- [x] Implementation
- [x] Documentation
- [x] Unit tests

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipal.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipal.java
@@ -32,7 +32,28 @@ public interface UserPrincipal extends Serializable {
      * 
      * @return the username
      */
+    String getUsername();
+
+    /**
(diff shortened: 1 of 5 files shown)


aidenkeating:
@secondsun @danielpassos @TommyJ1994 Mind taking a look?

danielpassos:
@aidenkeating Rebase and feel free to merge

TommyJ1994:
@aidenkeating LGTM :+1:

I created https://github.com/feedhenry/mobile-security-android-template/issues/38 to track updating the mobile security template app with the new functions. I can update this as part of some of my changes.

PR #137, opened 2018-03-07
Fixes from e2e tests

paolobueno:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-2218
https://issues.jboss.org/browse/AEROGEAR-2266

## Description

Readd the type and name fields to metrics data.

## Progress

- [x] Add `getDisplayName`
- [x] Rename `getName` to `getType`

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -94,8 +94,8 @@ public class AsyncSecurityCheckExecutor
      * Executes the checks asynchronously.
      *
      * Returns a {@link Map} containing the results of each executed test (a {@link Future}). The
-     * key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
-     * a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
+     * key of the map will be the output of {@link SecurityCheck#getDisplayName()}}, while the value
+     * will be a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
      *
      * @return {@link Map}
      */
@@ -107,8 +107,8 @@ public class AsyncSecurityCheckExecutor
      * Executes the checks asynchronously.
      *
      * Returns a {@link Map} containing the results of each executed test (a {@link Future}). The
-     * key of the map will be the output of {@link SecurityCheck#getName()}, while the value will be
-     * a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
+     * key of the map will be the output of {@link SecurityCheck#getDisplayName()}, while the value
+     * will be a {@link Map} of {@link Future} with the {@link SecurityCheckResult} of the check.
      *
      * @param securityCheckExecutorListeners list of listeners that will receive events about checks
      *        execution
@@ -129,7 +129,7 @@ public class AsyncSecurityCheckExecutor
         final AtomicInteger count = new AtomicInteger(checks.size());
 
         for (final SecurityCheck check : checks) {
-            res.put(check.getName(), (executorService.submit(() -> {
+            res.put(check.getDisplayName(), (executorService.submit(() -> {
                 final SecurityCheckResult result = check.test(getContext());
 
                 final int remaining = count.decrementAndGet();
--- a/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
@@ -94,8 +94,8 @@ public class AsyncSecurityCheckExecutor
      * Executes the checks asynchronously.
      *
      * Returns a {@link Map} containing the results of each executed test (a {@link Future}). The
-     * key of the map will be the output of {@link SecurityCheck#getType()}, while the value will be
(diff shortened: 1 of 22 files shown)


auth/src/main/java/org/aerogear/mobile/security/SecurityCheckType.java
```diff
@@ -72,9 +72,9 @@ public SecurityCheck getSecurityCheck() {
     /**
      * Gets the name of the security check.
```
darahayes: comment rot - Please update this.

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckType.java
```diff
@@ -72,9 +72,9 @@ public SecurityCheck getSecurityCheck() {
...
      *
-     * @return {@link String} {@link SecurityCheck#getName()}
+     * @return {@link String} {@link SecurityCheck#getType()}
      */
     public String getName() {
```
darahayes: This seems confusing to me. It's a method called getName() that internally calls getType()

auth/src/main/java/org/aerogear/mobile/security/AsyncSecurityCheckExecutor.java
```diff
@@ -129,7 +129,7 @@ public AsyncSecurityCheckExecutor build() {
         final AtomicInteger count = new AtomicInteger(checks.size());
 
         for (final SecurityCheck check : checks) {
-            res.put(check.getName(), (executorService.submit(() -> {
+            res.put(check.getDisplayName(), (executorService.submit(() -> {
```
ziccardi: Use getType instead of getDisplayName

wei-lee: +1. The display name is not guaranteed to be unique, but type is.

paolobueno:
@ziccardi do you have any other suggestions for the method names? like getName()` and `getDisplayName()`...

darahayes:
@ziccardi from a maintainer/contributor perspective is also unintuitive to have a map with the key of `type` and the value returned from `getName()`. This is even worse now because there is also a `name` key in the same map. This would be really confusing.

The key prop should map directly to the name of the accessor method and to the actual value, i.e. `{type: check.getType(), name: check.getName()}` I feel like not adhering to this will definitely come back to bite us later on.

I also agree that `getType()` might seem unintuitive to an end user of the SDK. 

The class name e.g. `org.aerogear.security.foo` is being used as a unique identifier for checks. , perhaps something like `getId()` would more appropriate.

For example

```
{id: check.getId(), name: check.getName()} //pseudocode
```

which results in:

```
{id: "org.aerogear.security.EmulatorCheck", name: "Emulator Check"}
```

What do you think?

darahayes:
I'd like one of the following:

### 1. The current PR which does the following

```
{type: check.getType(), name: check.getName()}
```

resulting in

```
{type: "org.aerogear.security.EmulatorCheck", name: "Emulator Check"}
```

This does not require any changes on the server side.

### 2. Changes Paolo suggested:

```
{name: check.getName(), displayName: check.getDisplayName()}
```

resulting in

```
{name: "org.aerogear.security.EmulatorCheck", displayName: "Emulator Check"}
```

### 3. Another suggestion:

```
{id: check.getId(), name: check.getName()} //pseudocode
```

which results in:

```
{id: "org.aerogear.security.EmulatorCheck", name: "Emulator Check"}
```

As long as the key name matches the method name I'm fairly happy. Options 2 and 3 would require a minor change on the server side but it's tiny.

wei-lee:
I prefer option 1 or option 3. Don't think we need to explicitly call it `displayName`. Option 3 might be a little bit better as value is indeed used as an id. Also calling it `id` usually implies it should be unique.

ziccardi:
@darahayes I believe that the 1st one is confusing for the end user.
I don't like the 2nd one

If I should choose, I would go for the 3rd one

darahayes:
I prefer option 3 also. I will submit an appropriate PR against the app metrics server in the next few minutes

auth/src/main/java/org/aerogear/mobile/security/metrics/SecurityCheckResultMetric.java
```diff
@@ -24,7 +24,7 @@
     private final String TAG = "SecurityCheckResultMetric";
 
     public static final String IDENTIFIER = "security";
-    public static final String KEY_TYPE = "type";
+    public static final String KEY_ID = "ID";
```
darahayes: can this be `id`? all other keys are lowercase

auth/src/main/java/org/aerogear/mobile/security/impl/SecurityCheckResultImpl.java
```diff
@@ -26,8 +26,8 @@
      */
     public SecurityCheckResultImpl(@NonNull SecurityCheck check, final boolean passed) {
         SecurityCheck nonNullCheck = nonNull(check, "check");
-        this.type = nonNullCheck.getType();
-        this.displayName = nonNullCheck.getDisplayName();
+        this.type = nonNullCheck.getId();
+        this.displayName = nonNullCheck.getName();
```
darahayes: is `this.type` and `this.displayName` confusing?

darahayes:
@ziccardi @paolobueno @wei-lee please see associated PR for the server side implementation. I've tagged you guys for approval.

PR #141, opened 2018-03-07
fixed mobile-services.json (it was broken)

wojta:
## Motivation
Previous metrics PR broke default  mobile-services.json. It was breaking my E2E tests.


paolobueno:
@pb82 let's replace this value with the default url for a `myproject` deployment of the metrics apb on local openshift origin

pb82:
@paolobueno yes, makes sense. Ideally we would always use the mobile-cli to generate the config, but for the Demo App i think it's fine to have a working default URL.

PR #145, opened 2018-03-08
Add getting started video to auth SDK

aidenkeating:
## Motivation

Currently we do not link the video providing an end to end example
of using Keycloak in our documentation. This change adds the video
in the form of an image that once clicked will open YouTube.

The reason for not using the asciidoc video tag is because it does
not create the desired effect when rendered on GitHub.

This seems to be the best way to go about it if we want some visual
thing instead of just a link.

## Description
Add an image that links to the YouTube video.

## Progress

- [x] Add image with link


aidenkeating:
@finp @danielpassos Mind taking a look?

laurafitzgerald:
@danielpassos it was requested that this video be included at the Review Meeting.

PR #146, opened 2018-03-08
Implement the http response handling for metrics

josemigallas:
## Motivation

Currently the outcome of `MetricsPublisher.publish()` is handled internally by the SDK, hence it can't be handled by the user or tested by us.

JIRA: https://issues.jboss.org/browse/AGDROID-784

This is blocking: https://issues.jboss.org/browse/AEROGEAR-2238

## Description

This PR adds a nullable callback to the publish methods signature in `MetricsPublisher` and adds new implementations in `MetricsService`. The callback will be called by the pertinent implementation, namely `NetworkMetricsPublisher` and `LoggerMetricsPublisher`, based on the publication being a success or a failure.

It also does a couple of refactorizations:
https://github.com/josemigallas/aerogear-android-sdk/commit/b77f11910f4f0e3726425b97bdfb4b78d81d4f98
https://github.com/josemigallas/aerogear-android-sdk/commit/3b8913052dd992b1083468240b015f4b4d0da5ce

## Progress

- [x] Handle HTTP response
- [ ] testCallbackErrorMethodIsCalled

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
@@ -32,6 +32,7 @@ class SecurityCheckMetricPublisher implements SecurityCheckExecutorListener {
 
     @Override
     public synchronized void onComplete() {
-        metricsService.publish(new SecurityCheckResultMetric(metricResults));
+        metricsService.publish(new SecurityCheckResultMetric[] {
+                        new SecurityCheckResultMetric(metricResults)}, null);
     }
 }
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
@@ -32,7 +32,6 @@ class SecurityCheckMetricPublisher implements SecurityCheckExecutorListener {
 
     @Override
     public synchronized void onComplete() {
-        metricsService.publish(new SecurityCheckResultMetric[] {
-                        new SecurityCheckResultMetric(metricResults)}, null);
+        metricsService.publish(new SecurityCheckResultMetric(metricResults));
     }
 }
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
@@ -132,7 +132,9 @@ public class SecurityService implements ServiceModule {
     public SecurityCheckResult checkAndSendMetric(final SecurityCheck securityCheck,
                     @NonNull final MetricsService metricsService) {
         final SecurityCheckResult result = check(securityCheck);
-        nonNull(metricsService, "metricsService").publish(new SecurityCheckResultMetric(result));
+        nonNull(metricsService, "metricsService").publish(
+                        new SecurityCheckResultMetric[] {new SecurityCheckResultMetric(result)},
+                        null);
         return result;
     }
 }
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
@@ -132,9 +132,7 @@ public class SecurityService implements ServiceModule {
     public SecurityCheckResult checkAndSendMetric(final SecurityCheck securityCheck,
                     @NonNull final MetricsService metricsService) {
         final SecurityCheckResult result = check(securityCheck);
-        nonNull(metricsService, "metricsService").publish(
-                        new SecurityCheckResultMetric[] {new SecurityCheckResultMetric(result)},
-                        null);
+        nonNull(metricsService, "metricsService").publish(new SecurityCheckResultMetric(result));
         return result;
     }
 }
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
@@ -1,5 +1,6 @@
 package org.aerogear.mobile.core.metrics;
 
+import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import org.aerogear.mobile.core.Callback;
@@ -12,6 +13,6 @@ public interface MetricsPublisher {
      * @param metrics a array of metrics objects to publish
      * @param callback callback of the publication
      */
-    void publish(Metrics[] metrics, @Nullable final Callback callback);
+    void publish(@NonNull final Metrics[] metrics, @Nullable final Callback callback);
 
 }
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
@@ -4,9 +4,10 @@ public interface MetricsPublisher {
 
     /**
      * Allows to publish metrics to external source
-     * 
+     *
+     * @param listener a callback listener to handle the result of a publication
      * @param metrics a array of metrics objects to publish
      */
-    void publish(Metrics... metrics);
+    void publish(MetricsPublisherListener listener, Metrics... metrics);
 
 }
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
@@ -1,5 +1,7 @@
 package org.aerogear.mobile.core.metrics;
 
+import android.support.annotation.Nullable;
+
 import org.aerogear.mobile.core.Callback;
 
 public interface MetricsPublisher {
@@ -10,6 +12,6 @@ public interface MetricsPublisher {
      * @param metrics a array of metrics objects to publish
      * @param callback callback of the publication
      */
-    void publish(Metrics[] metrics, Callback callback);
+    void publish(Metrics[] metrics, @Nullable final Callback callback);
 
 }
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
@@ -1,13 +1,15 @@
 package org.aerogear.mobile.core.metrics;
 
+import android.support.annotation.Nullable;
+
 public interface MetricsPublisher {
 
     /**
      * Allows to publish metrics to external source
      *
-     * @param listener a callback listener to handle the result of a publication
      * @param metrics a array of metrics objects to publish
+     * @param listener a callback listener to handle the result of a publication
      */
-    void publish(MetricsPublisherListener listener, Metrics... metrics);
+    void publish(Metrics[] metrics, @Nullable MetricsPublisherListener listener);
 
 }
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
@@ -1,6 +1,6 @@
 package org.aerogear.mobile.core.metrics;
 
-import android.support.annotation.Nullable;
+import org.aerogear.mobile.core.Callback;
 
 public interface MetricsPublisher {
 
@@ -8,8 +8,8 @@ public interface MetricsPublisher {
      * Allows to publish metrics to external source
      *
      * @param metrics a array of metrics objects to publish
-     * @param listener a callback listener to handle the result of a publication
+     * @param callback callback of the publication
      */
-    void publish(Metrics[] metrics, @Nullable MetricsPublisherListener listener);
+    void publish(Metrics[] metrics, Callback callback);
 
 }
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
@@ -1,8 +1,0 @@
-package org.aerogear.mobile.core.metrics;
-
-public interface MetricsPublisherListener {
-
-    void onPublishMetricsSuccess();
-
-    void onPublishMetricsError(Exception error);
-}
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
@@ -1,11 +1,8 @@
 package org.aerogear.mobile.core.metrics;
 
-
-import org.aerogear.mobile.core.http.HttpResponse;
-
 public interface MetricsPublisherListener {
 
-    void onPublishMetricsSuccess(HttpResponse response);
+    void onPublishMetricsSuccess();
 
     void onPublishMetricsError(Exception error);
 }
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
@@ -1,0 +1,11 @@
+package org.aerogear.mobile.core.metrics;
+
+
+import org.aerogear.mobile.core.http.HttpResponse;
+
+public interface MetricsPublisherListener {
+
+    void onPublishMetricsSuccess(HttpResponse response);
+
+    void onPublishMetricsError(Exception error);
+}
--- a/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
+++ b/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
@@ -16,6 +16,7 @@ public class MetricsService implements ServiceModule {
 
     private Metrics[] defaultMetrics;
     private MetricsPublisher publisher;
+    private MetricsPublisherListener listener;
 
     public MetricsPublisher getPublisher() {
         return publisher;
@@ -68,16 +69,14 @@ public class MetricsService implements ServiceModule {
      * Send metrics
      *
      * @param metrics Metrics to send
-     * @return this MetricsService instance
      */
-    public MetricsService publish(final Metrics... metrics) {
+    public void publish(final Metrics... metrics) {
         nonNull(metrics, "metrics");
         if (publisher == null) {
             throw new IllegalStateException(
                             "Make sure you have called configure or get this instance from MobileCore.getInstance()");
         }
         publisher.publish(metrics);
-        return this;
     }
 
 }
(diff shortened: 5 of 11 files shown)


wtrocki:
-1000 for that change.

wtrocki:
Looks good to me.

secondsun:
Why can't you pass in a mocked httprequest?

wtrocki:
Actually mocked httprequest should be the way to resolve this issue.

josemigallas:
@wtrocki looks good or not? I'm confused 😅 
@secondsun The listener is attached to the metrics service no a single request. Ot at least that was my intention. As per the mock, I tried to mock or spy the class that handles de response but Mockito wouldn't allow me to do it. We would need to use PowerMockito or something similar to mock static methods and I don't know how appropriate is that. I wanted to avoid doing something like this but I couldn't found I way.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
```diff
@@ -0,0 +1,11 @@
+package org.aerogear.mobile.core.metrics;
+
+
+import org.aerogear.mobile.core.http.HttpResponse;
+
+public interface MetricsPublisherListener {
```
danielpassos: 1. It's not a listener it's more a callback
2. Why we need a specific one for metrics?

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
```diff
@@ -0,0 +1,11 @@
...
+
+    void onPublishMetricsSuccess(HttpResponse response);
```
danielpassos: Why do we need to expose response if it was a success?

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
```diff
@@ -18,7 +19,7 @@ public LoggerMetricsPublisher(final Logger logger) {
     }
 
     @Override
-    public void publish(final Metrics... metrics) {
+    public void publish(MetricsPublisherListener listener, final Metrics... metrics) {
```
danielpassos: There is no real reason but usually, the callback is the last parameter. 

PS: For sure some is going to ask you to make it final, so....

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -56,12 +54,9 @@ public void publish(final Metrics... metrics) {
 
             final HttpResponse httpResponse = httpRequest.execute();
             httpResponse.onSuccess(() -> {
-                Logger logger = MobileCore.getLogger();
-                logger.debug(MessageFormat.format("Metrics response: {0}: {1}",
-                                httpResponse.getStatus(), httpResponse.stringBody()));
-                logger.debug("Metrics sent: " + json.toString());
+                listener.onPublishMetricsSuccess(httpResponse);
```
danielpassos: I think here the lib can check if it was an HTTP_OK (Status 200) and don't need to expose the response.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
josemigallas: 1. Listeners are Android's callbacks, like in UI components, it follows the same pattern.
2. The purpose is for the user to handle the outcome of the metrics service actions.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
josemigallas: No need really, only for the test it would be more useful.

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
josemigallas: I know but so do varargs (metrics), and that's mandatory. We would have to change it into a list of some sort.

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
josemigallas: Fair enough

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
secondsun: You are using the Listener like a callback in the metrics publisher which is causing the confusion.

secondsun:
@josemigallas Is the goal testing (IE unit/integration) or is the goal to expose results of a metrics push?  You keep saying testing, but there aren't any tests and you added it to the example app so I'm confused.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
```diff
@@ -4,9 +4,10 @@
 
     /**
      * Allows to publish metrics to external source
-     * 
+     *
+     * @param listener a callback listener to handle the result of a publication
      * @param metrics a array of metrics objects to publish
      */
-    void publish(Metrics... metrics);
+    void publish(MetricsPublisherListener listener, Metrics... metrics);
```
ziccardi: Parameters should be final

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
ziccardi: It is a `MetricsPublisherListener` object, we already know it is about Metric publishing. No need to call the method `onPublshMetricsSuccess`. Better call it `onPublished`.

I don't like `response` to be passed here, agree with @danielpassos 

Something that helps the receiver know what has been processed should be passed, otherwise the listener will receive the event but won't know what the event was about.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisherListener.java
```diff
@@ -0,0 +1,11 @@
...
+
+    void onPublishMetricsError(Exception error);
```
ziccardi: See previous comment

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -16,6 +16,7 @@
 
     private Metrics[] defaultMetrics;
     private MetricsPublisher publisher;
+    private MetricsPublisherListener listener;
```
ziccardi: Listener pattern allows attaching many listeners. Why limiting to just one?

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
ziccardi: Make parameter final

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
```diff
@@ -18,7 +19,7 @@ public LoggerMetricsPublisher(final Logger logger) {
...
         nonNull(metrics, "metrics");
         for (final Metrics m : metrics) {
```
ziccardi: This can be changed to 
```java
for (final Metrics m: nonNull(metrics, "metrics")) {
```

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -34,7 +32,7 @@ public NetworkMetricsPublisher(final Context context, final HttpRequest httpRequ
     }
 
     @Override
-    public void publish(final Metrics... metrics) {
+    public void publish(MetricsPublisherListener listener, final Metrics... metrics) {
```
ziccardi: Make parameter final

example/src/main/java/org/aerogear/mobile/example/ExampleApplication.java
```diff
@@ -1,11 +1,16 @@
...
 
 import org.aerogear.mobile.core.MobileCore;
+import org.aerogear.mobile.core.http.HttpResponse;
+import org.aerogear.mobile.core.metrics.MetricsPublisherListener;
 import org.aerogear.mobile.core.metrics.MetricsService;
 
-public class ExampleApplication extends Application {
+public class ExampleApplication extends Application implements MetricsPublisherListener {
 
     private MobileCore mobileCore;
```
ziccardi: Are these meant to be reinitialized? If that's not the case, they should be final

example/src/main/java/org/aerogear/mobile/example/ExampleApplication.java
```diff
@@ -35,4 +41,15 @@ public MetricsService getMetricsService() {
         return metricsService;
     }
 
+    @Override
+    public void onPublishMetricsSuccess(HttpResponse httpResponse) {
```
ziccardi: Make parameter final

example/src/main/java/org/aerogear/mobile/example/ExampleApplication.java
```diff
@@ -35,4 +41,15 @@ public MetricsService getMetricsService() {
...
+        String text = MessageFormat.format("Metrics response: {0}: {1}", httpResponse.getStatus(),
+                        httpResponse.stringBody());
+        Toast.makeText(this, text, Toast.LENGTH_SHORT).show();
+    }
+
+    @Override
+    public void onPublishMetricsError(Exception error) {
```
ziccardi: Make parameter final

wtrocki:
WOW: 
When you trying to get first contribution to Android SDK :)

screen shot 2018-03-13 at 1 33 31 pm

danielpassos:
@josemigallas I have moved the Auth `Callback` into the core module (https://github.com/aerogear/aerogear-android-sdk/pull/153). You should update the PR to use it

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -60,28 +61,22 @@ public void destroy() {}
...
 
     /**
      * Send metrics
      *
      * @param metrics Metrics to send
-     * @return this MetricsService instance
      */
-    public MetricsService publish(final Metrics... metrics) {
-        nonNull(metrics, "metrics");
+    public void publish(final Metrics[] metrics, final Callback callback) {
```
ziccardi: Add another
```java
public void publish(final Metrics... metrics) {
    publish(metrics, null);
}
```

so that callers who don't need to pass the callback are not forced to pass null

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
```diff
@@ -32,6 +32,7 @@ public synchronized void onExecuted(SecurityCheckResult result) {
 
     @Override
     public synchronized void onComplete() {
-        metricsService.publish(new SecurityCheckResultMetric(metricResults));
+        metricsService.publish(new SecurityCheckResultMetric[] {
```
ziccardi: If you add another publish method as suggested, this won't be needed

auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
```diff
@@ -132,7 +132,9 @@ public SecurityCheckResult checkAndSendMetric(final SecurityCheckType securityCh
     public SecurityCheckResult checkAndSendMetric(final SecurityCheck securityCheck,
                     @NonNull final MetricsService metricsService) {
         final SecurityCheckResult result = check(securityCheck);
-        nonNull(metricsService, "metricsService").publish(new SecurityCheckResultMetric(result));
+        nonNull(metricsService, "metricsService").publish(
```
ziccardi: see previous comment

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
```diff
@@ -1,12 +1,17 @@
...
 
     /**
      * Allows to publish metrics to external source
-     * 
+     *
      * @param metrics a array of metrics objects to publish
+     * @param callback callback of the publication
      */
-    void publish(Metrics... metrics);
+    void publish(Metrics[] metrics, @Nullable final Callback callback);
```
ziccardi: add another publish method here as suggested in the other comment

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
```diff
@@ -18,10 +21,13 @@ public LoggerMetricsPublisher(final Logger logger) {
     }
 
     @Override
-    public void publish(final Metrics... metrics) {
+    public void publish(final Metrics[] metrics, @Nullable final Callback callback) {
```
ziccardi: see other comments about this

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -34,7 +33,7 @@ public NetworkMetricsPublisher(final Context context, final HttpRequest httpRequ
     }
 
     @Override
-    public void publish(final Metrics... metrics) {
+    public void publish(final Metrics[] metrics, @Nullable final Callback callback) {
```
ziccardi: see other comments about this

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
josemigallas: Then, there would be a method with `Metrics... metrics` and another with `Metrics[] metrics`. Would that be OK?

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
josemigallas: Both in `MetricsService` and `MetricsPublisher`? But no one is going to implement that method, in `MetricsService` I simply added:
```
    public void publish(Metrics... metrics) {
        publish(metrics, null);
    }
```
And `MetricsService` don't implement that interface anyway.

ziccardi: You could add a `default implementation`. However, if you think it would not be useful or used at all, don't add it.

ziccardi: Make parameter final

ziccardi: Add `@NonNull` annotation

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -60,28 +61,32 @@ public void destroy() {}
...
+        publish(metrics, null);
+    }
+
+    /**
+     * Send metrics
+     *
+     * @param metrics Metrics to send
+     * @param callback callback of the publication
+     */
+    public void publish(final Metrics[] metrics, final Callback callback) {
```
ziccardi: add `@NonNull` annotation

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
ziccardi: Add `@nonNull` annotation

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -34,7 +36,7 @@ public NetworkMetricsPublisher(final Context context, final HttpRequest httpRequ
...
```
ziccardi: Add `@nonNull` annotation

josemigallas: According to the actual implementation, it is `@nullable`, because the publisher doesn't have another method without a callback

ziccardi: Yes, I meant add `@NonNull` annotation to the `metrics` array

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
```diff
@@ -1,12 +1,18 @@
...
 
     /**
      * Allows to publish metrics to external source
-     * 
+     *
      * @param metrics a array of metrics objects to publish
+     * @param callback callback of the publication
      */
-    void publish(Metrics... metrics);
+    void publish(@NonNull final Metrics[] metrics, @Nullable final Callback callback);
```
danielpassos: @ziccardi I have added `publish(Metrics... metrics);` to make easy to do `publish(new MyAlsoMetrics(), new MyNotSoAwesomeMetrics())`. Since it's not going to be used anymore, should we replace `Metrics[]` by a `List<Metrics>`?

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -61,27 +62,40 @@ public void destroy() {}
      * Send default metrics
      */
     public void sendAppAndDeviceMetrics() {
-        if (publisher == null) {
-            throw new IllegalStateException(
-                            "Make sure you have called configure or get this instance from MobileCore.getInstance()");
-        }
-        publisher.publish(defaultMetrics);
```
danielpassos: Why was the publisher check removed? What is going to happen if I do...

```java
MetricsService mService = new MetricsService();
mService.sendAppAndDeviceMetrics();
```

danielpassos:
@josemigallas LGTM. My only point is if we should replace `Metrics[]` by a `List<Metrics>`.

When I added it was to make easy to do `publish(new MyAlsoMetrics(), new MyNotSoAwesomeMetrics())`. Since it's not going to be used anymore, should we replace `Metrics[]` by a `List<Metrics>`?

danielpassos:
@josemigallas Can you fix the test?

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
josemigallas: I say we merge this once for all and after that we think about changing the array for collection.

danielpassos: np for me!

Issue #139, opened 2018-03-07
Allow retrieve Keycloak user attributes from the SDK


wei-lee:
Keycloak allows setting arbitrary attributes on users as described [here](https://www.keycloak.org/docs/3.3/server_admin/topics/users/attributes.html).

We should update the SDK to retrieve these attributes from the user profile and allow developers to access them.

TommyJ1994:
Implemented in https://github.com/aerogear/aerogear-android-sdk/pull/149

PR #149, opened 2018-03-12
Add Custom User Attributes Retrieval

TommyJ1994:
## Motivation
Add docs on how to configure Keycloak to set custom user attributes in the tokens. Create new function to retrieve custom attributes from the identity token. 

Related Issue: https://github.com/aerogear/aerogear-android-sdk/issues/139
Related Jira: https://issues.jboss.org/browse/AGDROID-785

## Description
Added documentation on how to get custom user attributes.
Add link to SSO docs as they it's not visible in any of the auth docs.
Add function to retrieve custom attributes on the identity token:

`String customUserAttribute = currentUser.getCustomAttribute("myCustomAttribute");`

## Progress

- [x] Add Docs https://github.com/aerogear/aerogear-android-sdk/commit/babf3d7afb7ffcf9f4c2b4b0650b5e90b93ab853 https://github.com/aerogear/aerogear-android-sdk/pull/149/commits/e55140e3f93005f8894ae81441858b02534b0658
- [x] Add function to get a custom user attribute https://github.com/aerogear/aerogear-android-sdk/pull/149/commits/0f8673330cc34cb9662c8d41e856c8fc91aedaee
- [x] Add tests https://github.com/aerogear/aerogear-android-sdk/pull/149/commits/784e7218e28207669c41d3c2ba95175bc891a0b2


wei-lee:
@TommyJ1994 is this change enough? Is there any changes required to the code?

For example, I think we should add a method on the User mobile to allow easily access the attribute?
e.g.
```
String attribute = user.getCustomAttribute(attributeName)
```

TommyJ1994:
@wei-lee Yeah I was thinking the same about providing some function to get a custom attribute, I was just worried about the attribute type handling mostly, but it looks like it shouldn't be an issue come to look at it now. I will work on adding this function now.

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -260,4 +267,34 @@ public String getRefreshToken() {
...
 
+
+    /**
+     * Returns the custom attribute.
+     *
+     * @param attribute the attribute name to retrieve
+     * @return the custom attribute.
+     */
+    @Override
+    public String getCustomAttribute(final String attribute) {
```
wojta: There is now way how to differentiate between attribute with value of empty string and attribute that doesn't exist. Shouldn't it be returning null or throwing exception? Maybe add another method, `hasCustomAttribute()`....

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -260,4 +267,34 @@ public String getRefreshToken() {
...
+        String attributeValue = "";
+        try {
+            String identityToken = getIdentityToken();
```
wojta: Is this identity token same for the whole session? Maybe we shouldn't parse it every time, but cache the result.

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -260,4 +267,34 @@ public String getRefreshToken() {
...
+
+    /**
+     * Returns the custom attribute.
+     *
+     * @param attribute the attribute name to retrieve
+     * @return the custom attribute.
+     */
+    @Override
+    public String getCustomAttribute(final String attribute) {
+        String attributeValue = "";
```
wei-lee: I think we should return null if the attribute doesn't exist?

wei-lee:
@wtrocki If you are working on supporting custom user attribute for ios, you can take a look at this PR.

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -260,4 +267,34 @@ public String getRefreshToken() {
...
+                JSONObject rawIdentityToken = new JSONObject(decoded);
+                if (rawIdentityToken.has(attribute)) {
+                    attributeValue = rawIdentityToken.getString(attribute);
+                } else {
+                    Log.d("Invalid Path", "No Object Exists in the JSON Path " + attribute);
+                }
+            } catch (JSONException e) {
+                Log.d("Error Getting Attribute", e.getMessage());
+            }
+        } catch (JoseException e) {
```
wtrocki: Made me laugh :) @josemigallas - JWT team have made exception for you :)

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
TommyJ1994: @wojta Updated in https://github.com/aerogear/aerogear-android-sdk/pull/149/commits/31234496cf87fb87aa5f2bff718eaf33c36944d3 to return null by default.

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
TommyJ1994: Yes, makes sense. Updated in https://github.com/aerogear/aerogear-android-sdk/pull/149/commits/31234496cf87fb87aa5f2bff718eaf33c36944d3

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
TommyJ1994: Hmm, I'm not sure if there is a need to cache it as we are really only Base64 decoding and parsing it as a JSON object.

wojta: I can live with it, as it won't be used to exchange large amounts of data.

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -260,4 +267,34 @@ public String getRefreshToken() {
...
+    @Override
+    public String getCustomAttribute(final String attribute) {
+        String attributeValue = null;
+        try {
+            String identityToken = getIdentityToken();
+            JsonWebSignature signature = new JsonWebSignature();
+            signature.setCompactSerialization(identityToken);
+            String decoded = signature.getUnverifiedPayload();
+            try {
+                JSONObject rawIdentityToken = new JSONObject(decoded);
```
wei-lee: it looks like keycloak supports a few types for the value: int, long, string and boolean. I think it might be better to add support to retrieve the value based on the types?

TommyJ1994: @wei-lee Do you mean like 4 functions like so?

```
string attr1 = user.getStringCustomAttribute(attributeName)
boolean attr2 = user.getBooleanCustomAttribute(attributeName)
int attr3 = user.getIntCustomAttribute(attributeName)
long attr4 = user.getLongCustomAttribute(attributeName)
```
I was thinking this might be a bit verbose initially, but if you think better typing would be better, I can update the PR.

wei-lee: yes, I think we should have those methods. They should all call to the same method to retrieve the value, just need to parse the value based on the expected types.

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -269,32 +269,110 @@ public String getRefreshToken() {
...
+     * @param attributeName the user attribute to retrieve
+     * @return the custom string attribute
+     */
+    public String getCustomStringAttribute(String attributeName) {
+        String attribute = null;
+        JSONObject jwt = getRawIdentityToken();
+
+        if (customAttributeExists(attributeName)) {
+            try {
+                attribute = jwt.getString(attributeName);
```
wei-lee: you can use `optString/optBoolean/optLong/optInt` methods. This way you don't need to catch the exceptions.

TommyJ1994: Ah, nice. Updated in https://github.com/aerogear/aerogear-android-sdk/pull/149/commits/2705672cc12346f6e67904120e68123420d069e1

PR #151, opened 2018-03-12
[WIP] Add Push

danielpassos:
## Motivation

JIRA: [AGDROID-777](https://issues.jboss.org/browse/AGDROID-777)

## Description

Migrate the old [AeroGear Android Push](https://github.com/aerogear/aerogear-android-push/) to the new SDK

## Progress

- [x] Register device on UPS
- [ ] Register the device on the variant topic
- [ ] Unregister device from up
- [ ] Refresh Firebase token


example/src/main/assets/push-config.json
```diff
@@ -0,0 +1,8 @@
+{
+    "pushServerURL": "https://testing.zeta.feedhenry.com/api/v2/ag-push/",
+    "android": {
+        "senderID": "847851272667",
+        "variantID": "518fd580-44b8-4a8c-b865-22300384800d",
+        "variantSecret": "a5ada26a-37bd-40f2-8f4a-66f846ffa395"
+    }
+}
```
secondsun: Is there an issue or JIRA to generate this config from mobile-cli?

example/src/main/java/org/aerogear/mobile/example/ExampleApplication.java
```diff
@@ -17,9 +21,36 @@ public void onCreate() {
         mobileCore = MobileCore.init(this);
         metricsService = mobileCore.getInstance(MetricsService.class);
 
+        sendDefaultMetrics();
```
secondsun: I wonder if we can make the default metrics sending part of MobileCore.init

push/src/androidTest/java/org/aerogear/mobile/push/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,26 @@
...
+import android.support.test.InstrumentationRegistry;
+import android.support.test.runner.AndroidJUnit4;
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+@RunWith(AndroidJUnit4.class)
+public class ExampleInstrumentedTest {
```
secondsun: Probably need a better class name.

example/src/main/assets/push-config.json
wtrocki: See: https://github.com/aerogear/proposals/pull/31

example/src/main/java/org/aerogear/mobile/example/ExampleApplication.java
wtrocki: See: https://github.com/aerogear/aerogear-app-metrics/pull/41

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,250 @@
...
+        return true;
+    }
+
+    @Override
+    public void destroy() {}
+
+    private void loadConfigJson() {
+        InputStream fileStream = null;
+
+        try {
```
secondsun: We can use try with resources finally !  So use try with resources.

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,250 @@
...
+
+    public void registerDevice(final Callback callback) {
+        registerDevice(new UnifiedPushConfig(), callback);
+    }
+
+    public void registerDevice(final UnifiedPushConfig unifiedPushConfig, final Callback callback) {
+
+        try {
+
+            JSONObject data = new JSONObject();
```
secondsun: I've seen JSONObjects constructed this way a lot.  We should probably look at making a utility class to make them easier to do

Idea.java
```java
JSONUtils.object(
  propery("deviceToken", FirebaseInstanceId.getInstance().getToken()),
  propery("deviceType", deviceType)
  //etc
).build();
```

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,250 @@
...
+
+            data.put("deviceToken", FirebaseInstanceId.getInstance().getToken());
+            data.put("deviceType", deviceType);
+            data.put("operatingSystem", operatingSystem);
+            data.put("osVersion", osVersion);
+            data.put("alias", unifiedPushConfig.getAlias());
+
+            final List<String> categories = unifiedPushConfig.getCategories();
+            if (categories != null && !categories.isEmpty()) {
```
secondsun: Looks like another case for a JSONUtils class.

```java
JSONArray jsonCategories = JSONUtils.array(catagories)
```

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,250 @@
...
+                for (String category : categories) {
+                    jsonCategories.put(category);
+                }
+                data.put("categories", jsonCategories);
+            }
+
+            String authHash = getHashedAuth(unifiedPushCredentials.getVariant(),
+                            unifiedPushCredentials.getSecret().toCharArray());
+
+            final HttpRequest httpRequest = core.getHttpLayer().newRequest();
```
secondsun: Can the Http operations be moved into their own method?

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,250 @@
...
+            httpRequest.addHeader("Authorization", authHash);
+            httpRequest.post(url + registryDeviceEndpoint, data.toString().getBytes());
+
+            final HttpResponse httpResponse = httpRequest.execute();
+            httpResponse.onSuccess(new Runnable() {
```
secondsun: the runnable could be a lambda

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,250 @@
...
+    private static final String PUSH_CONFIG_FILE_NAME = "push-config.json";
+    private static final String JSON_OBJECT = "android";
+    private static final String JSON_SENDER_ID = "senderID";
+    private static final String JSON_VARIANT_ID = "variantID";
+    private static final String JSON_VARIANT_SECRET = "variantSecret";
+
+    private static final String registryDeviceEndpoint = "rest/registry/device";
+
+    private static final List<MessageHandler> MAIN_THREAD_HANDLERS = new ArrayList<>();
+    private static final List<MessageHandler> BACKGROUND_THREAD_HANDLERS = new ArrayList<>();
```
secondsun: Do these still need to be static?

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,250 @@
...
+                        default:
+                            callback.onError(new HttpException());
+                            break;
+                    }
+                }
+            });
+
+        } catch (JSONException e) {
+            MobileCore.getLogger().error(e.getMessage(), e);
+            callback.onSuccess(e);
```
secondsun: Should onSuccess be called here?

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public String getAlias() {
+        return alias;
+    }
+
+    public void setAlias(String alias) {
+        this.alias = alias;
+    }
+
+    public List<String> getCategories() {
```
secondsun: Should we do a null check here and return an empty list if there are no categories?

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: Nopz, this is a poor copy and paste :tired_face:

push/src/main/java/org/aerogear/mobile/push/fcm/AeroGearFirebaseMessagingService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.aerogear.mobile.push.PushService;
+
+public class AeroGearFirebaseMessagingService extends FirebaseMessagingService {
+
+    @Override
+    public void onMessageReceived(RemoteMessage remoteMessage) {
+        super.onMessageReceived(remoteMessage);
+
+        PushService.notifyHandlers(remoteMessage.getData());
```
secondsun: Yeah so this is a good upside to the handlers being static.  We can send messages to them without needing mobile core when the message comes in and that speeds up handling.

It may be a good point of discussion for how we want to allow developers to control the various ways apps can start up.  For instance if a message comes to the app while it is killed we probably don't want to send metrics and such if all we are doing is refreshing the app...

example/src/main/java/org/aerogear/mobile/example/ExampleApplication.java
danielpassos: There is a discussion about it on ML https://groups.google.com/forum/#!topic/aerogear/QZn65_qglSU

push/src/test/java/org/aerogear/mobile/push/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
...
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+public class ExampleUnitTest {
```
secondsun: This test probably needs to be removed

push/src/androidTest/java/org/aerogear/mobile/push/ExampleInstrumentedTest.java
danielpassos: I'm going to kill it since we don't wanna have any instrumented test for now ;)

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: It's used for `notifyHandlers` and the method is static

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,339 @@
...
+                    break;
+                default:
+                    callback.onError(new HttpException());
+                    break;
+            }
+        });
+
+    }
+
+    private String getHashedAuth(String username, char[] password) {
```
ziccardi: Make params final

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
danielpassos: Done!

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,313 @@
...
+
+
+    /**
+     * When a push message is received, all main thread handlers will be notified on the main(UI)
+     * thread. This is very convenient for Activities and Fragments.
+     *
+     * @param handler a handler to added to the list of handlers to be notified.
+     */
+    public static void registerMainThreadHandler(MessageHandler handler) {
+        MAIN_THREAD_HANDLERS.add(handler);
```
ziccardi: If 2 threads calls this method, we will get a ConcurrentModificationException. If threading is not an issue, ignore this, otherwise consider using a synchronised collection.

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+
+    /**
+     * When a push message is received, all main thread handlers will be notified on the main(UI)
+     * thread. This is very convenient for Activities and Fragments.
+     *
+     * @param handler a handler to added to the list of handlers to be notified.
+     */
+    public static void registerMainThreadHandler(MessageHandler handler) {
```
ziccardi: Make param final

ziccardi: add notnull check

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /**
+     * When a push message is received, all background thread handlers will be notified on a non UI
+     * thread. This should be used by classes which need to update internal state or preform some
+     * action which doesn't change the UI.
+     *
+     * @param handler a handler to added to the list of handlers to be notified.
+     */
+    public static void registerBackgroundThreadHandler(MessageHandler handler) {
```
ziccardi: see previous comments

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /**
+     * This will remove the given handler from the collection of main thread handlers. This MUST be
+     * called when a Fragment or activity is backgrounded via onPause.
+     *
+     * @param handler a handler to be removed to the list of handlers
+     */
+    public static void unregisterMainThreadHandler(MessageHandler handler) {
+        MAIN_THREAD_HANDLERS.remove(handler);
```
ziccardi: see previous comments

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /**
+     * This will remove the given handler from the collection of background thread handlers.
+     *
+     * @param handler a handler to be removed to the list of handlers
+     */
+    public static void unregisterBackgroundThreadHandler(MessageHandler handler) {
+        BACKGROUND_THREAD_HANDLERS.remove(handler);
```
ziccardi: see previous comments

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /**
+     * This will deliver an message to all registered handlers.
+     *
+     * @param message the message to pass
+     */
+    public static void notifyHandlers(final Context context, final Map<String, String> message) {
```
ziccardi: make params final

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,313 @@
...
+
+        if (BACKGROUND_THREAD_HANDLERS.isEmpty() && MAIN_THREAD_HANDLERS.isEmpty()
+                        && defaultHandler != null) {
+            new Thread(new Runnable() {
+                public void run() {
+                    defaultHandler.onMessage(context, message);
```
ziccardi: if the caller will change the content of the map during this thread execution, we could get a threading issue. Consider copying to passed in map before working with it.

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
```diff
@@ -0,0 +1,30 @@
...
+        if (categories == null) {
+            return new ArrayList<>();
+        }
+        return categories;
```
ziccardi: Never return a reference to a mutable internal variable. Consider copying it or returning `Collection.unmodifiableList(categories);`

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public void setCategories(List<String> categories) {
+        this.categories = categories;
```
ziccardi: Never assign a mutable to an internal variable to avoid losing control about its content.
Consider copying the list.

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: Map should not be changed

ziccardi: Yes, but you don't have control over that, because the caller has reference to the Map.
To avoid issues, it is a good practice to copy its content so that you don't have to worry about what the caller will do with the Map.

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,335 @@
...
+                    callback.onError(new HttpException());
+                    break;
+            }
+        });
+
+    }
+
+    private String getHashedAuth(String username, char[] password) {
+        StringBuilder headerValueBuilder = new StringBuilder("Basic").append(" ");
+        String unhashedCredentials = username + ":" + String.valueOf(password);
```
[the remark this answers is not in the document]
danielpassos: Should I really care about this since this is coming from a JSON file inside of the app?

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: Not sure if I got you here! `MAIN_THREAD_HANDLERS` is a _static_ `List` of `MessageHandler`

push/src/main/java/org/aerogear/mobile/push/PushService.java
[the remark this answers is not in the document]
ziccardi: Don't think so ;) I wrote that as a best practice, but if in this case the password is inside a JSON file inside the app, I guess there is no use about discussing password security

push/src/main/java/org/aerogear/mobile/push/PushService.java
ziccardi: Exactly. Of two different threads will call `registerMainThreadHandler` at the same time, we will get a `ConcurrentModificationException` here, unless MAIN_THREAD_HANDLERS is a synchronised List

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: Sorry I got your best practice point here. My point is this care I really don't care if anyone changes it because the Services responsibility is only to pass it along and go home. I mean the lib is not going to use it anywhere anymore. Anyway, I'm not against to do a defensive copy of this.

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: Ok, not I got your point, sorry!

danielpassos: I can't be wrong but in theory I can't have a static synchronized list. Should I solve it doing:

```java
public static void unregisterBackgroundThreadHandler(final MessageHandler handler) {
    nonNull(handler, "handle");
    synchronized (PushService.class) {
        BACKGROUND_THREAD_HANDLERS.remove(handler);
    }
}
```

push/src/main/java/org/aerogear/mobile/push/PushService.java
ziccardi: Yes, but what would happen in the for loop we have just a few lines below? We could end up passing e different `message` to each `handler` (line 262)

push/src/main/java/org/aerogear/mobile/push/PushService.java
ziccardi: I'm not sure I got you.
I think you could do:

```java
private static final List<MessageHandler> MAIN_THREAD_HANDLERS = Collections.synchronizedList(new ArrayList<>());
```

the same for the background handlers.

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,318 @@
...
+
+    @Override
+    public void configure(MobileCore core, ServiceConfiguration serviceConfiguration) {
+        this.core = core;
+        this.url = serviceConfiguration.getUrl();
+        getDefaultHandler(core.getContext());
+
+        try {
+            JSONObject android =
+                            new JSONObject(serviceConfiguration.getProperties().get(JSON_OBJECT));
```
pb82: instead of `JSON_OBJECT` would `ANDROID_CONFIG_KEY` be a better name?

secondsun:
@ziccardi I'm actually -1 on all of the finals for parameters.  It doesn't make the code easier to read, since most values are references it doesn't help with immutability, the newer java compilers check for final automatically if you need it, etc etc.  All the word does at this point is add noise to the code, and unless it makes something more clear it shouldn't be used.

Finals on classes, fields, etc are a different beast.

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: Hrm, I was about to say I can't have `static synchronized` field. Nevermind!

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,318 @@
...
+     * This will deliver an message to all registered handlers.
+     *
+     * @param message the message to pass
+     */
+    public static void notifyHandlers(final Context context, final Map<String, String> message) {
+        nonNull(context, "context");
+
+        if (BACKGROUND_THREAD_HANDLERS.isEmpty() && MAIN_THREAD_HANDLERS.isEmpty()
+                        && defaultHandler != null) {
+            new Thread(new Runnable() {
```
secondsun: These new Threads could be appexecutors instead.

ziccardi:
@secondsun I see your point. However making parameters final is a best practice since it blocks the programmer from changing the parameter value (bad practice) and making the code less readable. It has nothing to do with immutability.
I don't think `final` parameters makes the code less readable, however it can be just my opinion.
@matzew WDY?

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: I have updated it to:

```
private static final String JSON_ANDROID_CONFIG_KEY = "android";
private static final String JSON_VARIANT_ID_KEY = "variantId";
private static final String JSON_VARIANT_SECRET_KEY = "variantSecret";
private static final String JSON_SENDER_ID_KEY = "senderId";
```

push/src/main/java/org/aerogear/mobile/push/PushService.java
secondsun: These are synchyronized already https://github.com/aerogear/aerogear-android-sdk/pull/151/files#diff-24684cb8332f25e88fee7344b4b376a7R46 don't double synchronize.

secondsun:
@ziccardi Final doesn't add anything for the end user, for us, for the compiler, etc.  We could write a script that just adds it everywhere and nobody(end users, developers, us, the compiler, etc) would notice/care.  Right now all this discussion is is bikeshedding and distracting from better discussions.

ziccardi:
@secondsun I don't agree, but, again, it can be just my opinion.
Tomorrow another developer could add some code and change the parameter value. If it is final, I'm sure that won't happen. That is the added value.

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: Ok, I have added `Collections.unmodifiableMap(message)` so!

secondsun:
@ziccardi It is an object reference, they can already change the value even if it isn't final.

Otherwise you end up with code like this

```java
myFunction (final List<String> myValues) {
  List<String> myValuesCopy = new Arraylist<>(myValues);
}
```

instead of 
```java
myFunction ( List<String> myValues) {
   myValues = new Arraylist<>(myValues);
}
```

The first example is far, FAR more verbose as you are now having to obfuscate your intentions.  Meanwhile, there is nothing in the first example preventing your from changing the value of the list which defeats the purpose of your intention.

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,313 @@
...
+            new AppExecutors().mainThread().execute(() -> defaultHandler.onMessage(context,
+                            Collections.unmodifiableMap(message)));
+        } else {
+
+            for (final MessageHandler handler : BACKGROUND_THREAD_HANDLERS) {
+                new AppExecutors().mainThread().execute(() -> handler.onMessage(context,
+                                Collections.unmodifiableMap(message)));
+            }
+
+            Looper main = Looper.getMainLooper();
```
danielpassos: @secondsun Should we also replace `Looper` by `AppExecutors`

secondsun: @danielpassos Yes

ziccardi:
@secondsun that was not the point, but I think this is not the right place to discuss about that.
The point is that with 

```java
myFunction (final List<String> myValues) {
  List<String> myValuesCopy = new Arraylist<>(myValues);
}
```

you can't do `myValues = new ArrayList(.....`

matzew:
@secondsun @ziccardi 

> I don't think final parameters makes the code less readable,

Of course, the `final` on the argument has just internal affect - it does not make the entire thing immutable, due to lack of decent support in java.  java-- 

However, IMO it helps to avoid _method argument overriding_, which can cause stupid bugs... 

My take here is... _if_ the passed in argument is not designed to be overridden, I'd sponsor it a `final`.. if _method argument overriding_ is desired (is it really?), It's OK to not use it ;-)

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+     * When a push message is received, all main thread handlers will be notified on the main(UI)
+     * thread. This is very convenient for Activities and Fragments.
+     *
+     * @param handler a handler to added to the list of handlers to be notified.
+     */
+    public static void registerMainThreadHandler(final MessageHandler handler) {
+        nonNull(handler, "handle");
+        MAIN_THREAD_HANDLERS.add(handler);
```
ziccardi: This can be changed to 
```java
MAIN_THREAD_HANDLERS.add(nonNull(handler, "handler"));
```

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     * When a push message is received, all background thread handlers will be notified on a non UI
+     * thread. This should be used by classes which need to update internal state or preform some
+     * action which doesn't change the UI.
+     *
+     * @param handler a handler to added to the list of handlers to be notified.
+     */
+    public static void registerBackgroundThreadHandler(final MessageHandler handler) {
+        nonNull(handler, "handle");
+        BACKGROUND_THREAD_HANDLERS.add(handler);
```
ziccardi: See previous comment

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+     * This will remove the given handler from the collection of main thread handlers. This MUST be
+     * called when a Fragment or activity is backgrounded via onPause.
+     *
+     * @param handler a handler to be removed to the list of handlers
+     */
+    public static void unregisterMainThreadHandler(final MessageHandler handler) {
+        nonNull(handler, "handle");
+        MAIN_THREAD_HANDLERS.remove(handler);
```
ziccardi: see previous comment

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /**
+     * This will remove the given handler from the collection of background thread handlers.
+     *
+     * @param handler a handler to be removed to the list of handlers
+     */
+    public static void unregisterBackgroundThreadHandler(final MessageHandler handler) {
+        nonNull(handler, "handle");
+        BACKGROUND_THREAD_HANDLERS.remove(handler);
```
ziccardi: see previous comment

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public List<String> getCategories() {
+        if (categories == null) {
+            return new ArrayList<>();
+        }
+        return Collections.unmodifiableList(categories);
+    }
+
+    public void setCategories(List<String> categories) {
+        Collections.copy(this.categories, categories);
```
ziccardi: if `this.categories` is null or is not big enough to contain the passed in list, this will throw an exception.
Better change this to 
```java
this.categories = new ArrayList(categories)
```

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
```diff
@@ -0,0 +1,31 @@
+package org.aerogear.mobile.push;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public final class UnifiedPushConfig {
+
+    private String alias;
+    private List<String> categories;
```
ziccardi: I would initialize this to `Collections.emptyList()` so that no check that this is null is needed.

example/build.gradle
```diff
@@ -41,12 +41,16 @@ ext {
 dependencies {
     implementation project(path: ":core")
     implementation project(path: ":auth")
+    implementation project(path: ":push")
     implementation "com.android.support:appcompat-v7:$android_support_version"
     implementation "com.android.support:support-v4:$android_support_version"
     implementation "com.android.support:design:$android_support_version"
     implementation "com.android.support.constraint:constraint-layout:1.0.2"
+    implementation 'com.google.firebase:firebase-messaging:11.8.0'
```
secondsun: Should the firebase dependency be in the bom?

push/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+
+}
+
+dependencies {
+    implementation project(path: ':core')
+    implementation 'com.android.support:appcompat-v7:26.1.0'
+    implementation 'com.google.firebase:firebase-messaging:11.8.0'
```
secondsun: Can these dependencies be moved to the bom?

push/src/androidTest/java/org/aerogear/mobile/push/ExampleInstrumentedTest.java
secondsun: Still needs a kill :)

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -0,0 +1,315 @@
...
+                            serviceConfiguration.getProperties().get(JSON_ANDROID_CONFIG_KEY));
+
+            unifiedPushCredentials = new UnifiedPushCredentials();
+            unifiedPushCredentials.setVariant(android.getString(JSON_VARIANT_ID_KEY));
+            unifiedPushCredentials.setSecret(android.getString(JSON_VARIANT_SECRET_KEY));
+            unifiedPushCredentials.setSender(android.getString(JSON_SENDER_ID_KEY));
+
+        } catch (JSONException e) {
+            MobileCore.getLogger().error(e.getMessage(), e);
+            // Should never happen
```
secondsun: Should this rethrow as a runtime exception?
I think there may be a exception type in the project we can use as well.

push/src/main/java/org/aerogear/mobile/push/fcm/AeroGearFirebaseInstanceIdService.java
```diff
@@ -0,0 +1,6 @@
+package org.aerogear.mobile.push.fcm;
+
+import com.google.firebase.iid.FirebaseInstanceIdService;
+
+public class AeroGearFirebaseInstanceIdService extends FirebaseInstanceIdService {
```
secondsun: I forget, is there supposed to be a body in this class?

danielpassos: Yes, I'm going to send a new PR about refresh token to not make it so, big and complicated!

example/build.gradle
danielpassos: Not in the sample app only on the library!

push/build.gradle
danielpassos: It's waiting for you ;)

https://github.com/aerogear/aerogear-parent/pull/83

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos:
```java
throw new ConfigurationNotFoundException("An error occurred while trying to load the push config");
```

wdyt?

example/build.gradle
danielpassos: I should remove this, push already have this as a dependency!

push/src/main/java/org/aerogear/mobile/push/PushService.java
matzew: I like this too

danielpassos:
Waiting for https://github.com/aerogear/aerogear-parent/pull/83 be released!

danielpassos:
@secondsun rebased using BOM

PR #152, opened 2018-03-14
[WIP] Aerogear 2223 - parse mobile-services.json and add cert values to httpservice requests

ciaranRoche:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2223

## Description

We want the user to be able to easily configure the certificates to be used to perform requests, without the need to write code.

## Progress

- [x] Parse JSON
- [x] Pin Certs with okHttp
- [ ] Unit Test
- [ ] JavaDoc

## To Verify
- Run Example App
- Navigate to Auth and click Authenticate
- Chrome Browser should boot and you should see keycloak log in
- Edit keycloak hash value in mobile-services.json file in the example app
- Restart Example App
- Navigate to Auth and click Authenticate
- Chrome Browser will not boot, and a `SSLPeerUnverifiedException` is logged


example/src/main/assets/mobile-services.json
```diff
@@ -26,6 +26,12 @@
             "config": {
             }
         }
+    ],
+    "https": [
+        {
+            "host": "https://keycloak-myproject.192.168.37.1.nip.io",
```
TommyJ1994: A hostname doesn't include the protocol in the address, so this might causing issues in performing pinning. Pinning libraries expect the format `keycloak-myproject.192.168.37.1.nip.io`. I think this might be more related to https://github.com/aerogear/mobile-cli/pull/71 maybe?

aidenkeating: @TommyJ1994 Yep this is an issue with the mobile-cli side, will be pushing up the change along with some other stuff in a little while.

TommyJ1994:
I'm not sure if pinning is being performed too, as requests seem to succeed when they should fail.

```
03-14 10:29:48.282 14000-14000/? I/System.out: >>>>>>>>>>>>>>> security.feedhenry.org sha256/o88ICr4sMHUiobmszVwBOyDxsaDzlhwwEbslc2Onm00=

```

image

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreJsonParser.java
```diff
@@ -19,12 +20,17 @@
  */
 public class MobileCoreJsonParser {
 
-    private final Map<String, ServiceConfiguration> values = new TreeMap<>();
+    private static final Map<String, ServiceConfiguration> values = new TreeMap<>();
+    private static final Map<String, String> hashes = new HashMap<>();
+
 
     private MobileCoreJsonParser(final InputStream jsonStream) throws IOException, JSONException {
```
ziccardi: Parser should have a `parse` method that returns a parse result.
If the parsing is done inside the constructor, then this is not a parser, but an object that is built from a InputStream.

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreJsonParser.java
```diff
@@ -69,18 +75,42 @@ private void parseConfigObject(final JSONObject jsonObject) throws JSONException
...
-     * @return A map of ServiceConfigs mapped by their name.
      * @throws IOException if reading the stream fails
      * @throws JSONException if the json document is malformed
      */
-    public static Map<String, ServiceConfiguration> parse(final InputStream jsonStream)
+    public static void parse(final InputStream jsonStream)
                     throws IOException, JSONException {
-        MobileCoreJsonParser parser = new MobileCoreJsonParser(jsonStream);
-        return parser.values;
+        new MobileCoreJsonParser(jsonStream);
```
ziccardi: See my previous comment: here you must instantiate and throw the object, because this is a malformed parser (no parse method defined)

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreJsonParser.java
```diff
@@ -69,18 +75,42 @@ private void parseConfigObject(final JSONObject jsonObject) throws JSONException
...
+    }
+
+    public static Map<String, ServiceConfiguration> getServicesConfig(){
+        return values;
```
ziccardi: Mutable should never be returned outside of the class.
If you return it here, the caller will be able to change the content of the `values` map: what is the use of making it private if you have no control of how and when it gets changed?
Moreover, it won't be thread safe.
Consider returning  `Collections.unmodifiedMap(values)`

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreJsonParser.java
```diff
@@ -69,18 +75,42 @@ private void parseConfigObject(final JSONObject jsonObject) throws JSONException
...
+    }
+
+    public static Map<String, String> getCertificatePinningHashes(){
+        return hashes;
```
ziccardi: see previous comment

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreJsonParser.java
```diff
@@ -19,12 +20,17 @@
  */
 public class MobileCoreJsonParser {
 
-    private final Map<String, ServiceConfiguration> values = new TreeMap<>();
+    private static final Map<String, ServiceConfiguration> values = new TreeMap<>();
+    private static final Map<String, String> hashes = new HashMap<>();
```
ziccardi: With this change, this class has no state anymore. Consider removing the constructor and making all methods static.

ciaranRoche: Just getting to you comments now, cheers for the input.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -64,7 +66,10 @@ private MobileCore(final Context context, final Options options)
 
         // -- Parse JSON config file
         try (final InputStream configStream = context.getAssets().open(configFileName)) {
-            this.servicesConfig = MobileCoreJsonParser.parse(configStream);
+            MobileCoreJsonParser.parse(configStream);
```
ziccardi: If you want to return the objects from `getXXX` methods, they should not be static variables: if you call the parse method 2 times, they will be overwritten.
So, if you want the `parse` method to be static, you should return an object with the result of the parsing.
If you don't want to return the results from the `parse` method but you want to return them with the `getXXX` methods, then make both the parse method, the instance variable and the `getXXX` methods not static.

example/src/main/assets/mobile-services.json
```diff
@@ -26,6 +26,12 @@
             "config": {
             }
         }
+    ],
+    "https": [
+        {
+            "host": "security.feedhenry.org",
+            "certificateHash": "292B039CAC326C7A81450C7C21F4235C6568DCF9F2A74FBB7E977318462DCABD="
```
TommyJ1994: This needs to be a Base64 encoded Subject Public Key Information fingerprint to work with OkHTTP's pinner. The mobile-cli should output `trENjoQnbWupnAtu1/WagBE0RgJ+p7ke2ppWML8vAl0=` for `security.feedhenry.org`.

ciaranRoche:
@TommyJ1994 yeah I know, I am still testing this, I just pushed the changes that I had made.
It is a weird one despite the cert been pinned it is still successfully completing a TLS handshake and allowing us to sign in a user. I am still digging into this.

wei-lee:
@ciaranRoche @TommyJ1994 I don't think AppAuth is configured to use the OkHttp provided by us though. It is still using the default HttpClient provided by itself and I think that's why it is still working. We can either:

1. check if we can make AppAuth use the OkHttp client we have created, or
2. Perform a http request to the Keycloak server before invoking the requests to make sure the the right cert is used.

example/src/main/assets/mobile-services.json
aidenkeating: @TommyJ1994 The actual output of the Mobile CLI for `security.feedhenry.org` is.

```json
...
"https": [
    {
        "host": "security.feedhenry.org",
        "certificateHash": "trENjoQnbWupnAtu1/WagBE0RgJ+p7ke2ppWML8vAl0="
    }
]
...
```

so the CLI _should_ not cause any issues here when we're actually using it.

aidenkeating:
@ciaranRoche @wei-lee We should be able to use the `setConnectionBuilder()` function in appauth for adding cert pinning to appauth (https://github.com/openid/AppAuth-Android/blob/fddab98931a4f256872ea4243ebbc1955cbdbd1f/README.md#customing-the-connection-builder-for-http-requests).

TommyJ1994:
Its bad practice to have multiple pinning techniques and implementations. Cert pinning implementations should be centralized. As suggested a few times before, if we use the Network Security Configuration, it doesn't matter what OkHTTP libraries we use or what OkHTTP instance are used by libraries we use (like AppAuth), they still obey whatever is in this XML file in a easy and centralized manner for end users. AppAuth and our OkHTTP instance in the core module work out of the box with Network Security Configuration file if you just add the hostname + the SPKI hash.

AppAuth has some built-in support for pinning, but for the next library we want to use, we will hit the same problem. The Network Security Configuration is a catch-all.

aidenkeating:
@TommyJ1994 I definitely agree with you, was thinking something similar. Looking at the approach I commented above it doesn't seem to be elegant at all. I'd be a fan of trying your suggested approach and seeing how we can keep the implementation of cert pinning simple to use for the end user _and_ make it simple to maintain for ourselves.

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
```diff
@@ -204,6 +204,10 @@ public void init(final Context context,
                             "configure method must be called before the init method");
         }
 
+        CertificatePinningCheck pinningCheck =
+                        new CertificatePinningCheck(this.mobileCore.getHttpLayer());
+        pinningCheck.check(this.serviceConfiguration.getUrl());
```
secondsun: Can this check be done in an earlier step of the build configuration process?  (For example when the mobile-services.json file is parsed).

Also what does cert pinning check do and how is it supposed to be used?

auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
```diff
@@ -60,12 +62,14 @@
...
      * @param jwksManager {@link JwksManager}
+     * @param certificatePinningCheck {@link CertificatePinningCheck}
      */
     public OIDCAuthenticatorImpl(final ServiceConfiguration serviceConfiguration,
                     final AuthServiceConfiguration authServiceConfiguration,
                     final AuthStateManager authStateManager,
                     final AuthorizationServiceFactory authorizationServiceFactory,
-                    final JwksManager jwksManager) {
+                    final JwksManager jwksManager,
+                    final CertificatePinningCheck certificatePinningCheck) {
```
secondsun: There are getting to be a LOT of parameters on this constructor.  This is seeming like at best a good candidate to adopt a builder or better, or at worst a code smell that it is taking on too many responsibilities.

auth/src/main/java/org/aerogear/mobile/auth/utils/CertificatePinningCheck.java
```diff
@@ -0,0 +1,32 @@
+package org.aerogear.mobile.auth.utils;
+
+import org.aerogear.mobile.core.http.HttpRequest;
+import org.aerogear.mobile.core.http.HttpResponse;
+import org.aerogear.mobile.core.http.HttpServiceModule;
+
+
+public class CertificatePinningCheck {
```
secondsun: Javadocs, jav
image

auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
```diff
@@ -96,11 +102,12 @@ public void authenticate(final AuthenticateOptions authOptions,
     // Authentication code
     private void performAuthRequest(final Activity fromActivity, final int resultCode) {
         nonNull(fromActivity, "fromActivity");
-
+        if (this.certificatePinningCheck.getError() != null) {
```
secondsun: The check is performed very far from this code.  This exception should probably be closer to where the check is originally made.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -76,11 +81,14 @@ private MobileCore(final Context context, final Options options)
         // -- Setting default http layer
         if (options.httpServiceModule == null) {
             OkHttpClient.Builder builder = new OkHttpClient.Builder();
+
+            OkHttpCertificatePinning certificatePinning = new OkHttpCertificatePinning(httpsConfig);
+            builder.certificatePinner(certificatePinning.pinCertificates());
```
secondsun: What happens if there are no pins defined in the mobile-services.json file?

core/src/main/java/org/aerogear/mobile/core/configuration/HttpsConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public String getHostName() {
+        return hostName;
+    }
+
+    public String getCertificateHash() {
+        return certificateHash;
+    }
+
+    public static Builder newHashConfiguration() {
```
secondsun: newHttpsConfigurationBuilder would be a better name.  Or just newBuilder()

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreJsonConfig.java
```diff
@@ -17,14 +19,19 @@
  * This class is responsible for consuming a reader and producing a tree of config values to be
  * consumed by modules.
  */
-public class MobileCoreJsonParser {
+public class MobileCoreJsonConfig {
```
secondsun: Good catch.  This class should stay MobileCoreJsonParser, but the parse method should return a MobileCoreJsonConfig.  The static parsing methods should stay in the parser class, the parser class should become final, and the config class should be extracted from this class to more clearly separate the concerns

auth/src/main/java/org/aerogear/mobile/auth/utils/CertificatePinningCheck.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public void check(final String url) {
+        HttpRequest request = httpModule.newRequest();
+        request.get(url);
+        HttpResponse response = request.execute();
+        response.onError(() -> this.error = response.getError());
+        response.onComplete(() -> this.isComplete = true);
+    }
+
+    public boolean isComplete() {
```
secondsun: IntelliJ is complaining that this method isn't used so be sure to include it in your tests.

auth/src/main/java/org/aerogear/mobile/auth/AuthService.java
wei-lee: I don't think the check should happen at this point. Also if there are some other errors (e.g. the device is offline at the time when the check is happening), the login will not work and there is no way to reset it unless the app is restarted. I don't think that is the expected behaviour. If I remember correctly, this check is only required for the login request, right before the browser is being opened. So I think this should be done right before the tab is opened.

auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
wei-lee: As I commented above, it is possible for user not being to able to login at all due to network issues. I don't think that's right.

aidenkeating:
@ciaranRoche Closing this for https://github.com/aerogear/aerogear-android-sdk/pull/171

PR #153, opened 2018-03-14
Move auth Callback to the core module

danielpassos:
There is a Callback class in the Auth module and it should be reused (for now) in the other modules, so it should be in Core instead of in Auth module.


TommyJ1994:
@danielpassos I'm guessing this affects the import statements on the example app and the [Android Mobile Security Template](https://github.com/feedhenry/mobile-security-android-template)? Just checking as I'll make a note to update the mobile security template once a new SDK version is published with these changes to make sure nothing breaks.

secondsun:
LGTM

Issue #154, opened 2018-03-14
Allow Custom OIDC Scopes to be Passed to the Authenticator


TommyJ1994:
## Expected Behavior
I can choose what OIDC scopes are passed as part of the authentication request.

## Actual Behavior
Only the hardcoded `"openid"` scope is defined. See [here](https://github.com/aerogear/aerogear-android-sdk/blob/455dc6032f9fa7cc3f996f6d32c5aa62fb507114/auth/src/main/java/org/aerogear/mobile/auth/authenticator/AuthorizationServiceFactory.java#L25). This also prevents OAuth support.

### Enhancements
Allows custom scopes to be defined as per [OIDC v1.0 Scopes](https://openid.net/specs/openid-connect-basic-1_0.html#Scopes)

TommyJ1994:
Implemented in https://github.com/aerogear/aerogear-android-sdk/pull/155

PR #155, opened 2018-03-15
Allow Custom Scopes to be Passed, Allow OAuth

TommyJ1994:
## Motivation
Detail in https://github.com/aerogear/aerogear-android-sdk/issues/154

## Description
Allow custom scopes to be passed in the auth request. Currently this is hardcoded to `openid` by default. We should let the developer decide what scopes they want, and fallback to `openid` if they don't define ones.

## Progress

- [x] Allow custom scopes to be set in the builder.

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/authenticator/AuthorizationServiceFactory.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/authenticator/AuthorizationServiceFactory.java
@@ -22,8 +22,6 @@ import net.openid.appauth.browser.VersionedBrowserMatcher;
  */
 public class AuthorizationServiceFactory {
 
-    private final static String SCOPE_OPENID = "openid";
-
     private final Context appContext;
 
     /**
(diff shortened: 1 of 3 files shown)


wei-lee:
@TommyJ1994 actually, if there are only a limit number of scopes that OpenID supports, we should probably use enums to list them in the SDK.

TommyJ1994:
@wei-lee Its going to be a lot more common to write your own scopes - eg. [GitHub's Scopes](https://developer.github.com/apps/building-oauth-apps/scopes-for-oauth-apps/#available-scopes) or [Google's IDP Scopes](https://developers.google.com/identity/protocols/googlescopes). OpenID Connect has only a couple - 2 of which are really only relevant to Open ID (`openid, offline_access`), the rest are for OAuth. I'm ok with enums, but for those familiar with OAuth scopes, i think its quicker to write the built in ones yourself as its the same format used for URL, so you can easily copy the String of scopes into Curl requests etc for testing.

wei-lee:
@TommyJ1994 never mind. I saw the OpenID spec only listed a few scopes, but there are a lot more. Using string is fine.

PR #158, opened 2018-03-15
Aerogear 2333 rx refactor [WIP]

secondsun:
## Motivation

Right now we use a combination of RxStyle APIS (HttpRequest and Response) and Callbacks (Callback).  This PR seeks to unify those under a more thought out reactive style API.

JIRA: https://issues.jboss.org/browse/AGDROID-2333

## Description

This PR contains a Reactive API that is used for attaching multiple handlers to a request that emits a single calculated value when they are attached.  This is similar to a Single Observable from RxJava, but it is not the same.

## Progress

- [x] Request and Responder Interfaces
- [x] Synchronous Implementation
- [x] Asynchronous and Threading configurable Implementation
- [x] Ability to Cancel Requests
- [x] Ability to Cache Requests
- [x] Ability to Detach a Responder from a request
- [x] Unit Tests for all of the above
- [ ] Replace HttpRequest and HttpResponse
- [ ] Replace Callbacks


core/src/main/java/org/aerogear/mobile/core/reactive/CacheRequest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private T cachedResult;
+    private Exception cachedException;
+
+    private List<AtomicReference<Responder<T>>> awaitingResponders =
+                    Collections.synchronizedList(new ArrayList<>());
+
+    public CacheRequest(Request<T> delegateTo) {
+        nonNull(delegateTo, "delegateTo");
+        this.delegateTo = delegateTo;
```
ziccardi: this can be changed to `this.delegateTo = nonNull(delegateTo, "delegateTo")`

core/src/main/java/org/aerogear/mobile/core/reactive/CallableRequest.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+public final class CallableRequest<T> extends AbstractRequest<T> {
+
+    private static final String ERROR_TAG = "UNCAUGHT_EXCEPTION";
+    private final Callable<T> callable;
+    private Thread callableThread = null;
+
+    public CallableRequest(Callable<T> callable) {
+        SanityCheck.nonNull(callable, "callable");
+        this.callable = callable;
```
ziccardi: this can be `this.callable = nonNull(callable, "callable");`

core/src/main/java/org/aerogear/mobile/core/reactive/CallableRequest.java
```diff
@@ -0,0 +1,98 @@
...
+
+        Responder<T> responder = responderRef.get();
+        if (responder == null) { // responder may have been disconnected while the calculation was
+                                 // performed.
+            return this;
+        }
+
+        try {
+            if (exception == null) {
+                responder.onResult(value);
```
ziccardi: if `responder` have been disconnected after the previous check, but before this call we will get a `NullPointerException`
I think we should synchronize this block

secondsun: The local repsonder reference can't be set to null by an external thread.  The AtomicReference may lose its reference, but we don't care about it anymore.

core/src/main/java/org/aerogear/mobile/core/reactive/CallableRequest.java
secondsun: I smell a new lint rule we can implement.

core/src/main/java/org/aerogear/mobile/core/reactive/CacheRequest.java
secondsun: See https://github.com/aerogear/aerogear-android-sdk/pull/160

secondsun: image

core/build.gradle
```diff
@@ -47,6 +47,8 @@ dependencies {
     implementation "com.android.support:appcompat-v7"
     implementation "com.squareup.okhttp3:okhttp"
 
+    lintChecks files('libs/checks.jar')
```
danielpassos: Is this related to this PR of with the lint one?

core/src/main/java/org/aerogear/mobile/core/Request.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param responder a responder
+     * @return a chainable instance of Request, not guaranteed to be `this`
+     */
+    Request<T> respondWith(Responder<T> responder);
+
+    /**
+     * Requests can be run off the calling thread. This method configures that.
+     *
+     * @param executorService service to Run the thread on
+     * @return a chainable instance of Request, not guaranteed to be `this`
```
danielpassos: Markdown on the Javadoc?

core/src/main/java/org/aerogear/mobile/core/Request.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Requests may be asynchronous and need to be cancelled.
+     */
+    void cancel();
+
+    /**
+     * This tells the Request chain to cache its value and not to rerun any underlying generating
+     * function.
+     *
+     * @return a chainable instance of Request, not guaranteed to be `this`
```
danielpassos: Markdown on the Javadoc?

core/src/main/java/org/aerogear/mobile/core/Request.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return a request instance that represents the request the parameter was disconnected from.
+     */
+    Request<T> disconnect(Responder<T> responderToDisconnect);
+
+
+    /**
+     * Responses can be run off the calling thread. This method configures that.
+     *
+     * @param executorService service to Run the thread on
+     * @return a chainable instance of Request, not guaranteed to be `this`
```
danielpassos: Markdown on the Javadoc?

core/src/main/java/org/aerogear/mobile/core/reactive/AbstractRequest.java
```diff
@@ -0,0 +1,59 @@
...
+import android.support.annotation.NonNull;
+
+import org.aerogear.mobile.core.Request;
+import org.aerogear.mobile.core.Responder;
+import org.aerogear.mobile.core.utils.SanityCheck;
+
+/**
+ * This class performs wrapping and checking for subclasses.
+ */
+abstract class AbstractRequest<T> implements InternalRequest<T> {
```
danielpassos: Should we merge `AbstractRequest` and `InternalRequest`?

core/src/main/java/org/aerogear/mobile/core/reactive/AbstractRequest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public final Request<T> respondOn(ExecutorService executorService) {
+        nonNull(executorService, "executorService");
+        return new RespondOnRequest<>(this, executorService);
+    }
+
+    @Override
+    public final Request<T> respondWith(@NonNull Responder<T> responder) {
+        SanityCheck.nonNull(responder, "responder");
```
danielpassos: I see in some places you are using the static import and in other you are using the full path `SanityCheck.****`

core/src/main/java/org/aerogear/mobile/core/Request.java
```diff
@@ -0,0 +1,61 @@
+package org.aerogear.mobile.core;
```
danielpassos: Should we move this class into the reactive package since this is reactive related?

core/src/main/java/org/aerogear/mobile/core/Responder.java
```diff
@@ -0,0 +1,29 @@
+package org.aerogear.mobile.core;
```
danielpassos: Should we move this class into the reactive package since this is reactive related?

secondsun:
@danielpassos, @ziccardi  Applied your feedback

core/src/main/java/org/aerogear/mobile/core/reactive/CallableRequest.java
wojta: You have there that sanity check but we should also have `@NonNull Callable<T> callable` there for it to work with IDE's checks and Kotlin's null-safety.

PR #159, opened 2018-03-16
Delete Jenkinsfile from Android SDK

wtrocki:
## Motivation
 
This may be controversial move, but I have noticed that Jenkins build is failing on master and PR"s. Circle ci seems to be working without problems and it's maintained by dev team. Having two CICD solutions seems to be too much overhead and it is confusing when Jenkins build is failing. There is not much benefit of using Jenkins build over circle. If there are some requirements for builds we could move that to Circle CI. As for testing build farm - that should happen for end user templates that will include SDK. IOS SDK works currently only with Circle so we kinda align this changes.

# Demo

screen shot 2018-03-16 at 9 43 18 am


danielpassos:
+9001 @AdamSaleh any thoughts?

PR #166, opened 2018-03-23
Auto init

danielpassos:
Today dev needs to explicit call `MobileCore.init()` before doing anything else related to AeroGear.  This PR make the `MobileCore.init()` be called by the Android lifecycle, **before** app starts (Application, Services, etc...)

It's based on the Firebase blog post: [How does Firebase initialize on Android](https://firebase.googleblog.com/2016/12/how-does-firebase-initialize-on-android.html)

This is part of the work I'm doing to decouple Services creation from the MobileCore ([AGDROID-796](https://issues.jboss.org/browse/AGDROID-796))

PS: Developer will still be able to call init with options with needs.


danielpassos:
Going to apply other PR first then re-open it!

PR #168, opened 2018-03-26
AEROGEAR-2224 Certificate Pinning Information

aidenkeating:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2224

## Description

Documentation on how our SDKs set up and implement certificate pinning. Including documentation on how to generate the required configuration.

## Progress

- [x] Information on how our SDK uses certificate pinning
- [x] Considerations

### DIFF
--- a/docs/getting-started/certificate-pinning.adoc
+++ b/docs/getting-started/certificate-pinning.adoc
@@ -1,0 +1,47 @@
+== Certificate Pinning
+
+Certificate pinning can be enabled in individual SDKs through the `mobile-services.json` file.
+Certificate pinning will only be enabled for services which are used directly by the SDKs. For
+other services SPKI pinning is used. Mobile services must have pinning configured
+separately. For more information on setting up certificate pinning for mobile services see the
+link:https://developer.android.com/training/articles/security-config.html#CertificatePinning[Android network security guide].
+
+=== Generating Pinning Configuration
+
+The Android SDK gets its configuration from the `https` section of the `mobile-services.json` file
+in a project.
+
+[source,json]
+----
+{
+    "services":[],
+    "https": {
+        "certificatePins": [{
+            "host": "example.sync.service",
+            "certificateHash": "exampleHash"
+        }]
+    }
+}
+----
+
+To include the `https` section in configuration when generating it using the Mobile CLI use the
+`--include-cert-pins` flag when retrieving a client configuration. By default, self-signed or
+invalid certs will not be permitted to be included in the certificate pinning configuration. To
+allow these to be included use the `--allow-self-signed-certs` flag.
+
+[source, bash]
+----
+./mobile get clientconfig <client name> --include-cert-pins --allow-self-signed-certs
+----
+
+=== Using Pinning Configuration
+
+If the `https` section is included in the `mobile-services.json` file then certificate pinning will
+automatically be enabled for mobile services.
+
+=== Considerations
+
+If the certificate authority of a service changes then the `mobile-services.json` file will need to
+be regenerated in order to retrieve the correct `https` configuration. The app will then need to be
+rebuilt and republished for the end users to consume. If this is not done then an app may become
+incapable of making network requests to other services.
--- a/docs/getting-started/getting-started.adoc
+++ b/docs/getting-started/getting-started.adoc
@@ -155,3 +155,6 @@ Please follow individual SDK's documentation to see how to add them to the proje
  * link:./core.adoc[Core]
  * link:./auth.adoc[Auth] 
 
+== Certificate Pinning
+
+To implement certificate pinning in the individual SDKs, see link:./certificate-pinning.adoc[the certificate pinning guide].


aidenkeating:
@TommyJ1994 @wei-lee Mind taking a look?

docs/getting-started/certificate-pinning.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+If the `https` section is included in the `mobile-services.json` file then certificate pinning will
+automatically be enabled in any SDK that supports it, for example the Auth SDK. Certificate pinning
+is achieved through the link:http://square.github.io/okhttp/[OkHttp HTTP client] and its
+certificate pinning implementation.
+
+=== Considerations
+
+If the certificate of a service changes then the `mobile-services.json` file will need to be
+regenerated in order to retrieve the correct `https` configuration. If this is not done then an app
```
wei-lee: we should also mention that the apps will need to be rebuilt and republished. All the existing app may stop working.

docs/getting-started/certificate-pinning.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+[source, bash]
+----
+./mobile get clientconfig <client name> --include-cert-hashes --allow-invalid-certs
+----
+
+=== Using Pinning Configuration
+
+If the `https` section is included in the `mobile-services.json` file then certificate pinning will
+automatically be enabled in any SDK that supports it, for example the Auth SDK. Certificate pinning
```
wei-lee: I think we can just mention that it will be automatically enabled by the http library. For android, it will be okhttp.

docs/getting-started/certificate-pinning.adoc
```diff
@@ -0,0 +1,46 @@
...
+        "host": "example.sync.service",
+        "certificateHash": "exampleHash"
+    }]
+}
+----
+
+To include the `https` section in configuration when generating it using the Mobile CLI use the
+`--include-cert-hashes` flag when retrieving a client configuration. By default, self-signed or
+invalid certs will not be permitted to be included in the certificate pinning configuration. To
+allow these to be included use the `--allow-invalid-certs` flag.
```
wei-lee: can we reuse this flag to support self-signed certs as described in this [proposal](https://github.com/aerogear/proposals/pull/34)?

aidenkeating: Yes I think we can. I can rename this to be `--allow-self-signed` or something if that suits both a bit better?

wei-lee: Yeah, I think I prefer `--allow-self-signed-certs`

secondsun:
Will look today :)

secondsun:
I'm +1 to wei's comments.  I know that future versions of Android will begin blocking cleartext http by default, but I can't think of any good ways to work that into this doc or if it is appropriate.

docs/getting-started/certificate-pinning.adoc
```diff
@@ -0,0 +1,46 @@
...
+=== Using Pinning Configuration
+
+If the `https` section is included in the `mobile-services.json` file then certificate pinning will
+automatically be enabled in any SDK that supports it, for example the Auth SDK. Certificate pinning
+is achieved through the link:http://square.github.io/okhttp/[OkHttp HTTP client] and its
+certificate pinning implementation.
+
+=== Considerations
+
+If the certificate of a service changes then the `mobile-services.json` file will need to be
```
TommyJ1994: "If the certificate of a service changes..." - I think this should be the Certificate Authority, rather than the Certificate since we are using SPKI pinning. The SPKI doesn't change when certs are rotated under the same CA.

docs/getting-started/certificate-pinning.adoc
```diff
@@ -0,0 +1,46 @@
+== Certificate Pinning
+
+Certificate pinning can be enabled in individual SDKs through the `mobile-services.json` file.
```
TommyJ1994: I think we should mention that we are using SPKI pinning somewhere here, and its the only pinning technique supported for internal services.

wei-lee:
@TommyJ1994 I think let's not worry about the backup pin for now. If we are using backup pin, I think we should generate it too. I don't think it's a good user experience if we still requires users to manually generate them.

aidenkeating:
@wei-lee Mind taking a look?

docs/getting-started/certificate-pinning.adoc
```diff
@@ -0,0 +1,47 @@
...
+Certificate pinning will only be enabled for services which are used directly by the SDKs. For
+internal services SPKI pinning is used. External services must have pinning configured
```
wei-lee: not sure if it is clear to developers what "internal" and "external" services mean. I think it might be better to use "mobile services" and "other services"?

ciaranRoche:
Closing as these changes where merged in another PR [#176](https://github.com/aerogear/aerogear-android-sdk/pull/176)

PR #171, opened 2018-03-26
[WIP] AEROGEAR-2223 cert pinning

aidenkeating:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2223

## Description

Fork of https://github.com/aerogear/aerogear-android-sdk/pull/152 with some cleaning up and modifications. These are still in progress

## Progress

- [ ] Certificate pinning for auth
- [ ] Testing


core/src/main/java/org/aerogear/mobile/core/http/pinning/CertificatePinningCheck.java
```diff
@@ -0,0 +1,85 @@
...
+ * Class for performing certificate pinning checks.
+ *
+ * This consists of a HTTP GET request using a provided {@link HttpServiceModule}. If the request is
+ * successful then a success listener will be invoked. If unsuccessful then a failure listener will
+ * be invoked.
+ *
+ * If a listener is attached after the request is already completed then the listener will be
+ * invoked immediately with the result of the request being provided.
+ */
+public class CertificatePinningCheck {
```
aidenkeating: @secondsun This PR is still early on, based on another PR which will be closed, should be fleshed out properly tomorrow. Would you mind taking a look at this class, is this what you meant when we spoke about the listener pattern?

core/src/main/java/org/aerogear/mobile/core/http/pinning/CertificatePinningCheck.java
```diff
@@ -0,0 +1,85 @@
...
+    public void execute(@NonNull final String url) {
+        HttpRequest request = this.httpModule.newRequest();
+        request.get(url);
+        HttpResponse httpResponse = request.execute();
+
+        // Handle the success and error responses and set isComplete to allow for immediate
+        // invocation if a new listener is attached.
+        httpResponse.onSuccess(() -> {
+            this.isSuccess = true;
+            this.listener.onSuccess();
```
wei-lee: this will throw a NPE if a listener is never set.

aidenkeating:
@wei-lee @secondsun Mind taking a look?

auth/src/main/java/org/aerogear/mobile/auth/authenticator/oidc/OIDCAuthenticatorImpl.java
```diff
@@ -60,12 +63,14 @@
...
      * @param jwksManager {@link JwksManager}
+     * @param httpModule {@link HttpServiceModule} Module used to make HTTP requests for
+     *        authenticator.
      */
     public OIDCAuthenticatorImpl(final ServiceConfiguration serviceConfiguration,
                     final AuthServiceConfiguration authServiceConfiguration,
                     final AuthStateManager authStateManager,
                     final AuthorizationServiceFactory authorizationServiceFactory,
-                    final JwksManager jwksManager) {
+                    final JwksManager jwksManager, final HttpServiceModule httpModule) {
```
ciaranRoche: Just to make note, as it was mentioned in the other PR, with 6 param's it might be an idea to look at adding a builder to this class as it seems to be doing a lot of the heavy lifting right now. Out of the scope of this PR, but worth nothing all the same.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpCertificatePinningParser.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public OkHttpCertificatePinningParser(final List<CertificatePinningEntry> pinningConfig) {
+        this.pinningConfig = pinningConfig;
+    }
+
+    @Override
+    public CertificatePinner parse() {
+        CertificatePinner.Builder certPinnerBuilder = new CertificatePinner.Builder();
+        for (CertificatePinningEntry pinningEntry : pinningConfig) {
+            String fullPinningHash = "sha256/" + pinningEntry.getCertificateHash();
```
wei-lee: is there a reason that this prefix is not part of the hash value? I think it should be part of the hash value, otherwise it is possible that hash generation using a different alg, but it is not changed here.

PR #172, opened 2018-03-27
 📝  doc: split the SSO doc so that the keycloak setup can be reused in

wei-lee:
other SSO docs

## Motivation

To allow the Keycloak setup to be reused in other sso docs

## Description

Split the existing SSO doc into 2 parts. One for Keycloak, and one for SDK setup.

ping @rachael-oregan mind taking a look?


docs/getting-started/sso/keycloak.adoc
```diff
@@ -0,0 +1,14 @@
+. In order to perform SSO, you need to make sure there is at least one client is configured in Keycloak. Ensure the Keycloak client has the following settings:
```
rachael-oregan: Just a nit pick, might be nice to have a heading here? Something like Keycloak Client Setup

rachael-oregan: "you need to make sure there is at least one client is configured in Keycloak".  Can we remove the second "is" -> "you need to make sure there is at least one client configured in Keycloak"

wei-lee: @rachael-oregan I added a heading here before, but if I do that, I will have to remove the heading from the other doc. So I leave the heading over there and removed it from this one. I think it is ok as this doc is only intended to be used internally and will be embedded in other docs.

wei-lee:
@rachael-oregan addressed your comment. I didn't add heading as if I do that, I will be hard to control the format from the other doc.

can you approve the PR if you are happy with it. Thanks.

PR #173, opened 2018-03-29
Replace Spring dependency management by a single gradle file

danielpassos:
While working on push stuff I figure out [Spring dependency management](https://github.com/spring-gradle-plugins/dependency-management-plugin) doesn't work properly with [Gradle transitive dependencies](https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html). 

It's ignoring the transitive dependencies and on push, it causes some dependency conflicts. Since Google Services plugin doesn't found any dependency it's going to inject it from your on.

```
Android dependency 'com.google.firebase:firebase-iid' has different version for the compile (11.4.2) and runtime (12.0.1) classpath. 
```
This PR moves all our dependencies version for a [single file](https://github.com/aerogear/aerogear-android-sdk/blob/master/constants.gradle) like most of the libraries do.

Related issues: 

* [AGDROID-799](https://issues.jboss.org/browse/AGDROID-799)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/code
     docker:
-      - image: circleci/android:api-26-alpha
+      - image: circleci/android:api-27-alpha
     environment:
       JVM_OPTS: -Xmx3200m
     steps:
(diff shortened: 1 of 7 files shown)


secondsun:
LGTM

PR #179, opened 2018-04-05
Make HTTP response calling error callback on 400 or 500

aliok:
## Motivation

While working on https://issues.jboss.org/browse/AEROGEAR-2395 , I noticed the metrics related integration tests pass all the time.

As written in  <https://square.github.io/okhttp/3.x/okhttp/okhttp3/Call.html#execute--> , execute method does not throw an exception when status is 400 or 500. It only throws an exception when there's a connection problem.

But dependent code assumes that `onError` is called when 400 or 500.

To test the wrong behavior, just run the integration tests without the change in the PR. Even when the URL in integration-test-mobile-services.json is something that returns 500, it will pass. Just copy paste metrics service url, but change a letter so that it returns "503 - Service unavailable".

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
@@ -101,7 +101,8 @@ class OkHttpResponse implements HttpResponse {
     @Override
     public HttpResponse onSuccess(Runnable successHandler) {
         this.successHandler = successHandler;
-        // If there is _any_ response and it is successful the success handler should run immediately
+        // If there is _any_ response and it is successful the success handler should run
+        // immediately
         if (response != null && (response.isSuccessful() || response.isRedirect())) {
             runSuccessHandler();
         }
--- a/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
@@ -30,7 +30,7 @@ class OkHttpResponse implements HttpResponse {
                 response = okHttpCall.execute();
                 requestCompleteLatch.countDown();
 
-                if(response.isSuccessful()) {
+                if (response.isSuccessful() || response.isRedirect()) {
                     // status 200 or 300
                     runSuccessHandler();
                 } else {
--- a/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
+++ b/core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
@@ -25,9 +25,19 @@ class OkHttpResponse implements HttpResponse {
     public OkHttpResponse(final Call okHttpCall, AppExecutors appExecutors) {
         appExecutors.networkThread().execute(() -> {
             try {
+                // this call will throw an exception only when a connection problem occurs
+                // even when there is a 400 or 500 no exception thrown
                 response = okHttpCall.execute();
                 requestCompleteLatch.countDown();
-                runSuccessHandler();
+
+                if(response.isSuccessful()){
+                    // status 200 or 300
+                    runSuccessHandler();
+                } else {
+                    // status 400 or 500
+                    error = new Exception(response.message());
+                    runErrorHandler();
+                }
             } catch (SSLPeerUnverifiedException e) {
                 error = e;
                 requestCompleteLatch.countDown();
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -25,9 +25,19 @@
...
                 response = okHttpCall.execute();
                 requestCompleteLatch.countDown();
-                runSuccessHandler();
+
+                if(response.isSuccessful()){
+                    // status 200 or 300
+                    runSuccessHandler();
+                } else {
+                    // status 400 or 500
+                    error = new Exception(response.message());
```
aliok: @danielpassos any better ideas here instead of plain Exception?

aliok:
@danielpassos @secondsun I have a question about redirects. Should the HTTP client follow redirects? It doesn't seem it does that. This is independent from my PR here.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
danielpassos: What about raise a [HttpException](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/exception/HttpException.java) instead and update the HttpException to receive the Http status code on the constructor?

danielpassos:
I think we can do the redirect in our library, we handle it for push on the old push library. Wdyt @secondsun?

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
aliok: ok, didn't notice we have that one. fixing it now.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -90,8 +101,8 @@ public HttpResponse onError(Runnable errorHandler) {
     @Override
     public HttpResponse onSuccess(Runnable successHandler) {
         this.successHandler = successHandler;
-        // If there is _any_ response the success handler should run
-        if (response != null) {
+        // If there is _any_ response and it is successful the success handler should run immediately
+        if (response != null && (response.isSuccessful() || response.isRedirect())) {
             runSuccessHandler();
```
aliok: As the success handler is immediately called upon setting a success handler using `onSuccess`, I needed to adjust the condition to run it. In some cases, I saw the response is received faster than execution arrives at setting `onSuccess` handler.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -108,9 +119,12 @@ public void waitForCompletionAndClose() {
...
             // If a success Handler was set before this wait then we need to make
             // sure that gets called before we free these resources.
-            if (error == null) {
+            if (error == null && (response.isSuccessful() || response.isRedirect())) {
                 runSuccessHandler();
             } else {
+                if (error == null) {
+                    this.error = new HttpException(response.code(), response.message());
+                }
                 runErrorHandler();
```
aliok: Another case where things fail timing happens differently.

aliok:
@danielpassos this PR is ready for the final review and the merge.

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -90,8 +101,9 @@ public HttpResponse onError(Runnable errorHandler) {
     @Override
     public HttpResponse onSuccess(Runnable successHandler) {
         this.successHandler = successHandler;
-        // If there is _any_ response the success handler should run
-        if (response != null) {
+        // If there is _any_ response and it is successful the success handler should run
+        // immediately
+        if (response != null && (response.isSuccessful() || response.isRedirect())) {
             runSuccessHandler();
```
aliok: As the success handler is immediately called upon setting a success handler using `onSuccess`, I needed to adjust the condition to run it. In some cases, I saw the response is received faster than execution arrives at setting `onSuccess` handler.

danielpassos:
@aliok I'm ok with ignore the Jenkins for now and move on if you want.

aliok:
@danielpassos ok, let's merge this one. unit tests pass, integration tests fail.

PR #184, opened 2018-04-05
Security metrics to contain default metrics

aliok:
## Motivation

JIRA: https://issues.jboss.org/browse/AGDROID-801

## Description

Depends on https://github.com/aerogear/aerogear-android-sdk/pull/179

2 changes:
- Sends event types ("init" or "security") with metrics
- Security events now contains app and device info

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
@@ -2,6 +2,7 @@ package org.aerogear.mobile.security;
 
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import org.aerogear.mobile.core.metrics.MetricsService;
@@ -14,7 +15,8 @@ import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
 class SecurityCheckMetricPublisher implements SecurityCheckExecutorListener {
 
     private final MetricsService metricsService;
-    private final List<SecurityCheckResult> metricResults = new ArrayList<>();
+    private final List<SecurityCheckResult> metricResults =
+                    Collections.synchronizedList(new ArrayList<>());
 
     /**
      * Builds the object.
@@ -26,12 +28,12 @@ class SecurityCheckMetricPublisher implements SecurityCheckExecutorListener {
     }
 
     @Override
-    public synchronized void onExecuted(SecurityCheckResult result) {
+    public void onExecuted(SecurityCheckResult result) {
         metricResults.add(result);
     }
 
     @Override
-    public synchronized void onComplete() {
+    public void onComplete() {
         metricsService.publish(SecurityService.SECURITY_METRICS_EVENT_TYPE,
                         new SecurityCheckResultMetric(metricResults));
     }
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
@@ -32,6 +32,7 @@ class SecurityCheckMetricPublisher implements SecurityCheckExecutorListener {
 
     @Override
     public synchronized void onComplete() {
-        metricsService.publish(new SecurityCheckResultMetric(metricResults));
+        metricsService.publish(SecurityService.SECURITY_METRICS_EVENT_TYPE,
+            new SecurityCheckResultMetric(metricResults));
     }
 }
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
@@ -33,6 +33,6 @@ class SecurityCheckMetricPublisher implements SecurityCheckExecutorListener {
     @Override
     public synchronized void onComplete() {
         metricsService.publish(SecurityService.SECURITY_METRICS_EVENT_TYPE,
-            new SecurityCheckResultMetric(metricResults));
+                        new SecurityCheckResultMetric(metricResults));
     }
 }
--- a/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
@@ -22,6 +22,7 @@ import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
  */
 public class SecurityService implements ServiceModule {
 
+    public static final String SECURITY_METRICS_EVENT_TYPE = "security";
     private final static String TYPE = "security";
 
     private MobileCore core;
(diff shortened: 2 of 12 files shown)


core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -62,7 +58,7 @@ public void destroy() {}
      * Send default metrics
      */
     public void sendAppAndDeviceMetrics() {
-        this.publish(defaultMetrics, null);
+        this.publish(INIT_METRICS_TYPE, new Metrics[0], null);
```
ziccardi: the `new Metric[0]` would be better to be cached to avoid the continuous creation of new objects.
Add a 
`private static final Metrics[] EMPTY_METRICS = new Metrics[0];`
and pass that to the publish method.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -71,31 +67,35 @@ public void sendAppAndDeviceMetrics() {
      * @param callback callback of the publication
      */
     public void sendAppAndDeviceMetrics(final Callback callback) {
-        this.publish(defaultMetrics, callback);
+        this.publish(INIT_METRICS_TYPE, new Metrics[0], callback);
```
ziccardi: See previous comment

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -71,31 +67,35 @@ public void sendAppAndDeviceMetrics() {
...
+    public void publish(@NonNull String type, @NonNull final Metrics[] metrics,
+                    final Callback callback) {
         if (publisher == null) {
             throw new IllegalStateException(
                             "Make sure you have called configure or get this instance from MobileCore.getInstance()");
         }
+        nonNull(type, "type");
         nonNull(metrics, "metrics");
-        publisher.publish(metrics, callback);
+        publisher.publish(type, metrics, callback);
```
ziccardi: `nonNull` returns the passed in object, so this could be changed to a one liner

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
```diff
@@ -2,30 +2,43 @@
...
         nonNull(metrics, "metrics");
+        logger.debug("Metrics: " + type);
+        // first log the default metrics (app and device info)
+        for (final Metrics m : defaultMetrics) {
+            logger.debug("-> [" + m.identifier() + "]:" + m.data().toString());
+        }
+        // then log the specific metrics
         for (final Metrics m : metrics) {
-            logger.debug("Metrics -> [" + m.identifier() + "]:" + m.data().toString());
+            logger.debug("-> [" + m.identifier() + "]:" + m.data().toString());
```
ziccardi: The explicit call to `toString` is useless

auth/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
```diff
@@ -32,6 +32,7 @@ public synchronized void onExecuted(SecurityCheckResult result) {
 
     @Override
     public synchronized void onComplete() {
```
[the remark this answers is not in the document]
aliok: This is how it is called:
`
        final AtomicInteger count = new AtomicInteger(checks.size());
            res.put(check.getId(), (executorService.submit(() -> {
                final SecurityCheckResult result = check.test(getContext());

                final int remaining = count.decrementAndGet();

                for (SecurityCheckExecutorListener listener : listeners) {
                    listener.onExecuted(result);

                    if (remaining <= 0) {
                        listener.onComplete();
                    }
                }

                return result;
            })));
        }
`


`onExecuted` above is sync because it could be called from any one of the checks and it may result in `ConcurrentModificationException` because of the list in there. But can't tell the same thing for `onComplete`. IMO, it can be non-sync.
However, I can see that there might be a case where `onComplete` is called more than once. WDYT?

BTW, I haven't changed this behavior, so let's get opinion from @danielpassos

auth/src/main/java/org/aerogear/mobile/security/SecurityService.java
```diff
@@ -22,6 +22,7 @@
  */
 public class SecurityService implements ServiceModule {
 
+    public static final String SECURITY_METRICS_EVENT_TYPE = "security";
     private final static String TYPE = "security";
```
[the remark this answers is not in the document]
aliok: Ok, will change it to package private

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
aliok: +1

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/LoggerMetricsPublisher.java
aliok: I haven't written this code, but honestly, I do like an explicit call to `toString` to make it obvious. Let's keep this one eh?

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
aliok: will do

aliok: @ziccardi 
```
publisher.publish(nonNull(type, "type"), nonNull(metrics, "metrics"), callback);
```
This looks ugly IMO, when there are multiple parameters.
I will change it only here. Won't change it for all other occurrences in the code.

aliok:
Let's wait until https://github.com/aerogear/aerogear-android-sdk/pull/179 for further review/verification.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
ziccardi: @aliok : yeah, you are right :D probably would look better formatting the code as:
```java
publisher.publish(
              nonNull(type, "type"), 
              nonNull(metrics, "metrics"), 
              callback);
```

but making the checks alone at the start of the method would be probably better

aliok:
```
publisher.publish(
              nonNull(type, "type"), 
              nonNull(metrics, "metrics"), 
              callback);
```
I went with that but our linter doesn't like that :)

aliok:
@ziccardi let's ignore the integration tests for now. unit tests pass. waiting for your approval to merge.

danielpassos:
I have sent a small PR to your branch to move all common stuff to the MetricsPublisher https://github.com/aliok/aerogear-android-sdk/pull/1

aliok:
@danielpassos give me your final approval and I will merge this.

PR #185, opened 2018-04-05
Check if push categories follow the Firebase topic name format

danielpassos:
Since we are creating a Firebase topic for each category, the categories name needs to follow the [Firebase topic name rules](https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/FirebaseMessaging.html#subscribeToTopic(java.lang.String)) 

Related issues:
* [AGDROID-804](https://issues.jboss.org/browse/AGDROID-804)

### DIFF
--- a/push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
+++ b/push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
@@ -6,6 +6,13 @@ import java.util.List;
 
 public final class UnifiedPushConfig {
 
+    /**
+     * Topics in GCM must conform to this pattern. See :
+     * https://developers.google.com/android/reference/com/google/android/gms/gcm/GcmPubSub#unsubscribe(java.lang.String,
+     * java.lang.String)
+     */
+    private static final String FCM_TOPIC_PATTERN = "[a-zA-Z0-9-_.~%]{1,900}";
+
     private String alias;
     private List<String> categories = Collections.emptyList();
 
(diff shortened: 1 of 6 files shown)


push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
```diff
@@ -22,7 +29,29 @@ public void setAlias(String alias) {
     }
 
     public void setCategories(List<String> categories) {
+        validateCategories(categories);
         this.categories = new ArrayList<>(categories);
```
ziccardi: a nonNull check should be added to avoid NullPointerException here

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
```diff
@@ -22,7 +29,29 @@ public void setAlias(String alias) {
...
     }
 
+
+    /**
+     * Validates categories against Google's pattern.
+     *
+     * @param categories a group of Strings each will be validated.
+     * @throws IllegalArgumentException if a category fails to match [a-zA-Z0-9-_.~%]{1,900}
+     */
+    private static void validateCategories(List<String> categories) {
```
ziccardi: If this method would return the `categories` object, we could change the `setCategories` method to a one liner.

Moreover, if `null` is a valid _list of categories_ if `categories` is null, it could return a `Collections.emptyList()`: that way we could avoid the `null` check (mandatory to invoke `new ArrayList<>(categories)`

wtrocki:
Looks good. Going to verify this once @ziccardi issue will be resolved.

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
danielpassos: Done!

push/src/main/java/org/aerogear/mobile/push/UnifiedPushConfig.java
danielpassos: I don't like to have validation methods returning anything

danielpassos:
@wtrocki I have added a `notNull` on setCategories like @ziccardi suggestion

Issue #98, opened 2018-02-26
UserIdentityParser exposes JSONException


secondsun:
## Expected Behavior
Methods should not declare that they throw JSONException unless they are primarily used to process JSON.

## Actual Behavior
Methods declare that they throw JSONException.

PR #186, opened 2018-04-05
Remove unused and uneccessary JSONExceptions

rachael-oregan:
## Motivation

Issue #98 

## Description

Refactored throws `JSONException` to `try/catch` where functions are not parsing JSON & remove unused `JSONException` from functions

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/utils/UserIdentityParser.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/utils/UserIdentityParser.java
@@ -16,8 +16,13 @@ import org.aerogear.mobile.auth.credentials.OIDCCredentials;
 import org.aerogear.mobile.auth.user.RoleType;
 import org.aerogear.mobile.auth.user.UserPrincipalImpl;
 import org.aerogear.mobile.auth.user.UserRole;
+import org.aerogear.mobile.core.MobileCore;
+import org.aerogear.mobile.core.logging.Logger;
 
 public class UserIdentityParser {
+
+    private static final Logger LOG = MobileCore.getLogger();
+    private static final String TAG = "UserIdentityParser";
 
     private static final String USERNAME = "preferred_username";
     private static final String EMAIL = "email";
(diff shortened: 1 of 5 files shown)


auth/src/main/java/org/aerogear/mobile/auth/utils/UserIdentityParser.java
```diff
@@ -56,13 +61,15 @@ public UserIdentityParser(final OIDCCredentials credential,
...
         String firstName = "";
         if (userIdentity != null && userIdentity.has(FIRST_NAME)) {
-            firstName = userIdentity.getString(FIRST_NAME);
+            try {
+                firstName = userIdentity.getString(FIRST_NAME);
+            } catch (JSONException e) {
+                LOG.debug(TAG, "Failed to get user first name from user identity", e);
+            }
         }
         return firstName;
```
wei-lee: This function can be replaced with a one liner:

```java
return userIdentity == null ? "" : userIdentity.optString(FIRST_NAME, "");
```

auth/src/main/java/org/aerogear/mobile/auth/utils/UserIdentityParser.java
```diff
@@ -71,13 +78,15 @@ public String parseFirstName() throws JSONException {
...
         String lastName = "";
         if (userIdentity != null && userIdentity.has(LAST_NAME)) {
-            lastName = userIdentity.getString(LAST_NAME);
+            try {
+                lastName = userIdentity.getString(LAST_NAME);
+            } catch (JSONException e) {
+                LOG.debug(TAG, "Failed to get user last name from user identity", e);
+            }
         }
         return lastName;
```
wei-lee: same as above

auth/src/main/java/org/aerogear/mobile/auth/utils/UserIdentityParser.java
rachael-oregan: Nice, I'll refactor these functions

rachael-oregan:
@wei-lee @ziccardi mind having another look when ye have some time. Thanks

rachael-oregan:
@ziccardi there are not being used externally at the moment no. I think it makes sense to make these private actually you're right

rachael-oregan:
@ziccardi mind having another look please

PR #188, opened 2018-04-06
Add warning dialog to example app

ciaranRoche:
## Motivation

There was an action from the scrum of scrums to add a dialog box on a certificate pinning failure while authenticating in app

## To Verify
- Edit `keycloak` hash value in `mobile-services.json`
- Proceed to Authenticate, request should fail and dialog shown.

### DIFF
--- a/example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
+++ b/example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
@@ -1,21 +1,15 @@
 package org.aerogear.mobile.example.ui;
 
 import android.app.AlertDialog;
-import android.app.Dialog;
-import android.content.DialogInterface;
-import android.os.Bundle;
-import android.support.annotation.Nullable;
 import android.util.Log;
 import android.widget.ImageView;
 import android.widget.TextView;
-import android.widget.Toast;
 
 import org.aerogear.mobile.auth.AuthService;
 import org.aerogear.mobile.auth.authenticator.DefaultAuthenticateOptions;
 import org.aerogear.mobile.auth.user.UserPrincipal;
 import org.aerogear.mobile.core.Callback;
 import org.aerogear.mobile.example.R;
-import org.aerogear.mobile.security.SecurityService;
 
 import butterknife.BindView;
 import butterknife.OnClick;
--- a/example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
+++ b/example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
@@ -61,8 +61,8 @@ public class AuthFragment extends BaseFragment {
 
     public void messageDialog(final String title, final String message) {
         activity.runOnUiThread(() -> {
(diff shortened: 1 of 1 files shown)


example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
```diff
@@ -41,23 +42,32 @@ public void doLogin() {
...
+            public void onError(Throwable error) {
+                // there is an error during the login
+                Log.e(TAG, "login failed due to error " + error.getLocalizedMessage());
+                failureDialog(error.getLocalizedMessage());
+            }
+        });
+    }
+
+    public void failureDialog(String e) {
+        String error = e;
```
ziccardi: What is the use of this?

example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
```diff
@@ -41,23 +42,32 @@ public void doLogin() {
...
-                }
-            });
-        } catch (IllegalStateException e) {
-            Log.e("Pinning Error", e.getMessage());
-        }
+        authService.login(authOptions, new Callback<UserPrincipal>() {
+            @Override
+            public void onSuccess(UserPrincipal models) {
+                // user logged in, continue on..
+                Log.i(TAG, "user logged in " + models.toString());
```
ziccardi: The call to `toString` is useless

example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
```diff
@@ -41,23 +42,32 @@ public void doLogin() {
...
+        });
+    }
+
+    public void failureDialog(String e) {
+        String error = e;
+        activity.runOnUiThread(new Runnable() {
+            @Override
+            public void run() {
+                AlertDialog.Builder certPinningErrorBuilder = new AlertDialog.Builder(activity);
+                certPinningErrorBuilder.setTitle("Failed to Authenticate").setMessage(error).show()
```
wei-lee: make the title a parameter. It will make this method reusable by others.

example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
ciaranRoche: Good catch, left over from a different approach I was taking

example/src/main/java/org/aerogear/mobile/example/ui/AuthFragment.java
```diff
@@ -41,23 +42,27 @@ public void doLogin() {
...
+                // there is an error during the login
+                Log.e(TAG, "login failed due to error " + error.getLocalizedMessage());
+                messageDialog("Failed to Authenticate", error.getLocalizedMessage());
+            }
+        });
+    }
+
+    public void messageDialog(final String title, final String message) {
+        activity.runOnUiThread(() -> {
+            AlertDialog.Builder certPinningErrorBuilder = new AlertDialog.Builder(activity);
```
aidenkeating: `certPinningErrorBuilder` could probably be renamed to be more generic.

Issue #190, opened 2018-04-06
Example app crashes when using local Keycloak


aidenkeating:
## Expected Behavior

App should allow me to connect to my Keycloak server if I use that instead of the provided one in the example app.

## Actual Behavior

App crashes when using own Keycloak server instead of provided one in example app.

```
FATAL EXCEPTION: pool-3-thread-1
                                                                           Process: org.aerogear.mobile.example, PID: 4954
                                                                           java.lang.NullPointerException: Attempt to invoke virtual method 'int okhttp3.Response.code()' on a null object reference
                                                                               at org.aerogear.mobile.core.http.OkHttpResponse.getStatus(OkHttpResponse.java:102)
                                                                               at org.aerogear.mobile.auth.credentials.JwksManager.lambda$fetchJwks$0$JwksManager(JwksManager.java:113)
                                                                               at org.aerogear.mobile.auth.credentials.JwksManager$$Lambda$0.run(Unknown Source:26)
                                                                               at org.aerogear.mobile.core.http.OkHttpResponse.runCompletionHandler(OkHttpResponse.java:52)
                                                                               at org.aerogear.mobile.core.http.OkHttpResponse.lambda$new$0$OkHttpResponse(OkHttpResponse.java:40)
                                                                               at org.aerogear.mobile.core.http.OkHttpResponse$$Lambda$0.run(Unknown Source:20)
                                                                               at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
                                                                               at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
                                                                               at java.lang.Thread.run(Thread.java:764)
```

## Steps to Reproduce the Problem

  1. Change the URL of Keycloak in `mobile-services.json` in the example app to a local Keycloak
  2. Try to initiate auth

aidenkeating:
@StephenCoady Are there any steps I'm missing here to reproduce the bug we saw?

StephenCoady:
No I think thats all @aidenkeating. I can take a look at this as a way to get familiar with the app and sdk.

ciaranRoche:
Hey @StephenCoady as a sanity check, can you have a look at the urls. I have seen this error before, but haven't been able to recreate it now. From memory it was a mistake in the url specified in the config file, when `okhttp` could not get a response it threw that weird error and crashed the app. 
If I use the following service config:
```
"services": [
        {
[...]

PR #191, opened 2018-04-06
Add error handler to JwksManager

ciaranRoche:
## Motivation

GitHub Issue : https://github.com/aerogear/aerogear-android-sdk/issues/190

## Description

Currently on our fetchJwks method we check the status of a response with out checking for an error first.

 In issue https://github.com/aerogear/aerogear-android-sdk/issues/190 when using a local instance of keycloak if a request was made to keycloak with the wrong realm name a `NullPointerException` caused the app to crash.

## Verify PR
In order to replicate issue https://github.com/aerogear/aerogear-android-sdk/issues/190 and verify this PR
- Create local instance of keycloak
- Edit `mobile-service.json` file in example app to reflect local keycloak instance
- Add wrong `realm` name to keycloak service in `mobile-service.json`
- Proceed to auth example in app, and click Authenticate.

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
@@ -106,7 +106,10 @@ public class JwksManager {
         HttpRequest getRequest = httpModule.newRequest();
         getRequest.get(jwksUrl);
         HttpResponse response = getRequest.execute();
-        response.onComplete(() -> {
+        response.onError(() -> {
+            logger.error("fetchJwksError", response.getError().getLocalizedMessage());
+        });
+        response.onSuccess(() -> {
             JsonWebKeySet jwks = null;
             JwksException error = null;
             // this is invoked on a background thread.
--- a/auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
@@ -102,14 +102,12 @@ public class JwksManager {
      */
     public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration,
                     @Nullable final Callback<JsonWebKeySet> callback) {
-        JwksException jwksException = new JwksException("failed to fetch JWKS from server");
         String jwksUrl = nonNull(keycloakConfiguration, "keycloakConfiguration").getJwksUrl();
         HttpRequest getRequest = httpModule.newRequest();
         getRequest.get(jwksUrl);
         HttpResponse response = getRequest.execute();
         response.onComplete(() -> {
             JsonWebKeySet jwks = null;
-            JwksException error = null;
             // this is invoked on a background thread.
             try {
                 if (response.getStatus() == 200) {
@@ -117,10 +115,11 @@ public class JwksManager {
                     try {
                         jwks = new JsonWebKeySet(jwksContent);
                     } catch (JoseException e) {
-                        jwks = null;
-                        error = new JwksException(e);
                         logger.warning("failed to parse JWKS key content. content = "
-                                        + jwksContent);
+                            + jwksContent);
+                        if(callback != null) {
+                            callback.onError(new JwksException(e));
+                        }
                     }
                     if (jwks != null) {
                         persistJwksContent(keycloakConfiguration.getRealmName(), jwksContent);
@@ -128,13 +127,11 @@ public class JwksManager {
                 } else {
                     logger.warning("failed to fetch JWKS from server. url = " + jwksUrl
                                     + " statusCode = " + response.getStatus());
-                    error = new JwksException(jwksException);
+                    throw new JwksException("failed to fetch JWKS from server");
                 }
                 if (callback != null) {
                     if (jwks != null) {
                         callback.onSuccess(jwks);
-                    } else {
-                        callback.onError(error);
                     }
                 }
             } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -106,7 +106,10 @@ public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration
         HttpRequest getRequest = httpModule.newRequest();
         getRequest.get(jwksUrl);
         HttpResponse response = getRequest.execute();
-        response.onComplete(() -> {
+        response.onError(() -> {
+            logger.error("fetchJwksError", response.getError().getLocalizedMessage());
```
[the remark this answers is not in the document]
secondsun: +1

ciaranRoche:
@aidenkeating @secondsun If you wouldn't mind having another look when you get a chance.

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -110,28 +111,35 @@ public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration
...
-                    jwks = null;
-                    error = new JwksException(e);
-                    logger.warning("failed to parse JWKS key content. content = " + jwksContent);
+            try {
+                if (response.getStatus() == 200) {
+                    String jwksContent = response.stringBody();
+                    try {
+                        jwks = new JsonWebKeySet(jwksContent);
+                    } catch (JoseException e) {
+                        jwks = null;
```
ziccardi: If an exception is thrown, jwks is already null

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -110,28 +111,35 @@ public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration
...
+                        logger.warning("failed to parse JWKS key content. content = "
+                                        + jwksContent);
+                    }
+                    if (jwks != null) {
+                        persistJwksContent(keycloakConfiguration.getRealmName(), jwksContent);
+                    }
+                } else {
+                    logger.warning("failed to fetch JWKS from server. url = " + jwksUrl
+                                    + " statusCode = " + response.getStatus());
+                    error = new JwksException(jwksException);
```
ziccardi: Why not simply throwing this exception and doing the `callback.onError` in the catch block?

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -108,30 +108,35 @@ public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration
...
+                        jwks = new JsonWebKeySet(jwksContent);
+                    } catch (JoseException e) {
+                        logger.warning("failed to parse JWKS key content. content = "
+                            + jwksContent);
+                        if(callback != null) {
+                            callback.onError(new JwksException(e));
+                        }
+                    }
+                    if (jwks != null) {
+                        persistJwksContent(keycloakConfiguration.getRealmName(), jwksContent);
```
ziccardi: Put this inside the try/catch block, and you won't need to check if `jwks` is null

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -108,30 +108,35 @@ public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration
...
+                    }
+                } else {
+                    logger.warning("failed to fetch JWKS from server. url = " + jwksUrl
+                                    + " statusCode = " + response.getStatus());
+                    throw new JwksException("failed to fetch JWKS from server");
                 }
-                if (jwks != null) {
-                    persistJwksContent(keycloakConfiguration.getRealmName(), jwksContent);
+                if (callback != null) {
+                    if (jwks != null) {
```
ziccardi: The only path to have `jwks` not null, is that it doesn't throw an exception. So if you put this in the try/catch block, you won't have to check for `jwks` not null

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -108,30 +108,35 @@ public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration
...
-                error = new JwksException("failed to fetch JWKS from server");
-            }
-            if (callback != null) {
-                if (jwks != null) {
-                    callback.onSuccess(jwks);
-                } else {
-                    callback.onError(error);
+            } catch (Exception e) {
+                if (callback != null) {
+                    callback.onError(e);
```
ziccardi: I think this whole block can be simplified with something similar to this:
```java
            if (callback == null) {
                  callback = new NOOP_CALLBACK; // This is a static final variable with a Noop callback impl.
            }
            try {
                if (response.getStatus() == 200) {
                    String jwksContent = response.stringBody();
                    jwks = new JsonWebKeySet(jwksContent);
                    persistJwksContent(keycloakConfiguration.getRealmName(), jwksContent);
                    callback.onSuccess(jwks);
                } else {
                    logger.warning("failed to fetch JWKS from server. url = " + jwksUrl
                                    + " statusCode = " + response.getStatus());
                    throw new JwksException("failed to fetch JWKS from server");
                }
            } catch (JoseException je) {
            	logger.warning("failed to parse JWKS key content. content = "
                            + jwksContent);
                    callback.onError(new JwksException(e));
            } catch (Exception e) {
                    callback.onError(e);
            }
```

ziccardi:
@ciaranRoche I believe that @secondsun is working on this as well to make it `reactive like`, so I think you should sync with him

ciaranRoche:
@ziccardi awesome thanks for that, yeah was working with @secondsun on fixing this issue last Friday.

ciaranRoche:
Closing this PR as changes will be added with https://github.com/aerogear/aerogear-android-sdk/pull/192

PR #192, opened 2018-04-06
Rxization

secondsun:
This applies to the HttpRequest core apis the reactive patterns that were merged a couple of weeks ago.  Additionally this PR includes enhancements to AppExecutors (main thread is now implemented as a executor service), and the reactive APIs (minor bug fixes, a mapping operation, resource cleanup, better canceling support).

Changes to the Http apis include removing the reactive nature of HtpRequest/response and moving most of that code into the more generic reactive framework.  Additionally this PR fixes all breaks  to APIs and modules and tests that it introduces.


auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -104,37 +105,47 @@ public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration
...
+            public void onResult(HttpResponse response) {
+                JsonWebKeySet jwks = null;
+                JwksException error = null;
+                // this is invoked on a background thread.
+                if (response.getStatus() == 200) {
+                    String jwksContent = response.stringBody();
+                    try {
+                        jwks = new JsonWebKeySet(jwksContent);
+                    } catch (JoseException e) {
+                        jwks = null;
```
ziccardi: if an exception is thrown, this is already null

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
```diff
@@ -104,37 +105,47 @@ public void fetchJwks(@NonNull final KeycloakConfiguration keycloakConfiguration
...
+                        LOGGER.warning("failed to parse JWKS key content. content = "
+                                        + jwksContent);
+                    }
+                    if (jwks != null) {
+                        persistJwksContent(keycloakConfiguration.getRealmName(), jwksContent);
+                    }
+                } else {
+                    LOGGER.warning("failed to fetch JWKS from server. url = " + jwksUrl
+                                    + " statusCode = " + response.getStatus());
+                    error = new JwksException("failed to fetch JWKS from server");
```
ziccardi: Any reason why we don't throw the exception and manage it in the `catch` block?

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
```diff
@@ -153,8 +61,23 @@ public String stringBody() {
         }
     }
 
+    @Override
+    public InputStream streamBody() {
+        if (response != null) {
+            return response.body().byteStream();
+        } else {
+            return new ByteArrayInputStream(new byte[0]);
```
ziccardi: Wouldn't be better to cache the empty response to avoid useless objects instantiations:
```java
private static final ByteArrayInputStream EMPTY_RESPONSE =new ByteArrayInputStream(new byte[0]);
...
...
    if (response != null) {
        return response.body().byteStream();
    } else {
        return EMPTY_RESPONSE;
```

core/src/main/java/org/aerogear/mobile/core/http/pinning/CertificatePinningCheck.java
```diff
@@ -67,26 +68,26 @@ public void detachListener() {
...
+                CertificatePinningCheck.this.isSuccess = true;
+                if (CertificatePinningCheck.this.listener != null) {
+                    CertificatePinningCheck.this.listener.onSuccess();
+                }
+                CertificatePinningCheck.this.isComplete = true;
+            }
+
+            @Override
+            public void onException(Exception exception) {
+                CertificatePinningCheck.this.listener.onFailure();
```
ziccardi: I think we should check if the listener is null.
What about initialising the listener to a NOOP listener, so that no null check needs to be done?

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -53,23 +54,27 @@ public void publish(@NonNull final Metrics[] metrics, @Nullable final Callback c
 
             json.put("data", data);
 
-            httpRequest.post(url, json.toString().getBytes());
+            httpRequest.post(url, json.toString().getBytes())
```
ziccardi: I think would be better to specify the charset to be used

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
```diff
@@ -53,23 +54,27 @@ public void publish(@NonNull final Metrics[] metrics, @Nullable final Callback c
...
+                            .respondWith(new Responder<HttpResponse>() {
+                                @Override
+                                public void onResult(HttpResponse value) {
+                                    if (callback != null) {
```
ziccardi: Same comment as before: I think it wouldn't be a bad idea to initialise the callback to a NOOP callback, so that no nullcheck will be necessary. WDYT?

core/src/main/java/org/aerogear/mobile/core/reactive/CallableRequest.java
```diff
@@ -52,6 +76,7 @@ public CallableRequest(Callable<T> callable) {
                 callableThread = null;
             }
         } catch (Exception e) {
+            System.out.println("Exception in callableRequest.call" + e.getMessage());
```
ziccardi: System.out? Is that correct?

core/src/main/java/org/aerogear/mobile/core/reactive/CallableRequest.java
```diff
@@ -80,18 +105,31 @@ public CallableRequest(Callable<T> callable) {
                  * Responders with uncaught exceptions should not blow up the reactive stack. For
                  * now we will log them, but one day a RxPlugin style mechanism may be appropriate.
                  */
+                System.out.println("Exception  thrown in responderThrowable"
```
ziccardi: I think we need just the log, right?

core/src/main/java/org/aerogear/mobile/core/http/OkHttpResponse.java
secondsun: +1

auth/src/main/java/org/aerogear/mobile/auth/credentials/JwksManager.java
secondsun: Yeah good point.

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
secondsun: Let's make this a github issue and feature enhance it.  We will have to make the character set configurable which introduces more complexity.  This PR is mostly about porting the HTTP apis to a reactive pattern not add features to metrics.

core/src/main/java/org/aerogear/mobile/core/metrics/publisher/NetworkMetricsPublisher.java
secondsun: Callbacks as a pattern are next on my refactoring after we merge this PR.  So yes, but not in this PR :)

core/src/main/java/org/aerogear/mobile/core/http/pinning/CertificatePinningCheck.java
secondsun: I fixed this since it doesn't change anything outside of the class and could have crashed

core/src/main/java/org/aerogear/mobile/core/http/pinning/CertificatePinningCheck.java
```diff
@@ -62,31 +71,31 @@ public void attachListener(@NonNull final CertificatePinningCheckListener listen
...
      */
     public void execute(@NonNull final String url) {
         HttpRequest request = this.httpModule.newRequest();
-        request.get(url);
-        HttpResponse httpResponse = request.execute();
+        request.get(url).respondWith(new Responder<HttpResponse>() {
+            @Override
+            public void onResult(HttpResponse httpResponse) {
+                CertificatePinningCheck.this.isSuccess = true;
+                if (CertificatePinningCheck.this.listener != null) {
```
wei-lee: looks like the `listener` is always set, don't think this `if` check is required here.

core/src/main/java/org/aerogear/mobile/core/reactive/MapFunction.java
```diff
@@ -0,0 +1,6 @@
+package org.aerogear.mobile.core.reactive;
+
+@FunctionalInterface
+public interface MapFunction<T, R> {
```
wei-lee: javadoc?

core/src/main/java/org/aerogear/mobile/core/reactive/MapRequest.java
```diff
@@ -0,0 +1,82 @@
+package org.aerogear.mobile.core.reactive;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import android.support.annotation.NonNull;
+
+class MapRequest<T, R> extends AbstractRequest<R> {
```
wei-lee: javadoc?

core/src/main/java/org/aerogear/mobile/core/http/OkHttpRequest.java
```diff
@@ -27,54 +31,116 @@ public OkHttpRequest(final OkHttpClient client, final AppExecutors executors) {
...
+            return (HttpResponse) okHttpResponse;// cast to force generic type
+        }, () -> {
+            /* Cleaner */
+            OkHttpResponse okHttpResponse = response.get();
+            if (okHttpResponse != null && !okHttpResponse.isClosed()) {
+                okHttpResponse.waitForCompletionAndClose();
+            }
+        }).cancelWith(() -> {
+            call.cancel();
+        }).requestOn(appExecutors.networkThread());
```
wei-lee: so looks like all the requests will be executed on the network thread, and it is a single thread. If a request is taking too long to execute, all the other requests will be blocked. I think that is a concern. Should we make the number of threads configurable for the network request executor?

secondsun: Yes, can you make a github issue?

PR #196, opened 2018-04-11
Make logic for returned values from checks consistent

paolobueno:
## Motivation

Returned boolean from existing security checks is not consistent.

In general a positive `passed()` result should point that the check has determined the environment the app is running in is *more* secure as opposed to have a particular feature enabled.
This is also the assumption the metrics dashboards do.

JIRA: https://issues.jboss.org/browse/AEROGEAR-2484

## Description

- Update existing checks to return true when the environment is 'more secure'
- Update example app
- Add more details to checks' javadocs

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/security/checks/EmulatorCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/checks/EmulatorCheck.java
@@ -15,16 +15,19 @@ public class EmulatorCheck extends AbstractSecurityCheck {
      * for x86' If the serial is equal to 'null' The manufacturer contains 'genymotion' If the brand
      * and device start with 'generic'
      *
-     * @return <code>true</code> if device is an emulator
+     * @return <code>true</code> if device is *not* running on an emulator and thus in a 'safer'
+     *         environment
      */
     @Override
     protected boolean execute(@NonNull Context context) {
-        return Build.FINGERPRINT.startsWith("generic") || Build.FINGERPRINT.startsWith("unknown")
+        final boolean isRunningOnEmulator = Build.FINGERPRINT.startsWith("generic")
+                        || Build.FINGERPRINT.startsWith("unknown")
                         || Build.MODEL.contains("google_sdk") || Build.MODEL.contains("Emulator")
                         || Build.MODEL.contains("Android SDK built for x86") || Build.SERIAL == null
                         || Build.MANUFACTURER.contains("Genymotion")
                         || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic"))
                         || "google_sdk".equals(Build.PRODUCT);
+        return !isRunningOnEmulator;
     }
 
     @Override
--- a/auth/src/main/java/org/aerogear/mobile/security/checks/RootedCheck.java
+++ b/auth/src/main/java/org/aerogear/mobile/security/checks/RootedCheck.java
@@ -11,14 +11,16 @@ import android.support.annotation.NonNull;
 public class RootedCheck extends AbstractSecurityCheck {
 
     /**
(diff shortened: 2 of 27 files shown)


psturc:
verifying this...

auth/src/main/java/org/aerogear/mobile/security/checks/EmulatorCheck.java
```diff
@@ -15,16 +15,19 @@
...
     protected boolean execute(@NonNull Context context) {
-        return Build.FINGERPRINT.startsWith("generic") || Build.FINGERPRINT.startsWith("unknown")
+        final boolean isRunningOnEmulator = Build.FINGERPRINT.startsWith("generic")
+                        || Build.FINGERPRINT.startsWith("unknown")
                         || Build.MODEL.contains("google_sdk") || Build.MODEL.contains("Emulator")
                         || Build.MODEL.contains("Android SDK built for x86") || Build.SERIAL == null
                         || Build.MANUFACTURER.contains("Genymotion")
                         || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic"))
                         || "google_sdk".equals(Build.PRODUCT);
+        return !isRunningOnEmulator;
```
ziccardi: no need to have the `isRunningOnEmulator` variable. We can just return.

auth/src/main/java/org/aerogear/mobile/security/checks/RootedCheck.java
```diff
@@ -11,14 +11,16 @@
...
 
     /**
-     * Check whether the device is rooted or not.
+     * Check whether the device is rooted or not. An attacker running the application on a rooted
+     * device can have direct access to its data storage and also has many more tools to eavesdrop
+     * on networked communications
      *
      * @param context {@link Context} to be used by the check
-     * @return <code>true</code> if the device is rooted
+     * @return <code>false</code> if the device is rooted
```
ziccardi: Usually the `true` is documented, and the false is the _otherwise_

paolobueno:
@psturc I've only tested these in a real device, I'll try reproducing your results in an emulator right now

paolobueno:
@ziccardi I think another way we can make these more intuitive is to also update the security check names to reflect their negative nature i.e.: `DeveloperModeDisabledCheck`, `NotRootedCheck`, etc. WDYT

auth/src/main/java/org/aerogear/mobile/security/checks/EmulatorCheck.java
paolobueno: The extra name here is for self-documenting

paolobueno:
@psturc I've checked a couple things out in an emulator and things seem okay, not sure about the rooted check but that's up to rootbeer's implementation.

The developer options are separate from whether a debugger is connected or not. In newer versions of android you need to go to `Settings > System > About device > click 7 times in Build Number` to enable them. This will also open a door to Diagon Alley, London.

I'll try debugging the metrics payload now.

auth/src/main/java/org/aerogear/mobile/security/checks/EmulatorCheck.java
ziccardi: Good point 👍

ziccardi:
@paolobueno 
> I think another way we can make these more intuitive is to also update the security check names to reflect their negative nature i.e.: DeveloperModeDisabledCheck, NotRootedCheck, etc. WDYT

Yes, that can be an option. However we have to clearly state in documentation the 'philosophy'  that lies behind this implementation, because I'm pretty sure that if we don't state it clear, in the future we will have again checks implemented the current way.

paolobueno:
@ziccardi Thanks for taking another look, rebased on master

PR #201, opened 2018-04-12
Autoinit

danielpassos:
Revisiting some Google and [Firebase stuff](https://firebase.googleblog.com/2016/12/how-does-firebase-initialize-on-android.html) @secondsun and I realized the Firebase Android SDK doesn’t need we do any init to the lib starts/be configured.

TL;DR what Firebase does is uses a [Content Provide](https://developer.android.com/guide/topics/providers/content-providers.html) (not really created for this) as a trick to [initialize the lib before the app starts](https://youtu.be/AJqakuas_6g?t=20m38s)

## Why Content provider?

Because it has access to the [applicationContext](https://developer.android.com/reference/android/content/ContentProvider.html#getContext()) and is called before all other Android kinds of stuff in the Android lifecycle (including Application)

I’d like to do the same for our Android SDK and initialize the lib using the same trick so we can provide a MobileCore.getInstance() already configured with all the things the developer needs (Logger, HttpLayer, Mobile Services info, etc…) without need to create an application class or call MobileCore.init(context) anywhere

It will also allow us to send the default metrics before the Android lifecycle starts the app and have some controls (when need) to know when the app is in background and foreground, listeners possible crashes and other things without the developer needs to call/configure anything.


wtrocki:
> It will also allow us to send the default metrics.. 

This is awesome

secondsun:
Taking a look :)

wtrocki:
Taking a look :)

core/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,9 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="org.aerogear.android.core" />
+    package="org.aerogear.android.core">
+    <application>
+        <provider
+            android:name="org.aerogear.mobile.core.bootstrap.AeroGearBootstrap"
```
wtrocki: replace with `{applicationId}`

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -37,18 +38,20 @@
...
     public static final String DEFAULT_CONFIG_FILE_NAME = "mobile-services.json";
 
-    private static final String TAG = "AEROGEAR/CORE";
     private static final int DEFAULT_READ_TIMEOUT = 30;
     private static final int DEFAULT_CONNECT_TIMEOUT = 10;
     private static final int DEFAULT_WRITE_TIMEOUT = 10;
+
+    @SuppressLint("StaticFieldLeak")
+    private static MobileCore instance;
+    // TODO Move to instance
```
wtrocki: Maybe we could add ticket, but it seems to be trivial task.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -37,18 +38,20 @@
 
     public static final String DEFAULT_CONFIG_FILE_NAME = "mobile-services.json";
 
-    private static final String TAG = "AEROGEAR/CORE";
     private static final int DEFAULT_READ_TIMEOUT = 30;
     private static final int DEFAULT_CONNECT_TIMEOUT = 10;
     private static final int DEFAULT_WRITE_TIMEOUT = 10;
+
+    @SuppressLint("StaticFieldLeak")
+    private static MobileCore instance;
```
wtrocki: [Optional] Probably overlooked final here?

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
danielpassos: Not so trivial because of the tests, that is why I wanna do it on a separate PR

core/src/main/AndroidManifest.xml
danielpassos: I can't use `{applicationId}` here or the class will not be found. This trick was used on the `authorities` only because Android doesn't allow have more than one authorities in the same app, and we wanna provide it automatically ;)

Is that makes sense?

wtrocki: Tried that now. You are right. Sorry for problems.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
danielpassos: I have no idea how to do that in a singleton static variable :/

My point is `init` is creating it, so... I think it can't be final!

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -76,73 +79,69 @@ private MobileCore(final Context context, final Options options)
...
      *
      * @param context Application context
      * @param options AeroGear initialization options
-     * @return MobileCore instance
      */
-    public static MobileCore init(final Context context, final Options options)
+    public static void init(final Context context, final Options options)
                     throws InitializationException {
-        return new MobileCore(context, options);
+        instance = new MobileCore(context, options);
```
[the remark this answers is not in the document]
danielpassos: I'm not 100% sure about it. The library will create an instance automatically but the user/dev should wanna replace some default options and re-create it. 

If we decide to prevent recreation we should provide some way to change user/dev replace/change default options. wdyt?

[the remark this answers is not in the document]
wei-lee: right, I get it now. Personally I don't like the idea to ask developers to re-init the who object just to reset the options. I think the MobileCore itself should provide a method to allow set the options after it's being created. This way the instance of the MobileCore will only need to be inited once.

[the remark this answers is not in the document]
danielpassos: +9001 I will add that and make it immutable.

secondsun:
Seems to work on my machine.  I'm having a trouble getting the metrics to send because I'm using a self signed cert, but that isn't related to this feature.

PR #202, opened 2018-04-16
Adapt to new Keycloak service config

aliok:
Adapt to new Keycloak service config as it is changed in https://github.com/aerogearcatalog/keycloak-apb/pull/70

New sample config:

```
{
            "id": "keycloak",
            "name": "keycloak",
            "type": "keycloak",
            "url": "https://www.mocky.io/v2/5a6b59fb31000088191b8ac6",
            "config": {
                "name": "keycloak",
                "public_installation": "{\"ssl-required\": \"external\", \"realm\": \"myproject\", \"resource\": \"juYAlRlhTyYYmOyszFa\", \"url\": \"https://keycloak-myproject.192.168.64.74.nip.io/auth\", \"auth-server-url\": \"https://keycloak-myproject.192.168.64.74.nip.io/auth\"}",
                "realm": "project2",
                "type": "keycloak",
                "uri": "https://keycloak-myproject.192.168.64.74.nip.io"
            }
        },
```

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/configuration/KeycloakConfiguration.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/configuration/KeycloakConfiguration.java
@@ -6,6 +6,8 @@ import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
 import android.net.Uri;
 
 import org.aerogear.mobile.core.configuration.ServiceConfiguration;
+import org.json.JSONException;
+import org.json.JSONObject;
 
 /**
  * A class to represent the configuration options of the Keycloak singleThreadService
@@ -22,6 +24,8 @@ public class KeycloakConfiguration {
     private static final String BASE_URL_TEMPLATE = "%s/realms/%s/protocol/openid-connect";
     private static final String LOGOUT_URL_TEMPLATE = "%s/logout?%s=%s&%s=%s";
 
+    private static final String PUBLIC_INSTALLATION_KEY = "public_installation";
+
     private final String serverUrl;
     private final String realmId;
     private final String resourceId;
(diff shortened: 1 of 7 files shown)


aliok:
@danielpassos there's something I couldn't figure out.
In the test `AuthServiceTest#testReady`
this code returns an empty JSONObject (KeycloakConfiguration.java)
```
final JSONObject publicConfigJSON = new JSONObject(publicConfig);
```
although `publicConfig` is a valid JSON object.
https://github.com/aerogear/aerogear-android-sdk/pull/202/files#diff-49f136d180d13dd9d35d407e92f084d7R47 

Couldn't understand why. Mind helping me there?

wei-lee:
@aliok I don't think the configuration is right. It shouldn't contain a nested json object. The content of the `public_installation` _IS_ the content of the config object. I don't think the changes here are necessary.

cc @maleck13 am I right?

maleck13:
@wei-lee I think currently it returns as Ali shows.

aliok:
@maleck13 now that @wei-lee points out the structure, can we even improve the structure so that  `public_installation`'s value is **the** generated config? Anything blocks that?

I was focusing on making things work only.

wei-lee:
@maleck13 But I think the cli is generating in a different format? see https://github.com/aerogear/mobile-cli/blob/b9e55a9c7caab1e485b14f2bb8051e043aa85028/pkg/cmd/types.go#L153.

danielpassos:
@aliok The only problems I see in the [circle-ci log](https://circleci.com/gh/aerogear/aerogear-android-sdk/745) is about format violations.

```
* What went wrong:
Execution failed for task ':auth:spotlessJava'.
> The following files had format violations:
```

You can easily fix it running `./gradlew spotlessApply` ;)

maleck13:
@wei-lee  you are correct. @philbrookes Seems the config format from the UI is different than the format from the CLI. They should be the same. The CLI uses a convertor based on the service.

maleck13:
hmmm ok I think we need to think about this one a little more. At the moment there is service specific code in the CLI which it would be better to avoid.
I think a better approach may be to use a secret with a single "config" field in which the json config is put. We then always read this field and unmarshal the payload

maleck13:
I will add an issue to Jira to look at this

secondsun:
Can I get some details on the change to the keycloak format?

maleck13:
@secondsun there are no changes to the keycloak format. Just an aligning of the CLI and UI to give the same format (which at the moment it doesnt)

aliok:
@philbrookes could you point me at the code piece? would like to start this work early tomorrow.

philbrookes:
@aliok hey, the logic is here: https://github.com/openshift/origin-web-console/pull/2925/files#diff-b04d87eb6adc8f6cdbbc52309b203a19R66

aliok:
Closing this one as this change in the SDK is not necessary. UI should be updated instead.

aliok:
Ok, hack is implemented: https://github.com/aliok/origin-web-console/tree/promotional-video-hack
Mobile CLI and the web console now prints the same mobile app config.

PR #203, opened 2018-04-16
Fix bg push notification

danielpassos:
## Problem

When the app is killed by the user or device restarts the app will stop to receive background push notifications because it's using [mainThread AppExecutor](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/executor/AppExecutors.java#L61) instead of [singleThreadService AppExecutor](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/executor/AppExecutors.java#L69) and `DefaultHandler` will be null because the SDK only retrieve it when `configure` is called.

## How to test it

1. Go to the push screen and register the device on UPS
1. Send a message from UPS with the app in Foreground in the Push screen and you should it theListView
1. Send a message from UPS with the app in Background or in whatever screen not push one and you should see an notification on notification bar
1. Kill/close/force close the app and send a message from UPS and you should see an notification on notification bar

### DIFF
--- a/push/src/main/java/org/aerogear/mobile/push/PushService.java
+++ b/push/src/main/java/org/aerogear/mobile/push/PushService.java
@@ -306,6 +306,10 @@ public class PushService implements ServiceModule {
     public static void notifyHandlers(final Context context, final Map<String, String> message) {
         nonNull(context, "context");
 
+        if (defaultHandler == null) {
+            getDefaultHandler(context);
+        }
+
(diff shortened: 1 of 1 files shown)


wtrocki:
Little bit late for the game, but I have reviewed that now and really like this change. @danielpassos Thank you so much for adding this!

PR #204, opened 2018-04-17
Aerogear 2512 Update docs directory structure to reflect antora usage

finp:
## Motivation

antora requires a different structure from asciibinder, need to update based on that.

JIRA: https://issues.jboss.org/browse/AEROGEAR-2512

## Description

update readme, and move files inside /docs
note: no edits to files and using same toc as per old _topic_map.yml file


wtrocki:
@finp Good stuff. Are you planning to do IOS SDK as well?

docs/local-site.yml
```diff
@@ -0,0 +1,17 @@
+site:
+  title: AeroGear Docs Site
+  start_page: android-sdk::index
+content:
+  sources:
+  - url: ../
+    branches: AEROGEAR-2512
```
secondsun: Is this the correct value for branch?  I feel like it should be master.

finp: @secondsun chances are,  this will never be the 'correct value', it's only for a local preview. The normal workflow I see is to create a branch for your work, then change this value to your branch name and validate your work as you proceed. This will all be documented in https://github.com/aerogear/processes/tree/master/modules/docs/pages 
... making it up as I go along ;)

finp:
@wtrocki  https://issues.jboss.org/browse/AEROGEAR-2479

docs/local-site.yml
secondsun: Was looking at the antora docs and it they say if you use HEAD then you will always pick the current branch for the project.  Is this a solution?

finp: @secondsun  that's perfect solution, thanks 
(glad someone reads docs)

secondsun:
Can you put build instructions in the readme?  I'm not able to get antora to run.

secondsun:
@finp Ok, I'm a +1, my problems seem like antora was trying to process the push branch of my repository as well as the head branch.  I have no idea why so that is something to keep in mind because I'm sure someone else will hit it one day.

secondsun:
@finp It was a hilarious bug in antora https://gitlab.com/antora/antora/issues/241

PR #205, opened 2018-04-18
Add push refresh token

danielpassos:
Firebase changes the Device token from time to time and we need to update that on the UPS

See: https://firebase.google.com/docs/cloud-messaging/android/client#monitor-token-generation

### DIFF
--- a/push/src/main/java/org/aerogear/mobile/push/PushService.java
+++ b/push/src/main/java/org/aerogear/mobile/push/PushService.java
@@ -17,6 +17,7 @@ import com.google.firebase.iid.FirebaseInstanceId;
 import com.google.firebase.messaging.FirebaseMessaging;
 
 import android.content.Context;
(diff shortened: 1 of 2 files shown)


push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -187,17 +234,7 @@ public void onException(Exception error) {
      * @param callback A callback to handle successful or failed unregistration
      */
     public void unregisterDevice(final Callback callback) {
```
secondsun: Can you refactor to return a Request instead of taking a callback?  I can help you with this.

danielpassos: I have created a new Jira for it => https://issues.jboss.org/browse/AGDROID-806

Issue #103, opened 2018-02-27
Improve how a user can get the Resource or Realm roles using the SDK


TommyJ1994:
## Expected Behavior
I can easily get all the client or realm roles for a token by following the documentation or having a function that can do this.

## Actual Behavior
This isn't really possible. Keycloak libraries tend to have individual `getClientRoles()` and `getRealmRoles()` to get the different types of roles. However, the SDK only exposes a `getRoles()` function with both types mixed in. The current `getRoles()` docs doesn't document how I can get either client roles or realm roles using this function. I think we should either document how we can extract the different role types using the `getRoles()` function or keep in line with Keycloak libraries that expose both using individual methods.

## Steps to Reproduce the Problem

  1. Read [auth.adoc#retrieving-a-users-roles](https://github.com/aerogear/aerogear-android-sdk/blob/master/docs/getting-started/auth.adoc#retrieving-a-users-roles)

## Specifications

  - Core SDK Version: 5.0.0.SNAPSHOT
  - Relevant Service Modules: auth

TommyJ1994:
At the moment its seems much easier for a developer for a user to just get the resource/realm arrays from the access token directly, rather than using our SDK to do this. I don't think this is the best user experience.

PR #208, opened 2018-04-23
add getRealmRoles and getResourceRoles methods

rachael-oregan:
## Motivation

Github Issue #103 

## Description

Added `getRealmRoles()` and `getResourceRoles()` methods to improve the user experience of getting different users roles.

### DIFF
--- a/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
@@ -44,16 +44,6 @@ public class UserPrincipalImpl implements UserPrincipal {
     private final Set<UserRole> roles;
 
     /**
-     * Realm roles associated with this principal.
-     */
-    private final Set<UserRole> realmRoles;
-
-    /**
-     * Resource roles associated with principal.
-     */
-    private final Set<UserRole> resourceRoles;
-
-    /**
      * Identity token. It is used for logout
      */
     private final String identityToken;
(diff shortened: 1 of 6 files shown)


auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -128,6 +145,20 @@ public Builder withRoles(final Set<UserRole> roles) {
...
         }
 
+        public Builder withRealmRoles(final Set<UserRole> realmRoles) {
+            if (realmRoles != null) {
+                this.realmRoles.addAll(realmRoles);
+            }
+            return this;
+        }
+
+        public Builder withResourceRoles(final Set<UserRole> resourceRoles) {
```
wei-lee: I don't think we need those methods. They can be retrieved from the existing `roles` set. They will have different `RoleType` in the roles set.

rachael-oregan:
@wei-lee I refactored to remove those builder methods and fields for the specific roles.  Let me know if its good to merge. Thanks :)

auth/src/main/java/org/aerogear/mobile/auth/user/UserPrincipalImpl.java
```diff
@@ -206,6 +206,42 @@ public String getName() {
...
+        return realmRoles;
+    }
+
+    /**
+     * Get's the user resource roles
+     *
+     * @return user's resource roles
+     */
+    @Override
+    public Set<UserRole> getResourceRoles() {
```
wei-lee: those 2 methods are the same, except the RoleType. We can make that a parameter and a single method can used by both methods here.

rachael-oregan: Added this single method there

PR #211, opened 2018-04-30
Sending metrics automatically on init

danielpassos:
Related issues:

* [AGDROID-802](https://issues.jboss.org/browse/AGDROID-802)

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/bootstrap/AeroGearBootstrap.java
+++ b/core/src/main/java/org/aerogear/mobile/core/bootstrap/AeroGearBootstrap.java
@@ -8,12 +8,24 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import org.aerogear.mobile.core.MobileCore;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/aerogear/mobile/core/bootstrap/AeroGearBootstrap.java
```diff
@@ -8,12 +8,24 @@
...
     public boolean onCreate() {
         MobileCore.init(getContext());
+
+        try {
+            MetricsService metricsService =
+                            MobileCore.getInstance().getService(MetricsService.class);
+            metricsService.sendAppAndDeviceMetrics(
+                            error -> MobileCore.getLogger().error(error.getMessage()));
+        } catch (ConfigurationNotFoundException e) {
+            // Metrics is not enable and should be ignored
```
wtrocki: `Metrics SDK is not enable*d*. Ignoring error`

danielpassos: Replaced by your suggestion as a Logger

PR #212, opened 2018-04-30
Prevent custom metrics

paolobueno:
## Motivation

Prevent extension of the metrics API to add custom metrics

JIRA: https://issues.jboss.org/browse/AGDROID-783

## Description

- Make `Metrics` an `abstract class` in order to support package-scoping its methods, making them visible only to children (i.e. metrics record implementations) in other packages.
- Add javadocs highlighting the internal usage intent of metrics-publishing related interfaces and methods
- Revise and reduce the scope of some of these entities
- Remove asciidocs related to extending metrics with custom data


core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -79,6 +77,8 @@ public void sendAppAndDeviceMetrics(final Callback callback) {
     /**
      * Send metrics
      *
+     * @apiNote Not intended by direct usage outside of the SDK
```
paolobueno: Another approach would be to avoid the internal API changes and just sprinkle these comments around.

wtrocki: I really like this! IMHO that is all we need to do here.

core/src/main/java/org/aerogear/mobile/core/metrics/NetworkMetricsPublisher.java
```diff
@@ -37,7 +35,7 @@ public NetworkMetricsPublisher(final Context context, final HttpRequest httpRequ
     }
 
     @Override
-    public void publish(@NonNull String type, @NonNull final Metrics[] metrics,
+    protected void publish(@NonNull String type, @NonNull final Metrics[] metrics,
```
wtrocki: Actually this method is used only internally so not sure why needs to be protected.

paolobueno: good catch, I thought I was getting compile errors by turning this package-private but changing the scope of the parent method should be okay :+1:

core/src/main/java/org/aerogear/mobile/core/metrics/Metrics.java
```diff
@@ -4,23 +4,25 @@
 
 /**
  * Interface for a record to be published to the metrics service
```
ziccardi: It is not an interface anymore

paolobueno:
@ziccardi any suggested approach?

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
```diff
@@ -79,6 +77,8 @@ public void sendAppAndDeviceMetrics(final Callback callback) {
     /**
      * Send metrics
      *
+     * WARN: Not intended by direct usage outside of the SDK
+     *
```
secondsun: @paolobueno Let's create an annotation and add it to our linter so that this warning is enforceable at compile time

core/src/main/java/org/aerogear/mobile/core/metrics/Metrics.java
```diff
@@ -3,24 +3,26 @@
...
- * Interface for a record to be published to the metrics service
- * 
+ * Record to be published to the metrics service
+ *
+ * WARN: Not intended by direct usage outside of the SDK
+ *
  * @param <T> A type that can be serialized into a {@link JSONObject} value
  */
-public interface Metrics<T> {
+public abstract class Metrics<T> {
```
secondsun: So I'm -1 on this change.  I think it might be better (and less hacky) if our publishers are declared final and we do instance of enforcement of metrics implementations at that level.

secondsun:
@paolobueno As I said in my review, let's just do instance of checks and put in linter warnings.

core/src/main/java/org/aerogear/mobile/core/metrics/Metrics.java
paolobueno: Okay, I'll revert this back to the regular interface.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java
secondsun: @paolobueno It doesn't require an annotation processor, we just have to code a rule for it in the lint rules which are in the codequality directory and rebuild the checks.jar file  I can help you with it because it is areally fun to do if you like source parsing and processing.

codequality/lint/checks/src/main/java/org/aerogear/android/lint/CustomMetricsDetector.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public List<String> applicableSuperClasses() {
+        // This avoids unnecessary visits
+        return Collections.singletonList(METRICS_INTERFACE);
+    }
+    @Override
+    public void visitClass(JavaContext context, UClass declaration) {
+        final boolean insideSDK = declaration.getQualifiedName().startsWith(SDK_NAMESPACE);
+        if(insideSDK) {
```
ziccardi: If the user creates a custom metric object in a package named `org.aerogear.mobile.[anything]` this won't work.

ziccardi:
@paolobueno @secondsun I don't think using the linter is a real solution. Linter could be disabled inside user build process. 
The only real solution I see would be to make the interface/abstract class visible only at package level and then sign the library so that all the classes inside the same package must be signed by the same signer.
I'm -1 on this since it doesn't provide a real solution: if we can't sign the jar, I think just documenting that the interface is not to be extended could be enough.

Maybe we could add @Deprecated to the interface so that a warning is raised if trying to implement it.

paolobueno:
Closed in favor of #220

PR #213, opened 2018-04-30
Avoid to refrehs push token when it does not exists yet

danielpassos:
TL:DR Firebase is calling the [onTokenRefresh()](https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceIdService#onTokenRefresh()) on the first time we call [getToken()](https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceId.html#getToken()) forcing the lib to refresh a token how does not exist yet.

Related issues:

* [AGDROID-808](https://issues.jboss.org/browse/AGDROID-808)


wtrocki:
I notice that registration is not saved as before. Debugging that now.

danielpassos:
@wtrocki I have added a refresh token button in example app

PR #214, opened 2018-04-30
Make MobileCore destroy a instance method

danielpassos:


wtrocki:
@danielpassos 🤔 Is there actually need to destroy service modules?

danielpassos:
@wtrocki The motivation is this is the only method I forgot to remove the static stuff on this PR https://github.com/aerogear/aerogear-android-sdk/pull/201 :tired_face:

I got you we don't need this method, but I wanna leave it there before we kill the service manager from MobileCore. See: [AGDROID-796](https://issues.jboss.org/browse/AGDROID-796)

PR #215, opened 2018-04-30
Removing LoggerMetricsPublisher

danielpassos:
There is no sense to have `LoggerMetricsPublisher` anymore since [`MobileCore`](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/MobileCore.java) is going to [raise](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/MobileCore.java#L164-L167) a [ConfigurationNotFoundException](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/exception/ConfigurationNotFoundException.java) is there is no _metrics_ entry in `mobile-service.json` and someone try to use [`MetricsService`](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/metrics/MetricsService.java)


wtrocki:
`NetworkMetricsPublisher` is not protected for null url and it is crashing with `NullPointerException`

wtrocki:
Ignore my previous comment. 
If there is no URL we probably do not want to collect metrics at all.

danielpassos:
@wtrocki I have added some SanityCheck to make you happy :kissing_heart:

secondsun:
I think that logger metrics are good for debugging, but we might want to think about how we enable / disable it instead.

core/src/main/java/org/aerogear/mobile/core/metrics/MetricsPublisher.java
```diff
@@ -24,7 +24,7 @@
     private final Metrics[] defaultMetrics;
 
     public MetricsPublisher(final Context context) {
-        this.context = context;
+        this.context = nonNull(context, "context");
```
ziccardi: This should be `this.context = nonNull(context, "context").getApplicationContext();` for safety

wtrocki:
@danielpassos This looks like forgotten PR :-) However we have removed that everywhere else. @ziccardi @secondsun are you ok to get that merged?

ziccardi:
@wtrocki : other than the comment I already left on this PR, LGTM

danielpassos:
@wtrocki How can I simulate the NPE problem?

PR #217, opened 2018-05-03
fix: use type for mobile-services

wtrocki:
## Motivation

Use type for fetching mobile configurations instead of name. 
Name is just human readable name and it's not guaranteed to be unique.

See: https://groups.google.com/forum/#!topic/aerogear/1hxhEPBASVY


wtrocki:
@wojta FYI in xamarin

wtrocki:
Right. Going to adjust the change and add new method.

wtrocki:
Closing as original value is right. Separate ticket created to fix this.

PR #218, opened 2018-05-04
AEROGEAR-2669 - Android SDK should query for configuration type and id

aliok:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2669

## Description

* After discussion on the mailing list [1] [2], we needed to adapt the SDK in terms of how to get the configs for the services.
* Before, we were using "name" properties for getting the service config. Now we need to use "id" and/or "type" properties.

What's done:
* Introduced methods on MobileCore to
  * get config by id --> needed for custom runtime connectors
  * get configs by type --> needed when there are multiple configs for a service type
  * get config by type --> similar to above, but gets the first one found. this is the one used by mobile core by default for services created automatically (Metrics, Keycloak etc). This is basically the old "getServiceConfiguration" method modified.
* Got rid of the service cache in MobileCore because of the idea that we can have multiple instances of a service for different configs
* Got rid of `destroy` methods in `ServiceModule` impls, because we don't store the initialized services in the cache anymore. Also, all of the `destroy` method impls were empty

[1]: https://groups.google.com/forum/#!topic/aerogear/1hxhEPBASVY
[2]: https://groups.google.com/forum/#!topic/aerogear/tXsNboCR-8A

## Additional Notes

~~WIP PR!
I will do the changes in the example app after we agree on the SDK changes.~~

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -4,7 +4,7 @@ import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
@@ -38,6 +38,8 @@ public final class MobileCore {
     private static final int DEFAULT_CONNECT_TIMEOUT = 10;
     private static final int DEFAULT_WRITE_TIMEOUT = 10;
 
+    private static final String TAG = "MobileCore";
+
     @SuppressLint("StaticFieldLeak")
     private static MobileCore instance;
     // TODO Move to instance
@@ -47,8 +49,8 @@ public final class MobileCore {
     private final String appVersion;
     private final String configFileName = "mobile-services.json";
     private final HttpServiceModule httpLayer;
-    private final Map<String, ServiceConfiguration> servicesConfig;
-    private final Map<Class<? extends ServiceModule>, ServiceModule> services = new HashMap<>();
+    private final Map<String, ServiceConfiguration> serviceConfigById;
+    private final Map<String, List<ServiceConfiguration>> serviceConfigsByType;
 
     /**
      * Get the user app version from the package manager
@@ -83,7 +85,8 @@ public final class MobileCore {
         try (final InputStream configStream = context.getAssets().open(configFileName)) {
             MobileCoreConfiguration jsonConfig = new MobileCoreJsonParser(configStream).parse();
             httpsConfig = jsonConfig.getHttpsConfig();
-            servicesConfig = jsonConfig.getServicesConfig();
+            serviceConfigById = jsonConfig.getServicesConfigPerId();
+            serviceConfigsByType = jsonConfig.getServiceConfigsPerType();
         } catch (JSONException | IOException exception) {
             String message = String.format("%s could not be loaded", configFileName);
             throw new InitializationException(message, exception);
@@ -100,7 +103,8 @@ public final class MobileCore {
                         .writeTimeout(DEFAULT_WRITE_TIMEOUT, TimeUnit.SECONDS)
                         .readTimeout(DEFAULT_READ_TIMEOUT, TimeUnit.SECONDS);
         final OkHttpServiceModule httpServiceModule = new OkHttpServiceModule(builder.build());
-        ServiceConfiguration configuration = this.servicesConfig.get(httpServiceModule.type());
+        ServiceConfiguration configuration =
+                        this.getServiceConfigurationByType(httpServiceModule.type());
         if (configuration == null) {
             configuration = new ServiceConfiguration.Builder().build();
         }
@@ -127,16 +131,6 @@ public final class MobileCore {
         return instance;
     }
 
-    /**
-     * Called when mobile core instance needs to be destroyed
-     */
-    public void destroy() {
-        for (Class<? extends ServiceModule> serviceKey : services.keySet()) {
-            ServiceModule serviceModule = services.get(serviceKey);
-            serviceModule.destroy();
-        }
-    }
-
     @SuppressWarnings("unchecked")
     public <T extends ServiceModule> T getService(final Class<T> serviceClass) {
         return getService(serviceClass, null);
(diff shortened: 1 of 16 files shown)


aliok:
@danielpassos @secondsun @wtrocki @josemigallas 
CircleCI fails because I haven't adapted the example app. Lets agree on the SDK changes first. I can adapt the example app afterwards.

This PR is ready for review.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -227,13 +215,43 @@ public HttpServiceModule getHttpLayer() {
...
+
+    /**
+     * Returns the configuration for this service type from the JSON config file.
+     *
+     * If there are multiple configs for the type, the first one will be returned.
+     *
+     * @param type Service type
+     * @return the first configuration for this service type from the JSON config file
+     */
+    public ServiceConfiguration getServiceConfigurationByType(final String type) {
```
danielpassos: I think we should not have this one. By type should always return a list.

danielpassos:
I was wondering if getConfig should throw a [ConfigNotFoundException](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/exception/ConfigurationNotFoundException.java) or return null. I don't know that I hate more :P

danielpassos:
I know it's probably blocking you, so let's move on with this, but I think it's a good opportunity to [decouple the creation of the services](https://issues.jboss.org/browse/AGDROID-796) from the `MobileCore`.

wtrocki:
IMHO - let's do some minimal work required to make Android work with current config as final solution will be done here: https://issues.jboss.org/browse/AGDROID-796

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
aliok: @danielpassos how about renaming it to "getSingleServiceConfigByType" ?

otherwise, I will have to duplicate the content of that method in the callers.

aliok: @danielpassos Also, John recommends having a single configs for services: 
So, I would like to keep that method.

secondsun:
taking a 👀

psturc:
@aliok failed test fixed in https://github.com/aerogear/aerogear-android-sdk/pull/221

core/src/test/java/org/aerogear/mobile/core/unit/configuration/ServiceConfigurationTest.java
```diff
@@ -12,11 +12,11 @@
 
     @Test
     public void testCreateConfig() {
-        ServiceConfiguration config = ServiceConfiguration.newConfiguration().setName("conf-name")
+        ServiceConfiguration config = ServiceConfiguration.newConfiguration().setId("conf-id")
                         .setType("conf-type").setUrl("http://test-uri.feedhenry.org")
                         .addProperty("prop1", "value1").addProperty("prop2", "value2").build();
 
-        Assert.assertEquals("conf-name", config.getName());
+        Assert.assertEquals("conf-id", config.getId());
```
josemigallas: Could we have this value in a final string? It is asking for a typo :P

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreJsonParser.java
```diff
@@ -49,7 +49,7 @@ private String readJsonStream(final InputStream jsonStream) throws IOException {
         Map<String, ServiceConfiguration> serviceConfigs = new HashMap<>();
```
ziccardi: You already know how many elements you are going to put inside the map (length).
Better initialize this to:
```java
Map<String, ServiceConfiguration> serviceConfigs = new HashMap<>(length);
```

to avoid allocating too much/few memory

core/src/main/java/org/aerogear/mobile/core/ServiceModule.java
```diff
@@ -26,10 +26,4 @@
      * @return <code>true</code> if the singleThreadService configuration should be defined.
      */
     boolean requiresConfiguration();
-
```
josemigallas: Does this change belong in this PR?

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreJsonParser.java
```diff
@@ -60,17 +60,18 @@ private ServiceConfiguration parseConfigObject(final JSONObject jsonObject)
 
         final ServiceConfiguration.Builder serviceConfigBuilder =
                         ServiceConfiguration.newConfiguration();
-        serviceConfigBuilder.setName(jsonObject.getString("name"));
+        serviceConfigBuilder.setId(jsonObject.getString("id"));
         serviceConfigBuilder.setUrl(jsonObject.getString("url"));
         serviceConfigBuilder.setType(jsonObject.getString("type"));
```
ziccardi: All these could be chained:
```java
final ServiceConfiguration.Builder serviceConfigBuilder =ServiceConfiguration.newConfiguration()
                      .setName(jsonObject.getString("name"))
                      .setId(jsonObject.getString("id"))
                      .setUrl(jsonObject.getString("url"))
                      .setType(jsonObject.getString("type"));
```

core/src/main/java/org/aerogear/mobile/core/ServiceModule.java
aliok: Yes. As I have deleted the service cache, there is no way to find instances of services and destroying them. Thus, destroy method was obsolete. All impls of that method was empty anyway.

core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreConfiguration.java
```diff
@@ -10,12 +13,30 @@
...
+    private final Map<String, ServiceConfiguration> serviceConfigsPerId;
+
+    // a map of <service type, list<config>> as there can be multiple configs for a service type
+    private final Map<String, List<ServiceConfiguration>> serviceConfigsByType;
+
+    public MobileCoreConfiguration(final Map<String, ServiceConfiguration> serviceConfigsPerId,
                     final HttpsConfiguration httpsConfig) {
         this.httpsConfig = httpsConfig;
-        this.serviceConfig = serviceConfig;
+        this.serviceConfigsPerId = serviceConfigsPerId;
```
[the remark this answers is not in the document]
aliok: Ok, will take a look.

core/src/test/java/org/aerogear/mobile/core/unit/configuration/ServiceConfigurationTest.java
aliok: I wouldn't create constants for those in the tests :) In case of a typo, the tests fail and you can fix the typo immediately.

aliok:
@psturc let's merge #221 separately. Thanks for pointing that out!

@josemigallas @ziccardi @psturc fixed the example app and removed the WIP status. ready to merge on your approval.

PR #219, opened 2018-05-04
Update docs

danielpassos:


wtrocki:
@danielpassos Reverting this changes. We removed metrics documentation here!

PR #220, opened 2018-05-08
Remove docs for creating custom metrics records

paolobueno:
## Motivation

Remove docs for creating custom metrics records

JIRA: https://issues.jboss.org/browse/AGDROID-783

## Description

The rest of the outdated documentation related to metrics were removed in #219.
The init metrics are now sent automatically and the other existing metrics record which is related to the security checks is documented in the doc files for that module, as such there are no longer references on how to send individual (hence possibly custom) records.

## Additional Notes

Revised from #212 after mailing list discussion.


wtrocki:
Actually this docs are kinda useful for contributors. We still have them for contributing section in IOS SDK. So let's not remove them. Just not link it in end user docs.

paolobueno:
Rejecting the removal in favor of just moving the existing file

PR #224, opened 2018-05-14
Add New Keycloak URL Reference

TommyJ1994:
## Motivation

Update the Auth SDK to use the top-level Keycloak Service URL field instead for authentication.

JIRA: https://issues.jboss.org/browse/AEROGEAR-2703

## Progress
- [x] Update the Keycloak URL
- [x] Update Tests/Build


TommyJ1994:
@maleck13 @david-martin Any volunteers to verify this with the DMZ/Proxy use case? 

Just need to check:
- [ ] Auth Login/Logout works fine.
- [ ] There are no JWT errors in logcat - eg `Invalid JWT provided - .....`

wtrocki:
@TommyJ1994 See: https://groups.google.com/forum/#!topic/aerogear/LBQKO0KERgQ

auth/src/main/java/org/aerogear/mobile/auth/configuration/KeycloakConfiguration.java
```diff
@@ -36,7 +38,9 @@
     public KeycloakConfiguration(final ServiceConfiguration configuration) {
         nonNull(configuration, "configuration");
 
-        this.serverUrl = nonEmpty(configuration.getProperty(SERVER_URL_NAME), SERVER_URL_NAME);
+        this.authServerUrl = nonEmpty(configuration.getProperty(AUTH_SERVER_URL_NAME),
```
maleck13: We should always use the url field in the config and not rely on the auth-server-url

wei-lee: @TommyJ1994 Sorry I didn't make it clear in the JIRA. The "auth-server-url" field should not be used anywhere at all. Only the `url` field should be used.

TommyJ1994: To verify JWT tokens (`iss` field), you need the actual Keycloak URL, not the Proxy/DMZ one as it will fail, so the `auth-server-url` is only used for token verification, the `url` field is used for everything else. That make sense?

wei-lee: ah, right. OK, at long as it's not used for establish the communication with the server, then I think we are fine.

@TommyJ1994 can you add a comment to the code just to make it clear that the `auth-server-url` should only be used for token verification?

TommyJ1994: @wei-lee Yeah, sure.

secondsun:
@TommyJ1994 Is proxy support in the mobile-cli or will I need to add the url field manually?

auth/src/main/java/org/aerogear/mobile/auth/configuration/KeycloakConfiguration.java
```diff
@@ -90,6 +94,16 @@ public String getHostUrl() {
         return this.serverUrl;
     }
 
+    /**
+     * Get the Auth Server URL string of the Keycloak singleThreadService This URL should only be
+     * used to get the issuer field for token verification.
+     *
+     * @return the Auth Server URL of the Keycloak singleThreadService
```
wei-lee: remove `singleThreadService`

auth/src/main/java/org/aerogear/mobile/auth/configuration/KeycloakConfiguration.java
```diff
@@ -90,6 +94,16 @@ public String getHostUrl() {
...
+     */
+    public String getAuthHostURL() {
```
wei-lee: If this is only used for issuer verification, would it better to rename it to `getIssuerHostname` to avoid confusion?

docs/getting-started/certificate-pinning-best-practices.adoc
```diff
@@ -0,0 +1,30 @@
+= Overview
```
wei-lee: is this file required as part of this PR?

filelist.log
```diff
@@ -0,0 +1,204 @@
+.circleci/config.yml
```
wei-lee: should this file be here?

docs/getting-started/certificate-pinning-best-practices.adoc
TommyJ1994: @wei-lee woops, no. I'll remove now.

wei-lee:
@TommyJ1994 I think we can merge this

PR #226, opened 2018-05-15
Reactive API for PushServer

secondsun:
## Motivation

We want to move away from Callbacks, but the PushService  class needed them because we didn't have a good way for chaining requests.  This PR adds the requestMap operation to the RX infrastructure and updates PushServices and the example app to not use callbacks.

JIRA: https://issues.jboss.org/browse/AGDROID-2670

## Progress

- [x] Added requestMap
- [x] Added requestMap tests
- [x] Updated PushService
- [x] Update example app push


wtrocki:
This resolves issue that @danielpassos was working on

danielpassos:
@secondsun Can you rebase it?

PR #230, opened 2018-05-22
AEROGEAR-2784 remove example app

rachael-oregan:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2784

## Description

Remove the example app from the SDK

Note: The showcase app is not working on master at the moment

## Progress

- [x] Delete example app


danielpassos:
Should we do that before fixing the showcase app?

rachael-oregan:
@danielpassos ya I'll hold off on merging this until the showcase app is fixed

wtrocki:
Showcase app do not seem to be fully functional. It constantly crashes with compilation errors. It do not provide good development workflow, there is no documentation and it is constantly broken because people do not update it. We need to add the same changes. Example: https://github.com/aerogear/android-showcase-template/pull/2

Changes are good but let's not merge this until we have something we can use like in cordova or ios.

rachael-oregan:
The showcase app is working both locally and with published SDKs (just verified this here -> https://github.com/aerogear/android-showcase-template/pull/9) so going to merge this

PR #231, opened 2018-05-22
Update readme

laurafitzgerald:
## Motivation

Update comms channels in README etc.

## Description

* AGDROID -> AEROGEAR
* Remove reference to deprecated mailing list

## Progress

- [x] AGDROID -> AEROGEAR
- [x] Remove reference to deprecated mailing list

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,7 @@
 
 <!-- The reason underlying the contents of the PR, can be a link to the originating JIRA -->
 
-JIRA: https://issues.jboss.org/browse/AGDROID-{}
+JIRA: https://issues.jboss.org/browse/AEROGEAR-{}
 
 ## Description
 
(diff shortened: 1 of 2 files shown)


danielpassos:
I'm -1 to have all project/platform under AEROGEAR Jira, we have this before and it was a pain in the ass, but this is a discussion for the ML

aidenkeating:
@laurafitzgerald This needs a rebase.

@danielpassos There is already a [discussion](https://groups.google.com/forum/#!searchin/aerogear/JIRA$20instances$3A$20One$20for$20the$20better$20(or$20one$20to$20rule$20them$20all)%7Csort:date/aerogear/DMUClcL577E/JiM08t5SAQAJ) started around this. Using AEROGEAR is the approach we've been following for quite a while so I'd be a fan of merging this as it reflects the current state of what we're doing. If we change our approach then we can update later on.

danielpassos:
@aidenkeating Sure, I don't wanna block this, just raise the discussion, but let's move this fro ML

laurafitzgerald:
@aidenkeating rebased. @danielpassos i can confirm Aiden's note that we agreed that we would align across the teams to use AEROGEAR project on JIRA.

PR #232, opened 2018-05-23
feat: initial auto release pipeline

darahayes:
## Description

Related ticket: https://issues.jboss.org/browse/AGDROID-807

This is the initial PR for a release pipeline. It most likely doesn't work and will need some testing and tweaking.

## How it Works

The SDKs are first released to [Nexus](https://oss.sonatype.org/). Nexus is an intermediate place where the artefacts go before being released to Maven. Nexus creates temporary repos where the released artefacts can be downloaded and tested by end users before the actual release to Maven.

This PR automates the initial push to Nexus, while the release to Maven is done at a later stage (by pushing a button in Nexus).

## How will team members use this?

The steps for creating a release will be as follows:

* Manually update the `VERSION_NAME` and `VERSION_CODE` properties and create a PR. This PR shows to other team members you intend to create a release and gives them a chance to approve.
* Once the PR is approved and merged, use Github to create a release where the tag == the new `VERSION_NAME`. This will kick off the automated release to Nexus.
* Once we are happy with that,  a user will log into nexus and kick off the release to maven.

## What's happening in CircleCI?

Firstly the release process requires some things:

* Nexus Credentials (username, token). We've set up an aerogear-ci user for this.
* A GPG keyring file (includes public + private key) which can be used to sign the release artefacts. This file is a binary file.
* A `gradle.properties` file that looks like the following:

```
signing.keyId=<id of keyring>
signing.password=<password for key file>
signing.secretKeyRingFile=<full path to keyring>
```

Before we can release we need to get these two files into the build container. The only method for injecting secrets in Circle is using environment variables.

What I've done is base64 encoded the keyring and gradle.properties files and stored them as environment variables. Then inside the build container, before the release is run we decode the variables back into the correct files. You can see this happening in the circle config with the following lines:

```
      - run:
          name: Create GPG Secret Key file from environment variable
          command: echo $GPG_SECRING_BASE64 | base64 --decode >> ~/secring.gpg
      - run:
          name: Create Gradle Proprerties config file from environment variable
          command: echo $GRADLE_SIGNING_PROPERTIES_BASE64 | base64 --decode >> ~/.gradle/gradle.properties
```

We then run a script called `./scripts/validateRelease.sh` which validates a number of things.

* git tag syntax must be valid semver i.e. `x.y.z` or it can include an extra label so things such as `x.y.z-SHAPSHOT` or `x.y.z-rc1` are allowed.
* validates that `VERSION_NAME` in `gradle.properties` is equal to the git tag. i.e. I cannot kick off a release using whatever tag I want, the VERSION_NAME must match.
* Validates that `VERSION_CODE` in `gradle.properties` is a number.

Lastly, the release is run using the appropriate gradle command.

## To Do

* Get it working
* Associated Docs on how to use it and how it works


darahayes:
Hey @danielpassos I think this PR is ready to go. I've included a fairly comprehensive doc on how to perform releases. Would love if you could take a look when you have the chance!

darahayes:
Also, if/when this gets approval please don't merge straight away. I'd like to do a squash + merge.

danielpassos:
@darahayes Can you rebase it? Please *don't* squash + merge it

docs/contrib/releases.adoc
```diff
@@ -0,0 +1,88 @@
+== SDK Release Documentation
+
+The release process results in new versions of the core and associated SDKs being published to Maven. This docmument should help you understand how releases work and how to create one.
```
danielpassos: s/docmument/document

docs/contrib/releases.adoc
```diff
@@ -0,0 +1,88 @@
...
+=== How to Prepare a Release
+
+Ensure the git repo is clean and there are no outstanding changes. Then check out a new branch.
+
+[source, bash]
+--
+git checkout -b release-x.y.z
+--
+
+Update the `VERSION_NAME` property in the `gradle.properties` file at the root of the repository. The `VERSION_NAME` **must be in the format x.y.z or x.y.z-label**
```
danielpassos: What about: The `VERSION_NAME` must follow the [Semantic Versioning](https://semver.org/)

PR #236, opened 2018-05-24
[WIP] updating README/Contrib to match other sdks

laurafitzgerald:
## Motivation

There have been some updates to the ios-sdk README/CONRIB

## Progress

- [x] Update README/CONTRIBUTION 
- [x] Update filenames in docs/modules/ROOT/pages as per [discussion](https://groups.google.com/forum/#!topic/aerogear/pdP8BFfxcTw) on mailing list


README.md
```diff
@@ -7,64 +7,31 @@
...
-1. [Troubleshooting](./docs/troubleshooting.adoc)
-1. [Service Contributor Guide](./docs/contributing-guide.adoc)
 
 ### List of SDKs
 
-AeroGear Services SDK consist of set of separate SDKs
+AeroGear Services consist of set of separate SDKs
+
+- Core - Core library used as base for other SDKs
+- Auth - Mobile Application Auth SDK (part of the core bundle)
```
danielpassos: Auth is not part of the core. It's a separated module.

README.md
```diff
@@ -7,64 +7,31 @@
...
+- Security - Mobile Application Self Defense Checks
```
danielpassos: I know for now we have only self-defense check, but I think we should say it's a security module, not self-defense module.

laurafitzgerald: @danielpassos yes makes sense, I will update that in both prs, Android and iOS.

README.md
laurafitzgerald: @danielpassos will update, thanks.

README.md
laurafitzgerald: @finp based on that comment should the file structure mentioned here: https://groups.google.com/forum/#!topic/aerogear/pdP8BFfxcTw be updated to be `using-auth-module.adoc`?

finp: @laurafitzgerald  let's stick with sdk for now

PR #237, opened 2018-05-24
Move Security features out of Auth module

ziccardi:
## Motivation

We need to move the security classes to its own module

JIRA: https://issues.jboss.org/browse/AEROGEAR-2849

See [PR12](https://github.com/aerogear/android-showcase-template/pull/12) for changes to the showcase app

## Description

To give users the ability to choose what to load depending on their needs, we decided to put the Security features into its own module

## Progress

- [x] Move security classes to its own module
- [x] Update and test the showcase app


security/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+
+dependencies {
+    implementation project(path: ":core")
+
+    implementation "com.scottyab:rootbeer-lib:${project.ext.rootbeer_version}"
+
+    testImplementation "junit:junit:${project.ext.junit_version}"
+    testImplementation "org.mockito:mockito-core:${project.ext.mockito_version}"
+    testImplementation "org.robolectric:robolectric:${project.ext.robolectric_version}"
+}
```
ciaranRoche: We need to add `apply from: "../gradle-mvn-push.gradle"` to this file for the security module to be picked up by gradle

ziccardi: Yeah, I forgot to commit them! Good catch!

ciaranRoche: It is @secondsun catch not mine haha

PR #240, opened 2018-05-25
Decouple Push Service creation from MobileCore

danielpassos:
## Related issues

This is part of the [AGDROID-796](https://issues.jboss.org/browse/AGDROID-796)

### Openshift projects

Credentials provided by _mobile-services.json_

```
{
    "version": 1,
    "clusterName": "https://someURL",
    "namespace": "myNamespace
    "clientId": "myClienId",
    "services": [
        {
            "id": "myPushId",
            "name": "push",
            "url": "https://someURL/",
            "type": "push",
            "config": {
                "android": {
                    "variantId": "[variant id]",
                    "variantSecret": "[variant secret]",
                    "senderId": "[firebase project number]"
                }
            }
        }
    ]
}
```

Replace:

```java
PushService pushService = MobileCore.getInstance().getService(PushService.class);
```

By

```java
PushService pushService = new PushService.Builder().openshift().build();
```

### DIFF
--- a/push/src/main/java/org/aerogear/mobile/push/PushService.java
+++ b/push/src/main/java/org/aerogear/mobile/push/PushService.java
@@ -91,13 +91,14 @@ public class PushService {
 
         public Builder() {}
 
-        public Builder openshift(String id) {
+        public Builder openshift() {
 
             MobileCore mobileCore = MobileCore.getInstance();
 
             context = mobileCore.getContext();
 
-            ServiceConfiguration serviceConfiguration = mobileCore.getServiceConfigurationById(id);
+            ServiceConfiguration serviceConfiguration =
+                            mobileCore.getServiceConfigurationByType("push");
 
             try {
                 JSONObject android = new JSONObject(
--- a/push/src/main/java/org/aerogear/mobile/push/PushService.java
+++ b/push/src/main/java/org/aerogear/mobile/push/PushService.java
@@ -40,6 +40,8 @@ import org.aerogear.mobile.core.reactive.Responder;
  */
 public class PushService {
 
+    public static final String TYPE = "push";
+
     private static final String TAG = PushService.class.getName();
 
     private static final String DEFAULT_MESSAGE_HANDLER_KEY = "DEFAULT_MESSAGE_HANDLER_KEY";
@@ -98,7 +100,7 @@ public class PushService {
             context = mobileCore.getContext();
 
             ServiceConfiguration serviceConfiguration =
-                            mobileCore.getServiceConfigurationByType("push");
+                            mobileCore.getServiceConfigurationByType(TYPE);
 
             try {
                 JSONObject android = new JSONObject(
--- a/push/src/main/java/org/aerogear/mobile/push/PushService.java
+++ b/push/src/main/java/org/aerogear/mobile/push/PushService.java
@@ -275,6 +275,8 @@ public class PushService {
 
     /**
      * Update the device token on Unified Push Server
+     *
+     * @param context Application context
      */
     public static void refreshToken(Context context) {
         nonNull(context, "context");
--- a/push/src/main/java/org/aerogear/mobile/push/PushService.java
+++ b/push/src/main/java/org/aerogear/mobile/push/PushService.java
@@ -1,5 +1,23 @@
 package org.aerogear.mobile.push;
 
+import static java.net.HttpURLConnection.HTTP_NO_CONTENT;
+import static java.net.HttpURLConnection.HTTP_OK;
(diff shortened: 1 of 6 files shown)


push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -394,7 +450,7 @@ public static void notifyHandlers(final Context context, final Map<String, Strin
      * </code>
      */
     @SuppressWarnings("unchecked")
-    private static void getDefaultHandler(final Context context) {
+    private static void setDefaultHandler(final Context context) {
```
secondsun: Now that I think about it, defaultHandler isn't a property so we probably shouldn't use "get" or "set".  Maybe initDefaultHandler, or createDefaultHangler?

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -34,75 +35,102 @@
 import org.aerogear.mobile.core.reactive.Requester;
 import org.aerogear.mobile.core.reactive.Responder;
 
-
 /**
  * The entry point for communication with Unified Push Server
  */
-public class PushService implements ServiceModule {
+public class PushService {
```
secondsun: Why are we removing ServiceModule?

secondsun: Why are we removing ServiceModule?

push/src/main/java/org/aerogear/mobile/push/PushService.java
secondsun: Since defaultHandler isn't a bean property I think that neither get not set is appropriate.  Maybe init?

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: @secondsun because it forced me to implement some methods used by MobileCore I wanna kill, but I don't wanna touch on MobileCore/ServiceModule for now.

secondsun: @wtrocki @danielpassos I was thinking it could still be a marker interface for our tooling, but we don't have any tooling, nvm :)

danielpassos: We can re introduce the interface later if we want

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -246,59 +276,85 @@ private String getHashedAuth(final String variant, final char[] secret) {
     /**
      * Update the device token on Unified Push Server
      */
-    public void refreshToken() {
-        JSONObject jsonObject = retrieveCache();
+    public static void refreshToken(Context context) {
```
secondsun: Add @param to javadoc

danielpassos: Done!

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -34,75 +35,103 @@
...
-            throw new ConfigurationNotFoundException(
-                            "An error occurred while trying to load the push config");
+        public Builder openshift() {
+
+            MobileCore mobileCore = MobileCore.getInstance();
+
+            context = mobileCore.getContext();
+
+            ServiceConfiguration serviceConfiguration =
+                            mobileCore.getServiceConfigurationByType("push");
```
wtrocki: Map to constant.

push/src/main/java/org/aerogear/mobile/push/PushService.java
```diff
@@ -34,75 +35,103 @@
...
-            unifiedPushCredentials.setSenderId(android.getString(JSON_SENDER_ID_KEY));
+        public Builder() {}
 
-        } catch (JSONException e) {
-            MobileCore.getLogger().error(e.getMessage(), e);
-            throw new ConfigurationNotFoundException(
-                            "An error occurred while trying to load the push config");
+        public Builder openshift() {
+
+            MobileCore mobileCore = MobileCore.getInstance();
```
wtrocki: Do you think it will make sense to just wrap this method into OpenShift configuration class. Then you can have StandardConfig class etc. and use strategy pattern. It will be cleaner to maintain extend. Just hint.

wtrocki:
Really like the way we doing tokenRefresh! 
Really good job! Let me know if you want me to do final verification.

push/src/main/java/org/aerogear/mobile/push/PushService.java
danielpassos: Done!

Issue #246, opened 2018-05-31
Ignore cases when lookup configurations based on types


wei-lee:
## Expected Behavior

When lookup the service configurations based types, the case shouldn't really matter. E.g. I can use "keycloak" or "Keycloak"

## Actual Behavior

It does matter.

PR #247, opened 2018-06-05
AEROGEAR-2905 - Ignore cases when looking up configurations based on …

rachael-oregan:
…types

## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-2905

GitHub Issue: #246 

## Description

Refactored service configuration *type* lookup to ignore the letter case requested and search for lowercase 

## Progress

- [x] refactor service config *type* lookup to ignore the letter case used in the request

### DIFF
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -246,7 +246,7 @@ public final class MobileCore {
      * @return the configurations for this service type from the JSON config file
      */
     public List<ServiceConfiguration> getServiceConfigurationsByType(final String type) {
-        return serviceConfigsByType.get(type);
+        return serviceConfigsByType.get(type.toLowerCase());
     }
 
     /**
@@ -258,7 +258,7 @@ public final class MobileCore {
      * @return the first configuration for this service type from the JSON config file
      */
     public ServiceConfiguration getServiceConfigurationByType(final String type) {
-        final List<ServiceConfiguration> configs = serviceConfigsByType.get(type);
+        final List<ServiceConfiguration> configs = serviceConfigsByType.get(type.toLowerCase());
         if (configs == null || configs.isEmpty()) {
             return null;
         }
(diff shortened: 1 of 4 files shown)


ziccardi:
@rachael-oregan : I think we should add a 'toLowercase' [here](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/configuration/MobileCoreConfiguration.java#L32) as well so that we are sure that map keys are lowercase.

rachael-oregan:
@ziccardi the search doesn't really matter here because `serviceConfigsByType.get(serviceType);` always returns null as `serviceConfigsByType` is just initialised 3 lines before doing the search. I'm not even sure why it's checking for null here `if (configsForType == null)` because `configsForType` is always going to be null

ziccardi:
@rachael-oregan : it is inside a loop. And inside the loop we do:
`
serviceConfigsByType.put(serviceType, configsForType);
`
so if more than one configuration exists for a given type, we could get a non null value.

rachael-oregan:
@ziccardi ah right ya ok, I'll update that now. Thanks 👍

rachael-oregan:
@ziccardi mind having another look

rachael-oregan:
@secondsun ya sure that makes sense, I'll add in some tests

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -246,7 +246,7 @@ public HttpServiceModule getHttpLayer() {
      * @return the configurations for this service type from the JSON config file
      */
     public List<ServiceConfiguration> getServiceConfigurationsByType(final String type) {
-        return serviceConfigsByType.get(type);
+        return serviceConfigsByType.get(type.toLowerCase());
```
wojta: maybe sanity check or non null check would be good in this case

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -258,7 +258,7 @@ public HttpServiceModule getHttpLayer() {
      * @return the first configuration for this service type from the JSON config file
      */
     public ServiceConfiguration getServiceConfigurationByType(final String type) {
-        final List<ServiceConfiguration> configs = serviceConfigsByType.get(type);
+        final List<ServiceConfiguration> configs = serviceConfigsByType.get(type.toLowerCase());
```
wojta: maybe sanity check or non null check would be good in this case

rachael-oregan:
@secondsun @wojta mind having another look please

PR #248, opened 2018-06-06
feat: release 0.1.0-2018-23

darahayes:


danielpassos:
You also need to change the version code

darahayes:
@danielpassos are we good to go with this now?

PR #255, opened 2018-06-13
Security metrics are now sent

secondsun:
## Motivation

The showcase app wasn't sending metrics for device checks.  This is now fixed

JIRA: https://issues.jboss.org/browse/AEROGEAR-3130

## Description

The metrics published was not attaching a responder to a request object.  Rx objects don't run unless they have a response to send things to.  There may be a need to update some javadocs to clarify this at the metricsPublisher level; this is documented behavior on the Request class as well as a standard Rx pattern (cold single observable) which is documented on the request class as well.

## Progress

- [x] Metrics send

## Additional Notes

To verify this issue you can use the showcase app, you will need to provision metrics and make sure that you either import the https certificate or use an insecure route.

image

### DIFF
--- a/security/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
+++ b/security/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
@@ -5,7 +5,9 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import org.aerogear.mobile.core.MobileCore;
 import org.aerogear.mobile.core.metrics.MetricsService;
+import org.aerogear.mobile.core.reactive.Responder;
 import org.aerogear.mobile.security.metrics.SecurityCheckResultMetric;
 
 /**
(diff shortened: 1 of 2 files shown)


security/src/main/java/org/aerogear/mobile/security/SecurityCheckMetricPublisher.java
```diff
@@ -35,6 +37,17 @@ public void onExecuted(SecurityCheckResult result) {
     @Override
     public void onComplete() {
         metricsService.publish(SecurityService.SECURITY_METRICS_EVENT_TYPE,
-                        new SecurityCheckResultMetric(metricResults));
+                        new SecurityCheckResultMetric(metricResults))
+                        .respondWith(new Responder<Boolean>() {
```
david-martin: I'm struggling to understand how this fixes the problem. (I don't doubt that it does)
Is it a case of you need to observe the response before it will actually happen? (Quantum HTTP) :)

secondsun: @david-martin Cold Single Observer pattern.  pretty normal Rx stuff

PR #264, opened 2018-07-16
Remove antora files

wtrocki:
## Motivation

Remove stale antora docs that were pushed to repository.


danielpassos:
@finp thoughts?

PR #266, opened 2018-07-19
Sync rx

danielpassos:
Related issues: 

* [AEROGEAR-7653](https://issues.jboss.org/browse/AEROGEAR-7653)

This PR create utility methods to use GraphQL query, mutation, and subscription wrapping up the Apollo client call using AeroGear Rx

Instead of doing:

```java
SyncService
        .getInstance()
        .getApolloClient()
        .query(MyQueryClassGeneratedByApollo.builder().build())
        .enqueue(new ApolloCall.Callback<MyQueryClassGeneratedByApollo.Data>() {
            @Override
            public void onResponse(@Nonnull Response<MyQueryClassGeneratedByApollo.Data> response) {
                // Yay!
            }

            @Override
            public void onFailure(@Nonnull ApolloException e) {
                // Oops!
            }
        });
```

You can do: 

```java
SyncService
        .getInstance()
        .query(MyQueryClassGeneratedByApollo.builder().build())
        .execute(MyQueryClassGeneratedByApollo.Data.class)
        .respondOn(new AppExecutors().mainThread())
        .requestMap(response -> {
            // Do some transform from Apollo model to your model
        })
        .respondWith(new Responder<List<MyModel>>() {
            @Override
            public void onResult(List<MyModel> myModelList) {
                // Do something with the list of your model
            }

            @Override
            public void onException(Exception exception) {
                // Oops!
            }
        });

```

### DIFF
--- a/sync/src/main/java/org/aerogear/mobile/sync/SyncService.java
+++ b/sync/src/main/java/org/aerogear/mobile/sync/SyncService.java
@@ -1,5 +1,6 @@
 package org.aerogear.mobile.sync;
 
+import static org.aerogear.mobile.core.utils.SanityCheck.*;
 import static org.aerogear.mobile.core.utils.SanityCheck.nonNull;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


wtrocki:
@danielpassos I have played with this a lot and I think that wrapper is really bad idea here. It's restrictive for users so they cannot use other methods on the client. We also specify the types. Apollo library already come with rx extensions.

When trying to build something with this version I could not execute additional methods on the client etc. For example `.httpCachePolicy(HttpCachePolicy.CACHE_FIRST)`.

wtrocki:
To do it without hiding top level API we may need to override classes like ApolloCall, ApolloPrefetch & ApolloWatcher

wtrocki:
@secondsun @danielpassos Let's merge this then. I'm going to use Apollo directly for the moment and then try to figure out how to move this changes to SDK.

Issue #265, opened 2018-07-19
Metrics integration test keeps failing


trepel:
## Expected Behavior

Executing `./gradlew :core:testDebug --tests *.IntegrationTestSuite` produces no errors.

## Actual Behavior

Errors occur:
`java.lang.AssertionError: expected null, but was:<org.aerogear.mobile.core.exception.HttpException>`

The reason for this is that mock is used (this is also uncommon, you typically don't use mocks in integration tests, mocks are for unit testing) and its URL is hard coded in the test. See [MetricsServiceIntegrationTest.java](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/test/java/org/aerogear/mobile/core/integration/MetricsServiceIntegrationTest.java#L33). It seems that this mock is not working as expected.

[CONTRIBUTING.md](https://github.com/aerogear/aerogear-android-sdk/blob/master/CONTRIBUTING.md) says that you need to configure the URL in [integration-test-mobile-services.json](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/test/assets/integration-test-mobile-services.json#L11), but that file does not exist anymore. There are various files that I thought were meant for configuration of integration tests (e.g [mobile-services.json](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/test/assets/mobile-services.json)) but since the URL for metrics endpoint is hard coded, changing the URL in the json has no effect.

I think the proper fix is to make the metrics endpoint configurable again and updating CONTRIBUTING.md to say how to configure it. I think it is OK to have a `http://localhost:3000/metrics` URL as default since that is the URL you can use if you run the service locally as described in its [README.md](https://github.com/aerogear/aerogear-app-metrics#run-entire-application-with-docker-compose)

## Steps to Reproduce the Problem

  1. `./gradlew clean install`
  1. `./gradlew :core:testDebug --tests *.IntegrationTestSuite`

danielpassos:
Hey, @trepel if it fixes the problem, send a PR and I will take a look ;)

I'm going to hold on it from my side since you already have a solution.

trepel:
@danielpassos Ok, I will do some polishing and create the PR then.

trepel:
Fixed by https://github.com/aerogear/aerogear-android-sdk/pull/268. Closing the issue

PR #268, opened 2018-07-20
Fixing URL for app metrics integration test

trepel:
## Motivation

Metrics Integration test keeps failing and the metrics endpoint is not configurable as described.

Issue: https://github.com/aerogear/aerogear-android-sdk/issues/265

## Description

Hard-coded metrics URL in the integration test has been removed. It has been made configurable via mobile-services.json file. Already existing classes for parsing and getting the configuration have been used for that.

Some other small fixes are included that are related to this - not using the mockable.io URL at all. If the URL does not matter, that fact has been made clear by changing the URL to contain word 'dummy'.

We have to remove this note that is automatically prefilled in description of PR because it is not that easy to run the integration test now:

```
NOTE 
To run the integration tests, select "test/integration" label in the right column. Check the Jenkins log to get the test results
```

And also this test should be running nightly somewhere or we should make it work per PR again (I will try to make it happen).

### DIFF
--- a/core/src/test/java/org/aerogear/mobile/core/integration/MetricsServiceIntegrationTest.java
+++ b/core/src/test/java/org/aerogear/mobile/core/integration/MetricsServiceIntegrationTest.java
@@ -30,7 +30,7 @@ public class MetricsServiceIntegrationTest {
     @Before
     public void setUp() {
         MobileCore.init(RuntimeEnvironment.application);
-        metricsService = new MetricsService("https://demo1623828.mockable.io/metrics");
+        metricsService = MobileCore.getInstance().getMetricsService();
 
         error = null;
     }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,15 +68,10 @@ For running the unit tests, simly run
 
 ### Integration tests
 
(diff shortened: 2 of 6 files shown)


core/src/test/java/org/aerogear/mobile/core/integration/MetricsServiceIntegrationTest.java
```diff
@@ -29,8 +38,22 @@
...
+            MobileCoreConfiguration jsonConfig = new MobileCoreJsonParser(configStream).parse();
+            Map<String, ServiceConfiguration> configsPerId = jsonConfig.getServicesConfigPerId();
+            ServiceConfiguration metricsConfig = configsPerId.get("metrics-myapp-android");
+            metricsEndpoint = metricsConfig.getUrl();
+        } catch (JSONException | IOException exception) {
+            System.err.println(exception);
+            fail(exception.getMessage());
+        }
+
+        metricsService = new MetricsService(metricsEndpoint);
```
danielpassos: The `MobileCore.init()` automatically create a MetricsService instance so you can replace all of this by `MobileCore.getInstance().getMetricsService()`

trepel:
Thanks @danielpassos for review. I did the updates you mentioned. Please have a look if it is OK now. Thanks again.

PR #269, opened 2018-07-26
Decouple service creation from MobileCore

danielpassos:
Related issues:

* [AEROGEAR-3088](https://issues.jboss.org/browse/AEROGEAR-3088)


secondsun:
LGTM, I trust @wtrocki when he says it works.

danielpassos:
@wei-lee Wanna take a look before the merge?

wtrocki:
It's worth to add that this change unifies entire SDK to use one common aproach and make it aligned with how other SDK's are build. Without this change on master we will have 3 different API variations available which could be confusing for end users.

Issue #170, opened 2018-03-26
Remove TestSuites in favor of packages


secondsun:
## Expected Behavior

Tests in the CI environment should be run based on if they are in the unit or integration test package.
## Actual Behavior
Tests are run based on if they are in the Unit or Integration test suite.

## Specifications
- We should remove the testsuite classes
- We should move any tests not in the unit or integration packages to those packages
- We should update the README and CONTRIBUTION files to document the expected test behavior, how to write them, and how to run them.

Hattip to @josemigallas

trepel:
Hi @danielpassos, @secondsun, @josemigallas Do you have any idea of how to do this?

* removal of testsuite classes is simple
* moving unit tests into unit package and integration tests (there is just one at the moment - MetricsServiceIntegrationTest) into integration package is also simple
* I am not sure how to make gradle run tests from unit package only - possibly by specifying just some subset of src/test/java directories as source directories in 'test' sourceSet or to move integration package out of src/test/java (default source dir for test classes). Any preference, or even other options?
* How to run integration tests only - creating custom task for this ?
** Using approach like in https://github.com/HIFILEO/ReactiveMvp/blob/master/app/build.gradle might be the way to go (no need to create modify 'test' sourceSet at all)
* Once done, documenting how to run them is simple
* Documenting why/when/how to write the tests is more complex but it can be done separately IMO

Any guidance/comments/concerns appreciated. Thanks. If you think it sounds good let me know, I can come up with a PR to discuss in more dept over it.

josemigallas:
My first attempt was to create packages and in theory it is perfectly possible to run a reduced set of tests by using `gradlew`. But for some reason, it wouldn't work so I finally went for the `TestSuite`.

It should be as simple as:
```bash
$ ./gradlew testDebug --tests org.aerogear.mobile.core.integration.*
$ ./gradlew testDebug --tests org.aerogear.mobile.core.unit.*
```

trepel:
Hi @josemigallas ,
my thought was that if you run `./gradlew test` or `./gradlew testDebug`, only unit tests would be executed. And that there would be another custom task for integration tests. The reason is that integration tests are somewhat special and should not be part of standard tasks and processes. If that makes sense.

trepel:
[...]

PR #270, opened 2018-07-30
Split of unit and integration tests execution

trepel:
## Motivation

Issue https://github.com/aerogear/aerogear-android-sdk/issues/170 (see also discussion there)

Test suite classes have been removed. All the unit test now are in `unit` package. CONTRIBUTING.md has been updated to describe how to run both the unit and integration tests.

Tests are now annotated and based on the annotation it is decided whether the test is unit or integration test.

Note that the modifications have only been done for core module. Other modules (auth, push, etc) only contain unit tests so this PR is not necessary for them. If you agree with the modifications I am happy to update other modules as well though.

## Aditional notes

I believe the https://github.com/aerogear/aerogear-android-sdk/commit/c67c0503a5d75ba754ef1475f84c42c69e2ad49b can be reverted now since `test` task now only run unit tests. I can send another PR if there is an agreement on that.

## How to test

Just follow the CONTRIBUTING.md


core/src/test/java/org/aerogear/mobile/core/integration/MetricsServiceIntegrationTest.java
```diff
@@ -7,13 +7,15 @@
 
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
```
danielpassos: It's nice to see that category annotation is experimental but the IntegrationTest category itself is not LOL

core/build.gradle
```diff
@@ -26,6 +29,18 @@ android {
     testOptions {
         unitTests {
             includeAndroidResources = true
+            all {
+              useJUnit()
+              if (integrationTests.toBoolean()) {
```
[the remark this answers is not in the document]
trepel: Exactly @danielpassos . I am not sure if you have specific use case in mind when you would like to run both unit and integration tests. The only use case I can think of is that you might want to measure combined code coverage. That is something I would like to have a look at soon, but since it can be changed easily I would leave it as it is for now.

core/src/test/java/org/aerogear/mobile/core/integration/MetricsServiceIntegrationTest.java
trepel: Yes, `@Category` is junit class whereas `UnitTest` and `IntegrationTest` are our custom interfaces.

trepel:
Thanks for review @danielpassos , I'll merge it some time today and I will also prepare another PR for reverting https://github.com/aerogear/aerogear-android-sdk/commit/c67c0503a5d75ba754ef1475f84c42c69e2ad49b

PR #273, opened 2018-08-02
Added Code Coverage

trepel:
## Motivation

Having code coverage computed is very useful.

## Description

Added some JaCoCo config and tasks that are able to generate the code coverage reports - for each module individually (jacoco.gradle) and for 'root' report that combines all the module reports together (build.gradle). The build. gradle also contains few variables used in both individual and root jacoco tasks to avoid duplicity definitions of these variables.

## Additional Notes

Next step would be to integrate with coverals or codecov and add nice badge to README.md so that we can track the trend of code coverage over the time.


danielpassos:
@trepel Can you rebase?

danielpassos:
@trepel just a note about the next steps, we have used [codecov](https://codecov.io/gh/aerogear) on some [AeroGear projects](https://codecov.io/gh/aerogear) with [shields.io](http://shields.io) for AeroGear Android Store badge

trepel:
Hi @danielpassos , rebased.

I know about both codecov and coveralls being used for some AeroGear projects. That's why I wanted to investigate that two. I read you comment so that you would prefer codecov, thus I will start with that.

I will be on holiday next week so don't be surprised if I don't reply on your comments.

PR #283, opened 2018-08-16
Core interceptors

wtrocki:
## Motivation
 
Add ability to have dynamic interceptors - currently we build fixed instance of the http layer every time and spread that across the services so there is not way to swap this. I think that having fixed instance of http later is not going to be good idea in long run as we need to have way to split between SDK that require different types of headers etc. I did not wanted to refactor this as this will affect certificate pinning and other functionalities I do not want to touch (and verify if they work)

## Top level usage

```javascript
MobileCore.getInstance().requestHeaderInterceptor().
add(authService.getAuthHeaderProvider());
```

## Progress

- [x] Add header provider to auth SDK
- [x]  Add interceptor layer to Core
- [x] Restore auth header
- [ ]  Doc update (separate PR)
 
## Verification

1. ./gradlew install sdk
2. Get memeolist branch: https://github.com/aerogear/aerogear-android-example-apps/pull/15
3. Get server branch:  https://github.com/aerogear/data-sync-server/pull/54

Verification is quite complex
Going to post end to end video for this as well to show how it works.

## Demo

https://groups.google.com/forum/#!topic/aerogear/kwtX5XFByf0

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: "com.android.library"
 
 android {
-    compileSdkVersion 27
+    compileSdkVersion project.ext.targetSdkVersion
 
     defaultConfig {
         minSdkVersion project.ext.minSdkVersion
--- a/auth/src/main/java/org/aerogear/mobile/auth/AuthHeaderProvider.java
+++ b/auth/src/main/java/org/aerogear/mobile/auth/AuthHeaderProvider.java
@@ -11,7 +11,7 @@ import org.aerogear.mobile.core.http.interceptors.HeaderProvider;
  */
 public class AuthHeaderProvider implements HeaderProvider {
 
-    public static final String TYPE = "Bearer ";
+    public static final String HEADER_TYPE = "Bearer ";
     public static final String HEADER_KEY = "Authorization";
 
     private AuthService authService;
@@ -30,7 +30,7 @@ public class AuthHeaderProvider implements HeaderProvider {
         UserPrincipal user = authService.currentUser();
         if (user != null && user.getAccessToken() != null) {
             String accessToken = user.getAccessToken();
-            return Collections.singletonMap(HEADER_KEY, TYPE + accessToken);
+            return Collections.singletonMap(HEADER_KEY, HEADER_TYPE + accessToken);
         }
         return Collections.emptyMap();
     }
--- a/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
+++ b/core/src/main/java/org/aerogear/mobile/core/MobileCore.java
@@ -53,7 +53,6 @@ public final class MobileCore {
     private final Map<String, ServiceConfiguration> serviceConfigById;
     private final Map<String, List<ServiceConfiguration>> serviceConfigsByType;
     private final MetricsService metricsService;
-    private final RequestHeaderInterceptor dynamicInterceptor;
 
     /**
      * Get the user app version from the package manager
@@ -97,7 +96,7 @@ public final class MobileCore {
 
         // -- HTTP layer --------------------------------------------------------------------------
         OkHttpClient.Builder builder = new OkHttpClient.Builder();
-        dynamicInterceptor = new RequestHeaderInterceptor();
+        RequestHeaderInterceptor dynamicInterceptor = new RequestHeaderInterceptor();
         builder.addNetworkInterceptor(dynamicInterceptor);
 
         OkHttpCertificatePinningParser certificatePinning =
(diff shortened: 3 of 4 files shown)


wtrocki:
Upon discussion we decided to not use fully featured request and just stick to simple hashmap that will hold headers only.

auth/build.gradle
```diff
@@ -1,13 +1,14 @@
 apply plugin: "com.android.library"
 
 android {
-    compileSdkVersion project.ext.targetSdkVersion
+    compileSdkVersion 27
```
danielpassos: Why was it changed?

auth/build.gradle
```diff
@@ -1,13 +1,14 @@
...
 
     defaultConfig {
         minSdkVersion project.ext.minSdkVersion
         targetSdkVersion project.ext.targetSdkVersion
 
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        android.defaultConfig.manifestPlaceholders = ['appAuthRedirectScheme': 'Invalid']
```
danielpassos: Why we need it?

auth/src/main/java/org/aerogear/mobile/auth/AuthHeaderProvider.java
```diff
@@ -0,0 +1,37 @@
...
+import org.aerogear.mobile.auth.user.UserPrincipal;
+import org.aerogear.mobile.core.http.interceptors.HeaderProvider;
+
+/**
+ * Default provider class that will build headers based on currently logged in user access token
+ */
+public class AuthHeaderProvider implements HeaderProvider {
+
+    public static final String TYPE = "Bearer ";
+    public static final String HEADER_KEY = "Authorization";
```
danielpassos: HEADER should be removed or add also in TYPE

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
```diff
@@ -236,4 +240,14 @@ public ServiceConfiguration getServiceConfigurationById(final String id) {
         return serviceConfigById.get(id);
     }
 
+    /**
+     * Returns manager that adds new interceptor to the chain of core http interceptors that could
+     * be used to add headers to network requests
+     *
+     * @return RequestHeaderInterceptor
+     */
+    public RequestHeaderInterceptor requestHeaderInterceptor() {
```
danielpassos: Should not this be [HttpService's](https://github.com/aerogear/aerogear-android-sdk/blob/master/core/src/main/java/org/aerogear/mobile/core/http/HttpServiceModule.java) responsibility?

auth/build.gradle
wtrocki: It was breaking stuff for production, but placeholder is needed for testing local usage. Without that build fails.

core/src/main/java/org/aerogear/mobile/core/MobileCore.java
wtrocki: Apollo is using raw okhttp.

danielpassos: Not sure if I follow you here. The sync okhttp instance come from the HttpService => `MobileCore.getInstance().getHttpLayer().getClient()`

auth/src/main/java/org/aerogear/mobile/auth/AuthHeaderProvider.java
wtrocki: You mean `HEADER_KEY` => `_KEY` => `TYPE_KEY`?

core/src/main/java/org/aerogear/mobile/core/http/interceptors/RequestHeaderInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Override
+    public Response intercept(Interceptor.Chain chain) throws IOException {
+        Request.Builder builder = chain.request().newBuilder();
+        // Process header providers
+        for (HeaderProvider provider : providers) {
+            Map<String, String> headers = provider.getHeaders();
+            for (String header : headers.keySet()) {
+                builder.addHeader(header, headers.get(header));
```
wojta: I'm thinking that it's possible to have headers with multiple same keys. I know that it's used in `Set-Cookie`, but I'm not sure if it's needed also for something else.

wtrocki: HTTP spec do not allows the same headers twice. It looks like last one wins in most of the cases:
https://github.com/nodejs/node/issues/3591

PR #288, opened 2018-08-24
Checking if the apollo response has no error

danielpassos:


sync/src/main/java/org/aerogear/mobile/sync/SyncService.java
```diff
@@ -166,7 +169,15 @@ public void onFailure(@Nonnull ApolloException e) {
                             .mutate(mutation).enqueue(new ApolloCall.Callback<T>() {
                                 @Override
                                 public void onResponse(@Nonnull Response<T> response) {
-                                    requestCallback.onResult(response);
+                                    if (response.hasErrors()) {
+                                        for (Error error : response.errors()) {
```
wtrocki: Actually just seen this. -1000 for that. This is not going to work with requirements for auth. 
Graphql is very rich and allows to send full objects as errors. So we can have authentication./sync errors etc. as part of that.
I believe that we should not do it as we will not allow people to get that data back

danielpassos: And... what do you suggest? Do the same of the Apollo client call the success and make this responsibility to the developer consume it or create a SyncException and add a List<Error> there?

wtrocki: Yeah.. I initially thought that this will work for us, but in fact users can define their own JSON so wrapping this to error object will prevent them from getting this details. It's nice to wrap things, but we also need to provide the same level of flexibility for users. I haven't seen that before but currently using the latest version of the SDK with your changes and see this becoming a problem even for our internal stuff.

wtrocki: Plus FIME for subscriptions etc.

danielpassos:
@wtrocki https://github.com/aerogear/aerogear-android-sdk/pull/290

wtrocki:
@danielpassos Awesome!

PR #290, opened 2018-08-27
Remove error apollo response error check

danielpassos:


wtrocki:
Codecoverage faling. Retesting again

PR #291, opened 2018-08-27
Make respondOn mainThread default for all sync rx call

danielpassos:


secondsun:
If you are having an opinion about what the response thread is then we should make sure that we add javadocs.  It would be frustrating for a developer if they got a network on main thread exception because they chained a method off this.

danielpassos:
@secondsun +1 but it `respondOn` can be changed by the developer right? ;)

wtrocki:
@danielpassos This is actually really good question.

secondsun:
@wtrocki @danielpassos  Yes it can always be changed, responses are build by using proxies.  So you will have the response call a proxy which will run on the main thread which will immediately call a proxy which will run on the thread the user wants.

PR #298, opened 2018-10-09
Renamed getFreshUser to getCurrentUser(boolean autorefresh)

ziccardi:
## Motivation

Both `getFreshUser` and `getCurrentUser` returns the current logged in user. The difference is that the `getFreshUser` automatically refresh the token if needed.

Instead of duplicating code, this PR changes it to a parameter.

Moreover, this PR make android code coherent with what we have in the other platforms (iOS and Xamarin)


ziccardi:
ping @secondsun @danielpassos

